PR #18, opened 2015-06-13
Fix build flags handling

debfx:
$CFLAGS/$CPPFLAGS/$LDFLAGS weren't always passed to the compiler.
These commits fix all cases I could find.

### DIFF
--- a/autoconf/configure.in
+++ b/autoconf/configure.in
@@ -4624,7 +4624,7 @@ if test "${support_bat}" = "yes" ; then
    touch bat
    chmod 755 bat
    rm -f Makefile
-   $QMAKE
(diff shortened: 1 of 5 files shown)


joergsteffens:
Hi, that for the patches.
2993fbd and 23ff029 look fine. However, e566850 doesn't seam to be correct, as CFLAGS should only be used with C-compilers, not with C++ compilers (as you did in the qmake patch). Did I miss something?
Out of curiosity: did you experienced any problems without your patches?

debfx:
> However, e566850 doesn't seam to be correct, as CFLAGS should only be used with C-compilers, not with C++ compilers (as you did in the qmake patch).
Indeed, I've fixed that.

> Out of curiosity: did you experienced any problems without your patches?
Not particularly, except that it didn't use hardening build flags in all places when exporting them in the Debian package.

joergsteffens:
Looks good. I will run some tests against it.

PR #20, opened 2015-06-19
Update bareos-dir.conf.in

rkarlsba:
Add heartbeat = 1m in default config
While this isn't strictly needed on fast, low-latency networks, I see it's needed on our 10Gbps network over 20km distance, although I'm not exactly sure why


joergsteffens:
Enabling the heartbeat cases overhead (additional thread in Bareos). However, as some (a few) people have reported problems here, I added a comment about this in the default configuration, see https://github.com/bareos/bareos/commit/76cebc59643fee73f338adbc41a9c15fab2adce8

I'm closing the pull request now. You may want to raise a discussion on the bareos-users mailinglist, so we can figure out how many people are affected by this problem. If many environments are affected, we may add this settings later on.

PR #21, opened 2015-06-19
Add default 60s heartbeat

rkarlsba:
Not all installations need heartbeat, but most do, and it won't hurt


joergsteffens:
see #20

PR #23, opened 2015-06-19
Default to Compatible = no

rkarlsba:
If people install the bareos client, it's quite likely they use bareos on the director/sd as well, so I suggest using Compatible = no as the default.


joergsteffens:
This most likely makes sense. We plan to change this default configuration with 15.2 and have this settings already in our internal testing branch.
Therefore I close this pull request, but until we find some problems, this setting will change soon.

PR #30, opened 2015-08-07
desktop file fixes

evgeni:
* drop redundant Comment from bat.desktop
* drop unused X-Desktop-File-Install-Version from desktop files
* drop deprecated Encoding key from the desktop files


evgeni:
Merged by @joergsteffens in 7e23e6f82dd2b173adf899277638be740a980a62, thanks!

PR #32, opened 2015-08-31
drop bareos-filedaemon-python-plugin.dir, it's unused

evgeni:


joergsteffens:
Thanks for the patch.

PR #35, opened 2015-10-12
improve package descriptions

evgeni:
small fixes found while reviewing the rest of the d/control, enjoy


joergsteffens:
Hello Evgeni,
thank you for your pull request. I will gladly apply it, if you would be so kind to merge the three commits into a single commit? If you don't have the time to do so, I can do it, of course with keeping for as the author. Only disadvantage is, that this pull request will then be closed, not merged.

evgeni:
Not that I see a reason for that squashing, but done :-)

PR #37, opened 2015-11-06
clarify appropriate use of PluginNames

athompso:
It seemed like leaving PluginNames blank was a good idea.  This may stop the next user from believing that.


joergsteffens:
Dear Adam, we discussed this change internally for a while. However, we decided not to add it.

Reasons:
- "For production use" is not a phrase, we like to have in our code. Instead the behaviour should be described.
- If we add this comment here, it should also be added to the other locations, where Plugin Names can be defined
- The description in the source code is only a summary. Additional information should be placed in bareos-docs.

Nevertheless, thanks for pointing this out. I'll keep this in mind and hopefully find a way to improve the documentation.

regards,
Jörg

athompso:
I agree that this belongs in the docs, not in the code, but as far as I 
can tell, that section of the docs is generated automatically from the 
code.  I couldn't figure out how to put it in -docs other than this way.
-Adam

joergsteffens:
Yes, the description from the source is taken as summary for the documentation. However, the documentation can include more information.
The additional information in this case is in https://github.com/bareos/bareos-docs/blob/master/manuals/en/main/director-resource-director-definitions.tex

PR #46, opened 2015-12-27
Start bareos-dir.service after mariadb.service

robert-scheck:
Fedora and RHEL are shipping mariadb.service rather mysql.service (at least by default)


joergsteffens:
Thank you for your patch. We have added it to the bareos-15.2 branch as https://github.com/bareos/bareos/commit/f2cbdcbf7ab940af932a59582b265870b687a259, which have in the meanwhile been merged to master.
Therefore I'll close this pull request.

PR #47, opened 2015-12-27
Merge bareos-database-common into bareos-common

robert-scheck:
Merge bareos-database-common into bareos-common (why having a separate package that is always needed?)


mvwieringen:
I think you missed one really important thing and that is that bareos-database-common is only
needed for things that interface with the catalog database e.g. Director and some of the btools.
So with your change people just installing the Filed on a client server also have to import the
catalog bits which make absolutely no sense on those installs.

joergsteffens:
As Marco pointed out, there is good reason why bareos-database-common and bareos-common are separated packages. I'll therefore close this pull request.

PR #49, opened 2016-01-28
Ceph Rados: added id argument to specify rados client id

Crapworks:
Hi all,

thanks for adding ceph rados support to bareos! As it is right now, you have no other choice than to use the client.admin key on bareos, because there is no way to specify a client id for `rados_create()`. I've added a config option "id" where the user can specify a client id.

Since you are using `rados_create()` and not `rados_create2()` the value should be "admin", not "client.admin".

Cheers,
Christian

### DIFF
--- a/src/plugins/filed/plugin-rados.conf
+++ b/src/plugins/filed/plugin-rados.conf
@@ -16,7 +16,7 @@ FileSet {
     # adapt this to your environment
     #   namespace: optional
     #   snapshotname: optional (a snapshot in any case)
-    Plugin = "rados:conffile=<ceph_conf_file>:id=<client_id>:poolname=<pool_name>:namespace=<name_space>:snapshotname=<snapshot_name>:"
+    Plugin = "rados:conffile=<ceph_conf_file>:clientid=<client_id>:poolname=<pool_name>:namespace=<name_space>:snapshotname=<snapshot_name>:"
   }
 }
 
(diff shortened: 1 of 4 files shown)


mvwieringen:
Hi,

Thanks for adding this. When we wrote the support rados_create2 was not available so that
explains why we didn't use it. You could add support for it if you add it in the same way we
support the changed write API e.g. #if LIBRADOS_VERSION_CODE <= 17408
you only need to find out since which version rados_create2 is part of API.

I would also like to rebase things into 2 commits e.g. one for the rados-fd and one for the backend.
As far as I can see the two others are mostly fixes to one of the previous ones.

Crapworks:
Hi Marco,

I see! I will have a closer look at the ceph integration in the next time. This is just a quick fix so we can use a non-admin key in our infrastructure and I wanted to make sure it will find it's way to upstream. I will have a look at a rados_create2() integration as well in the next time.

Did a rebase to combine the last commits.

Cheers,
Christian

mvwieringen:
I think there is one typo in the patch for the storage backend.

{ "id=", argument_id, 9 },

should probably be

{ "id=", argument_id, 3 },

Maybe we should also call it clientid ?
If you can fix that then I can merge the pull request.

Crapworks:
Hi, I changed the name to clientid.

mvwieringen:
I see one typo e.g. clientd instead of clientid.
I close the pull request just saved the patches and will commit them after some cleanups under your name in master. So it will be visible by the end of the week when we sync to github.

PR #50, opened 2016-06-13
dbcheck: do not find empty path as bad.

mastervolkov:
In file src/cats/bvfs.c function bvfs_parent_dir return '\0' as root
path, but in file src/dird/dbcheck.c function repair_bad_paths check
is "WHERE Path NOT LIKE '%/'". That query always find "root" = NULL
as bad and always fix that.

Current patch change that query to "WHERE Path NOT LIKE '%/' AND NOT
NULL", which solves the problem.

### DIFF
--- a/src/dird/dbcheck.c
+++ b/src/dird/dbcheck.c
@@ -1298,7 +1298,7 @@ static void repair_bad_paths()
 
    printf(_("Checking for Paths without a trailing slash\n"));
    query = "SELECT PathId,Path from Path "
-           "WHERE Path NOT LIKE '%/'";
+           "WHERE Path NOT LIKE '%/' AND NOT NULL";
    if (verbose > 1) {
       printf("%s\n", query);
    }


pstorz:
Hello,

I don't understand why this patch is needed; dbcheck checks what is stored in the path table, while bvfs_parent_dir() gives the parent dir in the virtual filesystem. The output of bvfs_parent_dir() is never stored into the database as far as I know.

If you can show a testcase where you can show that there is a problem, I will have a look but for now I don't see that this patch is needed.

Thanks for helping us with the project.

Best regards,

Philipp

mastervolkov:
Of course.

1. Do 'bareos-dbcheck -v -f ' and fix any problem
2. In bareos-dbcheck we do "4) Repair bad Path records"
```
Checking for Paths without a trailing slash
Found 0 bad Path records.
```
3. In bconsole we do '.bvfs_update jobid=AnyRealJob'
4. In bareos-dbcheck we do "4) Repair bad Path records"
```
Checking for Paths without a trailing slash
Found 1 bad Path records.
```

Repeated 100% on bareos-dbcheck "4) Repair bad Path records" after any .bvfs_update jobid

pstorz:
Ok, I see. Thank you.

Would not be the right solution to make bvfs_parent_dir() return '/' instead of NULL as root path?

mastervolkov:
Can be problem for windows fileset with letter... But i not sure

чт, 21 июля 2016, 13:30 pstorz <notifications@github.com>:

mastervolkov:
return '/' could create a problem for Win clients.

In next output i do .bvfs_lsdirs for Windows (jobid=4) and Linux (jobid=5) jobs. In Windows jobs first '/' unneded and drive letter used.

```
*.bvfs_update jobid=4
*.bvfs_lsdirs jobid=4 path=
243170  0       0       0       A A A A A A A A A A A A A A     .
243169  0       0       0       A A A A A A A A A A A A A A     E:/
*.bvfs_update jobid=5
*.bvfs_lsdirs jobid=5 path=
243170  0       0       0       A A A A A A A A A A A A A A     .
243172  0       0       0       A A A A A A A A A A A A A A     /
```

mastervolkov:
Do I need to do another pull request?(Just the status is closed, but fix adsent in master).
Or @pstorz already added fix and merge later?

pstorz:
Hello,

No, you don't need to add another pull request.
Your change does not fix the problem, it only fixes the symptoms.
The problem is that an empty path is added to the database which does not make sense.

We have a patch that fixes that problem but that needs to be tested a bit more.

thanks and best regards,

Philipp

PR #51, opened 2016-07-28
fix imperformant sql query for 'list jobs last'

toerb:
This pull request fixes a really imperformant sql query for 'list jobs last' which is executed every time the webui dashboard is loaded.
On our production servers with mysql the query process time goes down from 10 minutes to 0.03 seconds.

### DIFF
--- a/src/cats/sql_cmds.c
+++ b/src/cats/sql_cmds.c
@@ -96,6 +96,26 @@ const char *list_jobs =
    "%s "
    "ORDER BY StartTime%s";
 
+const char *list_jobs_last =
+   "SELECT DISTINCT "
+   "Job.JobId,Job.Name, "
+   "Client.Name as Client, "
+   "Job.StartTime,Job.Type,Job.Level,Job.JobFiles,Job.JobBytes,Job.JobStatus "
+   "FROM Job "
+   "LEFT JOIN Client ON Client.ClientId=Job.ClientId "
+   "LEFT JOIN JobMedia ON JobMedia.JobId=Job.JobId "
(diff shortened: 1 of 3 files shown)


mvwieringen:
I guess this should be against 15.2 and not master. Could you create a bug in bugs.bareos.com
and attach the format-patch to that new bug there ?

mvwieringen:
Ok I extracted the patch into a format-patch myself and applied it to 15.2.
It should be published tomorrow including a small layout tweak to the queries.
As such I close this pull request.

toerb:
Oh I was just fixing a bug introduced with this patch before commiting it into the bugtracker. Other keywords like jobstatus are not working as expected with this layout.
What tweaks did you apply?

toerb:
See my latest commit 
https://github.com/toerb/bareos/commit/46d5a07819f3a400e6b409adef96a35d312a2165
I just added the modifiers of the keywords into the subquery.

mvwieringen:
I removed a unused variable and redid the queries a bit to be better readable.

toerb:
Can you please add the changes from my latest commit?

mvwieringen:
I guess the problem is only with list/llist jobs right ? I squashed the second patch into the first one
and rebased the branch.

toerb:
The problem is only with list/llist jobs last.
list jobs last jobstatus=F resulted before my first commit in a list of the last failed job per client.
After my first commit the query resulted in a list of last jobs per client where jobstatus is failed.
My second commit has fixed this and reproduces the old behaviour.

PR #52, opened 2016-07-28
Add error message when action on purge is not set in the pool resource

wanderleihuttel:


mvwieringen:
Guess this should be against 15.2 not master so I did a clone/format-patch and added it there.
As such I close this pull request.

PR #53, opened 2016-09-23
plugins: Prevent segfault if plugin_list is null

M1Sports20:
I admit I haven't taken the time to fully understand the code execution path around plugins.  I do not know if this is the correct fix to <>.  It does prevent bareos-sd from crashing on my system.

Fixes: https://bugs.bareos.org/view.php?id=662

Commit Message:
This storage daemon can call list_plugins when plugin_list is null.  This
commit adds a check to make sure the plugin_list variable is not null.

### DIFF
--- a/src/lib/plugins.c
+++ b/src/lib/plugins.c
@@ -364,7 +364,10 @@ int list_plugins(alist *plugin_list, POOL_MEM &msg)
 {
    int i, len = 0;
    Plugin *plugin;
-
+   
+   if (!plugin_list) {
+      return 0;
+   }
    if (plugin_list->size() > 0) {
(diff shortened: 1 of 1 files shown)


M1Sports20:
Let me know if you want me to do something else or test something.

I do have the default plugin directory along with a single plugin in the directory:
/usr/lib/bareos/plugins/autoxflate-sd.so

joergsteffens:
Thanks for the patch. I added a comment in the https://bugs.bareos.org/view.php?id=662

joergsteffens:
As discussed in https://bugs.bareos.org/view.php?id=662, this bug is related to new gcc compiler and optimization. It is fixed in current 16.2 releases.

PR #56, opened 2016-11-08
PAM authentication of consoles

smanolache:
Hello,

This code lets Directors authenticate users using Consoles via PAM calls.

I have included some code from Marco van Wieringen.

The code changes the handshake between the Console and the Director. If the Director does not understand the new handshake, then the Console falls back to the current handshake.

The handshake runs as follows:

C->D: Hello console_name protocol=2 ssl=0|1|2
D->C: 1000 OK: dir_name version date protocol=2 ssl=0|1|2 ssl_auth_only=0|1
C->D: STARTTLS
D->C: 1000 OK STARTTLS
TLS handshake
C->D: auth plain
D->C: +
C->D: login base64-user-name base64-password
D->C: 1000 OK auth
C->D: auth cram-md5 challenge ssl=0|1|2
D->C: response_to_challenge
C->D: 1000 OK auth
TLS tear-down if ssl_auth_only

The Director refuses to authenticate the Console if:
* The Console is the default console
* PAM authentication is not configured for the named console
* TLS is not activated

Additionally, the Console configuration on the Director side contains a list of allowed users or groups that add an authorization layer on top of the PAM authentication.


joergsteffens:
Hi,

thank you for this pull request. I'm aware, that quite some people like to see PAM support in Bareos.
However, as this is a potential critical patch, we have to find the time to review and test it carefully. Also regression tests (https://github.com/bareos/bareos-regress) must exist, before we can add it.

I just wanted to let you know, that we have seen and discussed (also not in technical detail) your patch.

regards,
Jörg

smanolache:
I have pushed some modifications to my branch:

1. I have added a command-line switch to bconsole. By default, if the 'new' authentication method fails the console will abort. If the command-line switch '-p' is specified then the console falls back on the 'old' authentication method, i.e. cram-md5.

2. I have fixed a bug in my own code that caused the director to crash (segfault) when the console was not found by the Director in its configuration files. I was using cons->tls before checking whether cons is NULL.

3. I have removed the cram-md5 authentication that followed the user/pass authentication. Thus, there is no mechanism any more by which the Director authenticates with the Console. But, on the plus side, there is no need any more to store passwords in the configuration files. (Note that we still have authentications of Consoles with the Director, thanks to the user/pass that are passed. Just that we don't have authentications of Directors with Consoles.)

franku:
Starting with 18.2 we provide a direct TLS (PSK/Cert) handshake without sending "STARTTLS". Additionally our own PAM implementation has a different approach with an interactive callback. As a matter of fact we decided not to implement your pull request. However, thank you so much for your contribution.

PR #58, opened 2017-02-23
File size change should be treated as warning, not error.

glbyers:
File size change is treated as warning during backup, but as an error during restore. Backing up a live filesytem rather than a snapshot is often going to result in this behaviour and whilst not ideal, it isn't an error. I believe this should also be true for restore.


pstorz:
Hello,

thanks for your patch. I think that the right way of fixing this problem would be to make sure the real size that was backed up is recorded.

Then the check during restore make sense again and should still throw an error if it differs.

Best regards,
Philipp

arogge:
Thank you very much for your effort.
However, I'm going to close the PR as we don't accept PRs for bareos-16.2 anymore and we do not believe that your approach is the right solution. Sorry.

Best Regards,
Andreas

PR #60, opened 2017-04-26
fix cannot recongnition-l option

wangchuan2008888:
before fix run bregex -l -f filename will got -----./bregex: invalid option -- 'l'
I'm sure the getopt function's parm has some mistake.
"d:f:nl?" option would work properly.

### DIFF
--- a/src/tools/bregex.c
+++ b/src/tools/bregex.c
@@ -87,7 +87,7 @@ int main(int argc, char *const *argv)
    bindtextdomain("bareos", LOCALEDIR);
    textdomain("bareos");
 
-   while ((ch = getopt(argc, argv, "d:f:n?")) != -1) {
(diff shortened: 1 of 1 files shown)


pstorz:
Hello wangchuan2008888,

thank you for that patch. Do you have a
real-world name so that we can accept the patch
and add your name to the "authors" file?

wangchuan2008888:
Hi Pstorz,

thinks for accept my patch, it's be my honor to be parts of Bareos authors. My name is chuan wang.

pstorz:
Hello Chuan Wang,

I have merged your changes into our internal repo which will be pushed to github after it passes all tests.

Then this pach will be visible here on github as accepted, and I also updated the AUTHORS file with your name.

Thank you very much!

PR #62, opened 2017-08-02
Fix crash when debugging output is enabled

MedicMomcilo:
This should fix #0000826: Enabling debugging output crashes Director


joergsteffens:
I applied this patch.
However, I changed the commit message, so it automatically closes the related bug and applied it to the bareos-16.2 branch, instead of master.
Therefore, it pull request will not be automatically closed. Therefore I close it manually.

joergsteffens:
It will be merged to master soon.

PR #63, opened 2017-09-24
dbcheck.c, update for dver 2171, remove eliminate_duplicate_filenames…

unki:
…(), update all queries referencing the old Filename table

The schema upgrade from 2004 to >=2170 seems to have rendered ```bareos-dbcheck``` unusable, because it still tries to work on the ```Filename``` table, which got merged into ```File``` by the schema upgrade.

I'm not sure if I've done that correct, but this PR updates the SQL queries to use the ```File``` table instead.

Furthermore it removes the ```eliminate_duplicate_filenames()``` function, which now seems to be obsolete, as it only tried to find duplicate filename-entries in the ```Filename``` table.
As these duplicates might now always be the case in the ```File``` table, as there the files are associated with ```Paths``` or ```Jobs```, and by this might occur multiple times, I think there is no more use for ```eliminate_duplicate_filenames()```.

### DIFF
--- a/src/dird/dbcheck.c
+++ b/src/dird/dbcheck.c
@@ -80,7 +80,6 @@ static int make_name_list(const char *query, NAME_LIST *name_list);
 static void print_name_list(NAME_LIST *name_list);
 static void free_name_list(NAME_LIST *name_list);
 static char *get_cmd(const char *prompt);
-static void eliminate_duplicate_filenames();
 static void eliminate_duplicate_paths();
 static void eliminate_orphaned_jobmedia_records();
 static void eliminate_orphaned_file_records();
@@ -327,7 +326,6 @@ int main (int argc, char *argv[])
    if (batch) {
       repair_bad_paths();
       repair_bad_filenames();
-      eliminate_duplicate_filenames();
       eliminate_duplicate_paths();
       eliminate_orphaned_jobmedia_records();
       eliminate_orphaned_file_records();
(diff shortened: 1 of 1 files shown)


joergsteffens:
Thank you very much for this patch and sorry for the delay. We're going to merge this to bareos-17.2 branch soon.

joergsteffens:
I noticed, that you have modified the function eliminate_orphaned_filename_records(), to be based on PathIds instead of FilenameIds. However, I think this function must be removed, as the Filename table no longer exists. PathIds are covered by eliminate_orphaned_path_records().
Am I missing something?

unki:
Hello!

Imho ```eliminate_orphaned_file_records()``` is the one that removes orphaned _file entries_ that are no longer bound to a _job_ (join on the ```Job``` table).

And the ```eliminate_orphaned_path_records()``` is the one that removed orphaned _path entries_ where no more _files_ are associated with them (join on the ```File``` table).

I'm not 100% sure, but I think those two are still doing a different job. It was only ```eliminate_duplicate_filenames()``` that has become obsolete.

joergsteffens:
I fully agree. However I asked about ```eliminate_orphaned_filename_records()```, not ```eliminate_orphaned_file_records()```.  And after checking this again, I quite certain that this function should be removed.

unki:
Argl, sorry. Those function names are easily for me to be mixed up ;-)

* ```eliminate_orphaned_file_records()``` -> remove files that are no longer bound to any **job** (join on Job table)
* ```eliminate_orphaned_path_records()``` -> remove paths that are no longer referenced by any file (join on File table)
* ```eliminate_orphaned_filename_records()``` -> remove files that are no longer bound to any **path** (join on Path)

AFAIK there is no other function right now, that performs the same task as ```eliminate_orphaned_filename_records()```, to cleanup file residues in the File table, where no more path exists in the Path table (why ever this situation has occurred).

So perhaps ```eliminate_orphaned_filename_records()``` has still a right to exist, but might need to be renamed to ```eliminate_stale_file_records()``` or so?

Or extending the functionality of ```eliminate_orphaned_file_records()``` be modifying the SQL query there, to not only join on the Job table, but also on the Path table (so doing the same job as ```eliminate_orphaned_filename_records()``` right now).

joergsteffens:
Unfortunately, I've been distracted  and forget to integrate this into bareos-17.2 and then wondered, why this pull request is still open. Sorry for this. It is now merged to bareos-17.2 and master will follow. As I decided to merge it to bareos-17.2 (and not master as in the original pull request), this pull request will not be closed automatically. I will close it manually. Please reopen if you still see problems.

PR #64, opened 2017-10-13
self.jobType in BareosDirPluginBaseclass should be a char, not int

kozl:
Example:
for restore jobs self.jobType is 82, chr(82) is 'R'. It looks like bareosdir.GetValue(context, brDirVariable['bDirVarType']) should be chr'ed


aussendorf:
Hello kozl, your suggestion sounds reasonable.
Have you tested your patch with any director plugin derived from the modified base-class ?
Maik

kozl:
No, unfortunately we don't use any.

arogge:
Thank you very much for your effort.
However, jobType is actually really an int and not a char. Yes this is braindead, but unless you're willing to change it across all of Bareos, it is going to stay an integer for now.

Best Regards,
Andreas

PR #66, opened 2017-11-14
Fix compilation with MariaDB 10.2

ptrunk:
Hi,

bareos does not seem to compile with MariaDB 10.2 on Fedora 27 because of

```
mysql.c: In member function 'virtual bool B_DB_MYSQL::db_open_database(JCR*)':
mysql.c:194:15: error: 'MYSQL {aka struct st_mysql}' has no member named 'reconnect'
    m_instance.reconnect = 1;             /* so connection does not timeout */
               ^~~~~~~~~
make[1]: *** [Makefile:205: mysql.lo] Error 1
```
It looks like the same bug as described here: https://bugzilla.redhat.com/show_bug.cgi?id=1467706

Please have a look at the commit. With the new mysql.c file bareos compiles fine on Fedora 27 with MariaDB 10.2 for me.

Thank you!

Best regards,
Paul

### DIFF
--- a/src/cats/mysql.c
+++ b/src/cats/mysql.c
@@ -156,6 +156,7 @@ bool B_DB_MYSQL::open_database(JCR *jcr)
 {
    bool retval = false;
    int errstat;
+   my_bool reconnect = 1;
 
    P(mutex);
    if (m_connected) {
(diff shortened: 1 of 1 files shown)


joergsteffens:
Thank for for your patch. We are currently busy to integrate Fedora 27 into our build environment. As soon as this has happend, we will care for this bug.

chipitsine:
I also noted that libwrap is not specified as a dependency (which stops building on Fedora 27

tuxmaster:
Tested the patch on RHEL7, Fedora 26 and Fedora 27.
Compile + works on all three.

PR #67, opened 2017-12-11
Merge stable fixes into 16.2 branch

mvwieringen:
This PR merges all stable fixes needed to 16.2 to make GlusterFS and CephFS backups and restores usable in the first places. Until now restores for both glusterfs and cephfs always reopened the handle
to the API for every file which on glusterfs leads to running out of sockets pretty quickly as GFAPI opens
a new set of sockets for every glfs_init() call being done. It also fixes some typos like printing the wrong
variable (e.g. printing the value <NULL>) and parse blocksizes always using CFG_TYPESIZE32 so people can use things like 2 Mb everywhere and not sometimes have to use 2097152 to get the same
thing done.


src/include/version.h
```diff
@@ -4,13 +4,16 @@
 #define LSMDATE "09Oct17"
 
 #define PROG_COPYRIGHT "Copyright (C) %d-2012 Free Software Foundation Europe e.V.\n" \
+                       "Copyright (C) 2010-2017 Planets Communications B.V.\n" \
```
joergsteffens: Can you explain the legal implication of this? Of course, I know you have been working at the Bareos code, also in the years 2010-2012. However, the commit dates start at 2013 (with your Bareos email address).
I'm especially unsure about the years 2010-2012, in which the copyright have been assigned to the FSFE. Does it align with copyright law to later claim the copyright?

mvwieringen: Its quite simple FSFE only got the patches which got upstreamed to Bacula as that is what
the FLA says. So all patches for BAREOS are my copyright and my copyright only for 2010, 2011 and 2012. Further more you keep also copyright yourself as the FLA reassigns the contributor to also license his code anyway he likes. But the most important is that all reassignment ONLY applies to things explicitly upstreamed to Bacula. If that would be true
anything I would have been doing even for my company would have its copyright transfered.
That is of course non-sense. B.T.W. the code already had this copyright claim only the generic
header not and the program output.

PR #71, opened 2017-12-21
Use the rpm macro for name instead of bareos.

tuxmaster:


joergsteffens:
Hello,
thank you for your contribution. Of course, this can be done. But want is the actual benefit of changing this now? Do you want to have multiple versions of Bareos installed in parallel? Even with this patch, there will be conflicts.
regards,
Jörg

tuxmaster:
Multiple versions will fail in both cases, because the files will placed under the same path.
When using 2 different names, for the "name" macro, then multiple versions are possible.
For example bareos and bereos-new.
But for now, it make the spec file much easer to read.

joergsteffens:
If you provide your real name, I can add it to the Bareos authors.

PR #72, opened 2017-12-21
Migrate the tray monitor to Qt5

tuxmaster:
Testet on RHEL 7 + Fedora 26
RHEL 6 was left out, because it will only comes with Qt4


tuxmaster:
5.6 comes with RHEL7 and Fedora use 5.9.
So the lowest version which I can test for rpm packages are 5.6.

joergsteffens:
Do you see a possibility to make this work with both qt4 and qt5?

Another think I see are a lot of changes due to changed indentation. If the indentation would not change, the number of changes would be much less. Please have a look at http://doc.bareos.org/master/html/bareos-developer-guide.html#indenting-standards.
Maybe the commands
`indent -nbad -bap -bbo -nbc -br -brs -c36 -cd36 -ncdb -ce -ci3 -cli0 -cp36 -d0 -di1 -ndj -nfc1 -nfca -hnl -i3 -ip0 -l85 -lp -npcs -nprs -npsl -saf -sai -saw -nsob -nss -nbc -ncs -nbfda --no-tabs -T POOL_MEM -T json_t` will realign the code.

tuxmaster:
Not really without massive use of "#ifdef..."
But on the Qt page I must read, that the 4.8 tree are out of support since 2015.
https://blog.qt.io/blog/2015/05/26/qt-4-8-7-released/

So I think the next major version of bareos will be the target for the patch.

franku:
Meanwhile we could switch our build system to cmake and we were able to provide both builds, Qt4 and Qt5 depending on the distribution with much less effort. Anyway, thank you so much for your contribution.

PR #73, opened 2017-12-21
Allow compiling with LibreSSL

mskuta:
If you use LibreSSL instead of OpenSSL, building fails in two places because LibreSSL still corresponds to the OpenSSL-1.0.1 API.

This fix was tested with LibreSSL versions 2.2.2 (2015-08-05) and 2.6.4 (2017-12-19).

### DIFF
--- a/src/lib/crypto_openssl.c
+++ b/src/lib/crypto_openssl.c
@@ -128,7 +128,7 @@ IMPLEMENT_ASN1_FUNCTIONS(RecipientInfo)
 IMPLEMENT_ASN1_FUNCTIONS(SignatureData)
 IMPLEMENT_ASN1_FUNCTIONS(CryptoData)
 
-#if OPENSSL_VERSION_NUMBER < 0x10100000L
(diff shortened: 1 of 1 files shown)


joergsteffens:
Thank you for this Pull Request. The changes should not hurt our build. However, I've got a few questions:
  * Does LIBRESSL_VERSION_NUMBER somehow correspond to the OPENSSL_VERSION_NUMBER?
  * Does LibreSSL intend to upgrade their API to correspond to newer openssl versions?
  * We use OPENSSL_VERSION_NUMBER checks in more than the two places where you added the LIBRESSL_VERSION_NUMBER check. Does bareos compiled against LibreSSL work? Or is this only a fix to allow it to compile?

mskuta:
* LIBRESSL_VERSION_NUMBER and OPENSSL_VERSION_NUMBER are independent of each other. See: https://github.com/libressl-portable/openbsd/blob/master/src/lib/libcrypto/opensslv.h

* I don't know about that.

* I have performed and restored backups with data encryption with my patched version. Everything went well.

joergsteffens:
Thank you for the patch. I added it to our internal system. As long as no errors come up, it will be published soon.

PR #74, opened 2018-01-01
remove duplicate index identified by pt-duplicate-key-checker

chipitsine:
identified by Percona Toolkit:

```
pathvisibility_jobid is a left-prefix of PRIMARY
 Key definitions:
   KEY `pathvisibility_jobid` (`JobId`)
   PRIMARY KEY (`JobId`,`PathId`),
 Column types:
	  `jobid` int(11) not null
	  `pathid` int(11) not null
```

### DIFF
--- a/src/cats/ddl/creates/mysql.sql
+++ b/src/cats/ddl/creates/mysql.sql
@@ -369,8 +369,6 @@ CREATE TABLE PathVisibility
    Files int4 DEFAULT 0,
    CONSTRAINT pathvisibility_pkey PRIMARY KEY (JobId, PathId)
 );
-CREATE INDEX pathvisibility_jobid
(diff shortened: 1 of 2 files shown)


joergsteffens:
This change looks valid to me. @sduehr, can you please comment on this?

chipitsine:
There are might be similar duplicate indexes for Postgres as well. Someone who understands postgres better should have a look

sduehr:
Thank you for the hint. Indeed this index is useless.
I've run some tests to find out if instead adding an index on pathid would make any of .bvfs_update, .bvfs_lsdirs or .bvfs_lsfiles faster, but that is not the case.
Dropping the index with the update to the next Bareos version will be implemented together with other DB schema changes in PR #360

PR #75, opened 2018-01-11
Fix Univention bareos-dir-restart script

palstek:
Add missing absolute path to service command. Otherwise it fails with
/usr/share/univention-bareos/bareos-dir-restart: 3: /usr/share/univention-bareos/bareos-dir-restart: service: not found

### DIFF
--- a/platforms/univention/bareos-dir-restart
+++ b/platforms/univention/bareos-dir-restart
@@ -1,5 +1,5 @@
 #!/bin/sh
 
-su - bareos -s /bin/sh -c "/usr/sbin/bareos-dir -t" && service bareos-dir restart
+su - bareos -s /bin/sh -c "/usr/sbin/bareos-dir -t" && /usr/sbin/service bareos-dir restart
 exit $?
 


joergsteffens:
Where did this problem occur? I tested it again with UCS 4.2, and I don't see a problem with this. Configuration gets reloaded as expected.
I agree, that this change would not hurt, but I don't see it is required (in my installation).

palstek:
I got those on the very same UCS DC slave after each Bareos update until I fixed it manually:
Bareos Upgrade from 15.2.2-46.1 to 16.2.4-12.1 under UCS 4.0.
Bareos Update from 16.2.4-12.1 to 16.2.5-18.1 under UCS 4.0 which was later at the day upgraded to 4.2-3.
Bareos Upgrade from 16.2.5-21.2 to 17.2.4-9.1 under UCS 4.2-3.

bareos-dir-restart is called by Cron on 20:30 every day (/etc/cron.d/univention-bareos, which has been created by very first Bareos install from UCS app center and never touched since then). May there be a difference, because Bareos director is running on a DC slave instead of a DC master?

joergsteffens:
Thank you for your patch. I missed, that this script is also called by cron. I merged it to master (in our internal system). It will be published to github soon. I integrated your patch also in bareos-17.2 branch.

PR #76, opened 2018-01-24
BUG: Diff backup does not backup files changed since last Full backup

vermaden:
I wasn't able to find 'ISSUES' tab so I paste it here.

Bareos seems to have BUG when it comes to omitting files/dirs in DIFF/INC backups.

Here are the details:
https://groups.google.com/forum/#!topic/bareos-users/qJDL2fsfV9E

Regards.


davidegiunchidiennea:
@vermaden this is not a bug, it's the normal behaviour of the incr/diff backup without the accurate mode.
http://doc.bareos.org/master/html/bareos-manual-main-reference.html#directiveDirJobAccurate
If you want to include renamed file in the incr/diff, you have to enable accurate mode.

pstorz:
Hallo,

bareos works with timestamps to determine whether something is new when it is not working in accurate mode.
When you move files the timestamps are not updated and so not backed up.

vermaden:
Thank You for explanation, Accurate option solves the problem.

PR #77, opened 2018-04-02
Correctly configure LC_COLLATE and LC_CTYPE for PostgreSQL 10

dne:


joergsteffens:
Thank you for your patch. I also added it to the bareos-17.2 branch.

PR #80, opened 2018-04-16
change names of lex tokens T_* to BCT_* avoid conflict with python

rivimey:


joergsteffens:
Merged into our internal master. It is planed to push it in 2 weeks.

PR #81, opened 2018-04-16
annotating dev.h struct members not used, and spool device creation

rivimey:
This change is not useful in itself: nothing’s fixed. Rather, someone more familiar with the code should check.

- CLion inspections noted a number of functions not defined and/or not used, and this seemed worth noting with comments.

- spool.cc makes a new DEVICE without letting the compiler set up vtables and also without calling m_init_dev(); this seems to me to be bad.


pstorz:
Thank you for the information provided with this pull request.

We have quite a lot of those problems where objects are treated like structs, and we are working on
treating objects like objects with correct initialization etc.

Best regards,

Philipp

PR #85, opened 2018-04-16
rpc/xdr function arg missing; avoid warning about little endian redef

rivimey:


joergsteffens:
Merged into our internal master. It is planed to push it in 2 weeks.

PR #86, opened 2018-05-03
Fix link to documentation and bconsole typo

thraizz:
First commit fixes #727: wrong error message in disable schedule
Description: Disabling a non-existent schedule over bconsole no longer
throws "Client not found" error, now it is "Schedule not found"

Second commit fixes a documentation link mentioned [here](https://groups.google.com/forum/#!topic/bareos-users/0EIy1hCJcac), where a fuzzy directory structure would throw a non-existent link.

### DIFF
--- a/core/src/dird/ua_cmds.cc
+++ b/core/src/dird/ua_cmds.cc
@@ -1030,7 +1030,7 @@ static void DoEnDisableCmd(UaContext *ua, bool setting)
       }
 
       if (!sched) {
-         ua->ErrorMsg(_("Client \"%s\" not found.\n"), ua->argv[i]);
+         ua->ErrorMsg(_("Schedule \"%s\" not found.\n"), ua->argv[i]);
          return;
       }
    } else {


thraizz:
Done, sorry for the delay. Also rewrote the commit message so that the fix is more obvious.

joergsteffens:
Unfortunately, this commit message is still not like requested.
If this references a https://bugs.bareos.org ticket, it should look like:

```
    bugfix: prevents file corruptions by SHA1
    
    Fixes a bug when using SHA1 file signatures in Bareos version compiled
    without openssl.
    
    Fixes #838: File corruption with SHA1 signature

```

This way, the bug in https://bugs.bareos.org is automatically closed and refers to the change.

Please adapt your commit message.

As the source have been moved to the core subdirectory, please rebase your commit onto master.

thraizz:
I rebased and changed both the commit message as well as the bugfix itself so that it matches the new style how error messages are written in ua_cmds.cc are written.

PR #87, opened 2018-05-04
Fix ambiguous sample queries

dne:
When the JobBytes column was added to the JobMedia table (in commit
d31cd38e59f979e256370268d679d6f281ec758c), some sample queries became
ambiguous.


joergsteffens:
Thank you very much for your contribution!

PR #88, opened 2018-05-17
fixed height in RunningJobs.phtml only shows 3 jobs

schenklklopfer:
Why does this panel had an fixed height?
Only 3 jobs fitted in there.
If you run more than 3 jobs you need to scroll...

I'd rather scroll the whole page than just the box.


fbergkemper:
We do have some kind of other idea here, so we don't like to merge this PR.
But thanks anyways as this shows and confirms that it needs a kind of
improvement in that place.

PR #89, opened 2018-05-19
added Hungarian translation

Gabrov:


joergsteffens:
Thank you for your contribution. I just merged your changes into master. It will also be part of bareos >= 18.2.

PR #91, opened 2018-06-14
Merge bareos-16.2-droplet into bareos-17.2

joergsteffens:
Merges the droplet changes from bareos-16.2-droplet into bareos-17.2

### DIFF
--- a/README.droplet
+++ b/README.droplet
@@ -2,52 +2,57 @@ Using droplet S3 as a backingstore for backups.
 
 The droplet S3 storage backend writes chunks of data in an S3 bucket.
 
-For this you need to install the libdroplet-devel and the storage-droplet packages which contains
-the libbareossd-chunked*.so  and  libbareossd-droplet*.so shared objects and the droplet storage backend which implements a dynamic loaded
-storage backend.
+For this you need to install the the bareos-storage-droplet packages which contains
+the libbareossd-chunked*.so  and  libbareossd-droplet*.so shared objects and the droplet storage backend which implements a dynamic loaded storage backend.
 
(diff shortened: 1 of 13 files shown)


joergsteffens:
* autoconf/configure.in has the ceph_statx.h check now two times. We only need this check once

I'll fix this.


  * msgchan.c: why was the error changed to M_FATAL?
   If we are sure that M_FATAL is correct, we should remove the above comment.

Well, I'm not 100% sure, that why I requested the review. However, I can confirm, that this is required for the droplet backend. If not, a Job can finish with status JS_Warnings, without having all data (or none att all) written  to the Storage Daemon.

  * acquire.c: the job flush should only be called if the device really needs a flushing, otherwise people will be confused of the job messages.

Is it okay, to rephrase this to "releasing device"? Because that is what is done on all backends. On some, this will be fast, on others (droplet) this can take longer.

  * block.c: why was errno = EIO removed?

See commit 346907ff228e1d9ac9fc30947d493cb01704aee0:

    write_block_to_dev: don't retry on EIO, only on EBUSY
    
    EIO (io error) is normally permanent, so a retry will not help.
    When doing a retry on a EIO on the droplet_device,
    this results in lost data (because of the chunked_device caching).
    This could be fixed, however this is the quicker solution.

Retrying on EIO will only work as expected, when NO data has been written. If part of the data has been written successfully and the "cursor" in the backend has already moved on, rewriting the full data block  will result in having the already data duplicated  and therefore a corrupted Bareos block .
I've not found out, if a EIO should guarantee that no data has been written. At least, for the droplet backend, this is not the case.

As the original author have already been unsure if this code segment is useful ("feeble attempt"), I opt for removing it.

PR #95, opened 2018-07-10
webui: rename selenium test, fix exception import

thraizz:
Renames the selenium test to fix the import of exceptions in the same file. Python does not allow '-' dashes in import names.

Also enhances the test for available languages in login.
We used to test against a list of expected languages, now we iterate over the directories in public/js/locale and compare that counter against the present options while logging in.

### DIFF
--- a/core/platforms/win32/winbareos-nsi.spec
+++ b/core/platforms/win32/winbareos-nsi.spec
@@ -212,7 +212,7 @@ for flavor in %{flavors}; do
       cp /etc/bareos-webui/*.ini install
       cp -av /etc/bareos install
       mkdir -p tests/selenium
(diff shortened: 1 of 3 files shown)


joergsteffens:
This PR does not make sense.
Why importing your current file??? Instead you should just move the Exception classes above the  WebuiSeleniumTest class (as it has been before).

test_languages will never work in our testing environment, as the webui server  is not guaranteed to run on the same host and especially not at this path! Why have you not tested this in the test environment???

From prior commits: the get_env() method has never been intended to be inside the WebuiSeleniumTest. This is why it addresses attributes by Class and not by instance. Also the logic of reading environment variables does not belong inside WebuiSeleniumTest. Please move it again outside the class.

Please start from scratch.

PR #97, opened 2018-07-19
travis: build and test webui as new build environment

thraizz:
- Modify travis files to build the webui as new environment
- Modify selenium test to test the webui over saucelabs if built over travis
- Clean dead code in webui

1. Travis CI builds the core and webui components over multiple build.
2. The separated webui build executes a proxy service for Saucelabs, called Sauce Connect. 
3. It also executes Selenium Tests. These selenium tests don't open a browser instance for executing commands on the webui, they open a remote connection that handles the commands.
4. The remote connection goes out of the Travis build and into Saucelabs, which spawns browser instances that connect into the travis build over the open proxy.

[How it works as a diagram](https://s3.eu-central-1.amazonaws.com/thraizzstudium/travis.png)
[Latest build in private travis](https://travis-ci.org/thraizz/bareos/builds/405767917)
[Example webui test in Saucelabs](https://saucelabs.com/beta/tests/9ad3a8bcdb994ac284ef491d5521c99b)


webui/tests/selenium/webui-selenium-test.py
```diff
@@ -109,79 +108,77 @@ def setUp(self):
...
             raise ClientStatusException(self.client, 'enabled')
         self.logout()
 
-    def disabled_test_job_canceling(self):
-        driver = self.driver
-        self.login()
-        job_id = self.job_start_configured()
-        self.job_cancel(job_id)
-        self.logout()
-
```
thraizz: I missed this in the git diff, I will add this back into the pull request.

webui/tests/selenium/webui-selenium-test.py
```diff
@@ -330,6 +312,16 @@ def logout(self):
...
+    def getChromedriverpath(self):
+        # On OS X: Chromedriver path is 'usr/local/lib/chromium-browser/chromedriver'
+        for chromedriverpath in {'/usr/lib/chromium-browser/chromedriver', '/usr/local/lib/chromium-browser/chromedriver'}:
+            try:
+                os.path.isfile(chromedriverpath)
+            except FileNotFoundError:
+                pass
+            else:
+                return chromedriverpath
+
```
thraizz: This new method searches for the right path where the chromedriver is located. On os x, `/usr/lib/` is reserved for system access only. `/usr/local/lib` is the right path.

webui/tests/selenium/webui-selenium-test.py
```diff
@@ -410,8 +392,20 @@ def close_alert_and_get_its_text(self, accept=True):
...
+                if sys.exc_info() == (None, None, None):
+                    sauce_client.jobs.update_job(self.driver.session_id, passed=True)
+                else:
+                    sauce_client.jobs.update_job(self.driver.session_id, passed=False)
+            finally:
+                self.driver.quit()
+                self.assertEqual([], self.verificationErrors)
+        else:
+            self.driver.quit()
+            self.assertEqual([], self.verificationErrors)
```
thraizz: `sys.exc_info()` is pythons method to check if exceptions were thrown. If there were none we pass the test success to saucelabs by using the sauceclient module.

webui/tests/selenium/webui-selenium-test.py
```diff
@@ -443,6 +437,12 @@ def get_env():
     sleeptime = os.environ.get('BAREOS_DELAY')
     if sleeptime:
         SeleniumTest.sleeptime = float(sleeptime)
+    if(os.environ.get('TRAVIS') == 'true'):
+            SeleniumTest.browser = 'none'
+            SeleniumTest.username = 'citest'
+            SeleniumTest.password = 'citestpass'
+            SeleniumTest.sauce_username = os.environ.get('SAUCE_USERNAME')
+            SeleniumTest.access_key = os.environ.get('SAUCE_ACCESS_KEY')
```
thraizz: Set the test variables different if we are currently in a travis build.
`SAUCE_USERNAME` and `SAUCE_ACCESS_KEY` need to be set in Travis' build settings so that they will not be exposed.

.travis/travis_before_install.sh
```diff
@@ -1,12 +1,20 @@
 #!/bin/bash
 
 sudo apt-get -qq update
+sudo pip install selenium sauceclient
```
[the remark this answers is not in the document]
thraizz: No, because they are python modules. The installation and binding handling is always handled by [pip](https://pypi.org/project/pip/).

.travis.yml
```diff
@@ -37,9 +37,16 @@ matrix:
     - env: DB=mysql
     - env: DB=sqlite3
     - env: DB=postgresql COVERITY_SCAN=1
+    - env: DB=postgresql BUILD_WEBUI=true
+      addons:
+        sauce_connect: true
+        apt:
+          packages:
+            - libjansson-dev
```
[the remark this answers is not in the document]
thraizz: Into the dependencies handed out over downloads.bareos.org ? Or in the general travis build dependencies ? If you meant the first, I'll ask tomorrow how to do so.

joergsteffens:
Well, at least https://packages.ubuntu.com/trusty/python-selenium is available as Deb package.

Am 19. Juli 2018 22:31:16 MESZ schrieb "A. Schueler" <notifications@github.com>:

joergsteffens:
https://github.com/bareos/bareos/blob/master/core/debian/control and https://github.com/bareos/bareos/blob/master/core/debian/control.in

Am 19. Juli 2018 22:38:55 MESZ schrieb "A. Schueler" <notifications@github.com>:

thraizz:
I will need to revert the installation process of selenium back to installation via pip, as the apt package does not contain selenium.webdriver components. We rely heavily on those, for example all expected conditions are imported out of those components. This is also documented here:
https://packages.debian.org/en/sid/python-selenium

thraizz:
- Rebased on top of master
- Commit for webui build fix dropped as it was fixed
- Adds saucelabs badge into the readme

The readme currently contains the saucelabs badge two times. @joergsteffens, choose what looks better, @fbergkemper recommended to add it as a new sponsors section. Maybe we can put it into the `SPONSORS.md`?

README.md
```diff
@@ -1,6 +1,7 @@
 # <img src="https://raw.githubusercontent.com/bareos/bareos/master/webui/public/img/bareos.png" alt="Bareos" />
 
-[![Build Status](https://travis-ci.org/bareos/bareos.png?branch=master)](https://travis-ci.org/bareos/bareos/branches)
+
+[![Build Status](https://travis-ci.org/bareos/bareos.png?branch=master)](https://travis-ci.org/bareos/bareos/branches)<br><a href="https://www.saucelabs.com/"> <img src="webui/public/img/saucelabs_white_badge.svg" alt="Saucelabs"  width="90px"/></a>
```
joergsteffens: Is this the recommend way to handle this? Not using img src, that refers to an icon provided by sauce Lab? Also, wouldn't it be better to link to our sauce Lab test results?

thraizz: It was only sent to me via email, but I've found it hosted by them. The link will point to our sauce lab results as soon as they go live, I've noted this on my todo list.

joergsteffens:
Unfortunately, the selenium test do all fail: https://travis-ci.org/bareos/bareos/jobs/437528496
Do you have an explanation for this?

thraizz:
Python client can't use urllib3 to parse our webcontent. As a solution, we should upgrade both pip (as we use v9.x when there's v18 available) and urllib3 to newest version. I'll do this in the 'travis_before_script' soon as I get home from university 👍

thraizz:
I needed to specify to use urllib3 at version 1.22 as python requests dont like versions above (latest is 1.23) or below:
`requests 2.18.4 has requirement urllib3<1.23,>=1.21.1, but you'll have urllib3 1.23 which is incompatible.`

thraizz:
This is a bit difficult to reproduce as my travis builds always fail without a paid saucelabs subscription, but I believe that the commit fixes the failures (and also just run `pip` commands when we're going to need `sauceclient` and `selenium`)
https://travis-ci.org/thraizz/bareos/jobs/438988236

joergsteffens:
Luckily your commits are tested by Travis / SauceLabs inside this Pull Request. It still fails, however, the error message has changed. Can you please take another look at it?
I wondering, what have changed as it succeeded Travis before in your branch, right?

thraizz:
This is due to security rules of travis, as stated [here](https://docs.travis-ci.com/user/environment-variables/#defining-encrypted-variables-in-travisyml):
> Encrypted variables are not available to untrusted builds such as pull requests coming from another repository.

joergsteffens:
I see. So I merged it now to check the status after merge.

thraizz:
The tests can be skipped if we add `@unittest.skipIf` followed by a nice condition, before the unittest class starts. The condition should check if the env var `TRAVIS_PULL_REQUEST` is set and not false, if both are true we will skip the test case and exit with 0. Something like

`@unittest.skipIf((os.environ.get('TRAVIS_PULL_REQUEST') != 'false' and os.environ.get('TRAVIS_PULL_REQUEST') != 'None'), 'Pull requests are not tested.')
` 
would work but doesn't look too good...

thraizz:
Documentation for skipping tests: https://docs.python.org/2/library/unittest.html#skipping-tests-and-expected-failures
Documentation of travis env vars: https://docs.travis-ci.com/user/environment-variables/#convenience-variables

joergsteffens:
One reason for the problems have been, that enviroment variables have not been set in our travis environment. It also needed the variable TRAVIS to be set. Now the sauce connect seams to work fine. 
However, another problem did arise, see https://travis-ci.org/bareos/bareos/jobs/439190518:

`AttributeError: 'SeleniumTest' object has no attribute 'driver'`

Can you please check on that?

thraizz:
As `TRAVIS` is a default environment value I thought I could depend on that to determine if the test was executed in a CI environment, seems like we can't. 
The test fails because all if-conditions for driver determination were false, so no driver was set. Locally the tests work absolutely fine - tested with travis env vars set and also just with `BAREOS_BROWSER=chrome`. You could investigate why `TRAVIS` env is not `true` or do e.g. `BUILD_WEBUI == true` instead, we set that env explicitly.

PR #103, opened 2018-08-06
Fix d'tor out by one error

gnif:
This corrects an out by one error that causes a crash at shutdown and at configuration reload. See bug #0000996


gnif:
Closing as invalid, not sure where my head was when I was working through this.

PR #104, opened 2018-08-06
fix triplicate free of tls_cert.dhfile

gnif:


gnif:
This patch is not complete, see #105

PR #105, opened 2018-08-06
fix crashed caused by triplicate freeing of dhfile

gnif:
Replaces #104 as it was incomplete


franku:
However, the mentioned problem applies to all these files: 
- core/src/dird/dird_conf.cc (this one is referenced in the pull request)
- core/src/stored/stored_conf.cc (not referenced in this pull request)
- core/src/filed/filed_conf.cc (not referenced in this pull request)
- core/src/console/console_conf.cc (not referenced in this pull request)

franku:
Will rework all files manually, rather than applying this pull request directly. The source code in master has changed significantly since you handed in your changes. Thank your very much for your contribution. 

Commit f1e4e643cccc8512ef4d96ec7603350ca06e3dc6 contains all relevant changes.

PR #110, opened 2018-08-29
pass a IPv6 capable sized struct to accept - broken on x64_win32

NixM0nk3y:
Patch to fix - https://bugs.bareos.org/view.php?id=477

accept on windows failed to accept a IPv6 connection as the sockaddr struct wasn't large enough.  Moved to recommended sockaddr_storage struct.

Patch resolves issues in Windows 7 x64.

### DIFF
--- a/core/src/lib/bnet_server_tcp.cc
+++ b/core/src/lib/bnet_server_tcp.cc
@@ -154,7 +154,7 @@ void BnetThreadServerTcp(
 {
   int newsockfd, status;
   socklen_t clilen;
-  struct sockaddr cli_addr; /* client's address */
+  struct sockaddr_storage cli_addr; /* client's address */
   int tlog;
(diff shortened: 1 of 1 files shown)


arogge:
The `struct sockaddr_storage` is the larger brother of `struct sockaddr` that can contain both `struct sockaddr_in` and `struct sockaddr_in6` so this is basically a good thing to do, especially if we want to support IPv6 on Windows.
However, the c-style casts are bad. I guess this should be a reinterpret_cast<>()

NixM0nk3y:
Hi , thanks for the comments - rebased the pull request on current master and switched to reinterpret_cast

arogge:
Thanks. That is really appreciated!
This PR is on my TODO list for this week, so I hope we can merge it before Friday.

PR #112, opened 2018-09-07
Dev/joergs/bareos 16.2 droplet/truncate volumes

joergsteffens:
Fixes a problem on truncating S3 volumes via droplet.


pstorz:
Changes look reasonable to me. Agreed.

PR #114, opened 2018-09-19
Revert rename of python "create_file"

gnif:
The python entry point `create_file` has been incorrectly renamed to `CreateFile`, this will break existing plugins and is not consistent with the naming convention of other python entry points. This PR simply reverts this change.


joergsteffens:
Thank you for pointing this out.

PR #118, opened 2018-09-21
core: include psql parameters in database check

thraizz:
Fixes warning when we created a database with `--host` or `--port` parameters that are not the default, as they were ignored in line 130 psql call. This led to checking only local databases for the encoding. Now, possible parameters are respected.

### DIFF
--- a/core/src/cats/create_bareos_database.in
+++ b/core/src/cats/create_bareos_database.in
@@ -127,7 +127,7 @@ ALTER DATABASE ${db_name} SET datestyle TO 'ISO, YMD';
 END-OF-DATA
       retval=$?
 
-      if psql -l ${dbname} | grep " ${db_name}.*SQL_ASCII" >/dev/null; then
+      if psql -l ${dbname} $* | grep " ${db_name}.*SQL_ASCII" >/dev/null; then
          echo "Database encoding OK"
       else
          echo " "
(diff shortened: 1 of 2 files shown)


joergsteffens:
What does this fix help, as you pass the parameter only to one occurrence of the psql command. The the remaining psql command will connect against a local database, right? 
Have you tested your change? Have you been able to initiate a Bareos database on a remote host? I expect, username and password would also be required, right?

I can see, that this can be implemented reliable this easily.

thraizz:
Yes, I have been able to set up the PostgreSQL database by using the scripts (`make_bareos_database`,`create_bareos_tables` and `grant_bareos_privileges`) by calling them with `--host` and `--username` parameters. The password is promted afterwards for each psql command.

The only thing that did not work was the encoding check in the `make_bareos_database` script, this pull request patches that by passing all trailing parameters to the psql command. This is also done in line 98 and 123 by appending `$*` to the psql command.

thraizz:
The commit 48a2915 implements @joergsteffens suggestion to always pass `-h` option for every psql call. Without a defined address in the catalog resource we pass `localhost` as parameter.

thraizz:
Passing `-h` to a local psql server will break the script functionality and is therefore removed - there is a difference in a `psql -U bareos` and a `psql -h localhost -U bareos` script. 56a0715 reverts those calls to default.

30264cb adds documentation about configuring a remote psql system with dbconfig and/or the database scripts.

joergsteffens:
Thanks. Could you squash the commit together into one commit? Is there a bug ticket you are solving with this? If yes, could you please mention it in the commit message?

thraizz:
Done 👍

joergsteffens:
Thank you!

PR #119, opened 2018-09-22
Fix failure to send RestoreObject for Jobs

gnif:
When using plugins such as Percona Xtrabackup restoration fails due to failure to send the RestoreObjects for the related jobs. This patch corrects this.

### DIFF
--- a/core/src/dird/ua_restore.cc
+++ b/core/src/dird/ua_restore.cc
@@ -1552,6 +1552,16 @@ static int JobidFileindexHandler(void *ctx, int num_fields, char **row)
    AddFindex(rx->bsr, rx->JobId, str_to_int64(row[1]));
    rx->found = true;
    rx->selected_files++;
+   
+   if (bstrcmp(rx->last_jobid, row[0])) {
+      return 0;                       /* duplicate id */
+   }
+   bstrncpy(rx->last_jobid, row[0], sizeof(rx->last_jobid));
+   if (rx->JobIds[0] != 0) {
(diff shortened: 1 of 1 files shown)


aussendorf:
Hi,
thanks for the explanation, it's the same with the VMWare plugin and related to https://bugs.bareos.org/view.php?id=805
We will test the patch (probably next week) to make sure it does not have any side effects.
Regards
Maik

pstorz:
Hello,

thanks for your contribution. 

I have looked into what you did, and it seems that you have copied the content of 
JobidHandler() into JobidFileindexHandler().

I would like to  suggest to not repeat yourself and do the following:

--------------------------------------------------------------------------------------------------
```
diff --git a/core/src/dird/ua_restore.cc b/core/src/dird/ua_restore.cc
index 23946071b..457a924ae 100644
--- a/core/src/dird/ua_restore.cc
+++ b/core/src/dird/ua_restore.cc
@@ -1556,6 +1556,9 @@ static int JobidFileindexHandler(void *ctx, int num_fields, char **row)
    AddFindex(rx->bsr, rx->JobId, str_to_int64(row[1]));
    rx->found = true;
    rx->selected_files++;
+
+   JobidHandler(ctx, num_fields, row);
+
    return 0;
 }
```
--------------------------------------------------------------------------------------------------

What do you think?


Best regards,


Philipp

gnif:
Sure, but IMO either is fine since it's such a small amount of code. I will try to find time to update this PR but at this time my schedule is very full.

PR #120, opened 2018-09-25
webui: add remember me functionality for login

thraizz:
Fixes #956: Add Remember Me feature to webui login


fbergkemper:
Rejected because it is not persistent.

PR #124, opened 2018-10-17
travis/webui: improve dependance on env vars

thraizz:
This is an addition to #97, it makes the travis test depend on an environment variable we set.
Also raise an error if we try to continue the setup without a driver.


joergsteffens:
These distinction is already covered by 63c21e2b311ab70401348c15b51b95b87c90e754

PR #125, opened 2018-10-17
Somes fixes to the FD gluster plugin

Arakmar:
- Removal of the basedir option : The basedir option is conflicting (and overwritten) with the dir option already present in the gluster URI. It also causes segfault at the end of each backups

- Fix for unix sockets : If we don't define a sub directory when we use an unix socket, it fails to parse the URI because the first '/' in the socket path is used as the subdirectory.

### DIFF
--- a/core/src/plugins/filed/gfapi-fd.cc
+++ b/core/src/plugins/filed/gfapi-fd.cc
@@ -1432,11 +1432,19 @@ static inline bool parse_gfapi_devicename(char* devicename,
 
       /*
        * See if there is a dir specified.
+       * As we use an unix socket, we need to check if we are not using the
+       * path specified for the unix socket.
+       * It can happen if there is no subdirectory in the URI
        */
-      bp = strchr(bp, '/');
-      if (bp) {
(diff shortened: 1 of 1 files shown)


core/src/plugins/filed/gfapi-fd.cc
```diff
@@ -1482,11 +1473,19 @@ static inline bool parse_gfapi_devicename(char *devicename,
...
-      if (bp) {
+      char *before_parameter = strchr(bp, '?');
+      char *before_dir = strchr(bp, '/');
+
+      if (before_dir && (!before_parameter || before_parameter > before_dir)) {
+         bp = before_dir;
          *bp++ = '\0';
          *dir = bp;
+      } else {
+         *dir = 0;
```
arogge: I think you're meaning nullptr instead of 0
```suggestion
         *dir = nullptr;
```

arogge:
Hi,
would you mind rebasing your changes to the latest master so we can finally review and merge them?
All in all this looks reasonable and I would like to merge it.

Best Regards,
Andreas

PR #131, opened 2018-11-01
Fixed error installation on non standard postgres db port

JayDi85:
Installation script are missing from `PGPORT` environment setup. That's why it's can't install bareos on systems with non standard postgres port. User gets error about connection problem and can't continue to install. That's PR will fix that error.


arogge:
Hi Oleg,
thank you very much for your pull-request.
The change looks good. Did you rebuild the installer to actually try it out?

Best Regards,
Andreas

JayDi85:
No, I don't rebuild it. But I tested workaround to manually set PGPORT environment variable before install -- it's works correct.

PR #132, opened 2018-11-01
Fixed infinite intall dialog for VC 2012 checks on x86 windows

JayDi85:
`Visual C++ Redistributable for Visual Studio 2012` checks in install wizards is wrong. It's broken on x86 windows (32 bit) -- users gets infinite install dialog. It's work for x64 windows only.

Error was in wrong reg paths -- there are different on x86 and x64 windows.

That's PR will fix that error and x86 will works fine too. Info about reg values can founded [here](https://stackoverflow.com/a/14878248/1276632).

### DIFF
--- a/core/platforms/win32/winbareos.nsi
+++ b/core/platforms/win32/winbareos.nsi
@@ -975,20 +975,26 @@ Section "Bareos Webui" SEC_WEBUI
    File /r "bareos-webui"
 
 IfSilent skip_vc_redist_check
-   # check  for Visual C++ Redistributable für Visual Studio 2012 x86 (32 Bit)
+   # check  for Visual C++ Redistributable für Visual Studio 2012 x86 (on 32 and 64 bit systems)
    ReadRegDword $R1 HKLM "SOFTWARE\Wow6432Node\Microsoft\VisualStudio\11.0\VC\Runtimes\x86" "Installed"
+   ReadRegDword $R2 HKLM "SOFTWARE\Microsoft\VisualStudio\11.0\VC\Runtimes\x86" "Installed"
 check_for_vc_redist:
    ${If} $R1 == ""
-      ExecShell "open" "https://www.microsoft.com/en-us/download/details.aspx?id=30679"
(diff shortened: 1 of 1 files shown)


pstorz:
Hello,

thank you for your contribution.

Can you please give us a bit more information about your test environment?
On which OS are you exactly testing?

Best regards,

Philipp

JayDi85:
Test it on Windows 7 x86 edition.

core/platforms/win32/winbareos.nsi
```diff
@@ -975,20 +975,26 @@ Section "Bareos Webui" SEC_WEBUI
...
-      ExecShell "open" "https://www.microsoft.com/en-us/download/details.aspx?id=30679"
-      MessageBox MB_OK|MB_ICONSTOP "Visual C++ Redistributable for Visual Studio 2012 x86 was not found$\r$\n\
+      ${If} $R2 == ""
+         ExecShell "open" "https://www.microsoft.com/en-us/download/details.aspx?id=30679"
+         MessageBox MB_OK|MB_ICONSTOP "Visual C++ Redistributable for Visual Studio 2012 x86 was not found$\r$\n\
                                  It is needed by the bareos-webui service.$\r$\n\
                                  Please install vcredist_x86.exe from $\r$\n\
                                  https://www.microsoft.com/en-us/download/details.aspx?id=30679$\r$\n\
                                  and click OK when done." /SD IDOK
+      ${EndIf}
```
pstorz: This line has many trailing spaces. Can you please remove them?

pstorz:
Thank you for the quick answer. I could replicate the  problem on Win10/32. 
Can you please change the commit to remove the trailing spaces after the "EndIf", I also commented that directly here in github.


I will run some internal tests with the commit to see if everything still works.

Thank you very much.

JayDi85:
Trailing spaces removed.

pstorz:
Than you very much for your contribution, I have verified on 32 and 64Bit Windows 10 and everything works.

I have merged the two commits into one and added your name to the AUTHORS file and commited both into master.

Thank you very much!

PR #133, opened 2018-12-06
add missing leading slash to logdir

mschiff:
This comment add a leading slash to the logdir-Path that is currently missing (cmake output will list only "var/log/bareos" instead of "/var/log/bareos" without the patch

### DIFF
--- a/core/cmake/BareosSetVariableDefaults.cmake
+++ b/core/cmake/BareosSetVariableDefaults.cmake
@@ -121,7 +121,7 @@ ENDIF()
 
 # logdir
 IF(NOT DEFINED logdir)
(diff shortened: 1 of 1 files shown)


mschiff:
> Hello,
> 
> please refer to the documentatin of the cmake variables:
> https://cmake.org/cmake/help/v3.13/module/GNUInstallDirs.html
> 
> Your change does not make sense in my view.
> Did it fix any problem or is it just a cosmetic change?

cmake shows an overview of paths at the end of its run. There was one slash missing in it and this PR fixes it.

pstorz:
Ok, so it is only a cosmetic change and as I don't think that it is required I will not accept this PR.

Thanks anyway!

PR #134, opened 2018-12-23
docs: improve LaTex to Sphinx conversion

joergsteffens:
Add special post pandoc conversion script "latex-scan.py",
that cope better with LaTex contexts then the perl regexes.
The script "post_conversion_changes.sh" is no longer called.

index are now inline markups, instead of extra blocks.
This matches the current use of index better.
However, this required a text, that will be displayed and indexed.
We therefore write the index as [TAG={0}->{1}->{2}].
This can be manually replaced later.

The conversion of tables do not work in a lot of cases
and therefore the converted tables are not shown at all.
We cope with this by doubling the tables:
1: original LaTex table as a raw block,
2: table converted to Sphinx/rst. Probaly not shown or shown with errors.

Also done a lot of smaller fixes.


pstorz:
Hello Jörg,

this looks very good.

I had only one problem in win32.tex:333 where pandoc did not convert beacause of the % signs in the URL. Manually escaping them to \% helped here.

Did you not have this problem?

pstorz:
I just see, that constructs like 
"*<input>wait jobid=1</input>" are not handled yet. In the previous scripts these markers were removed.

pstorz:
Great, the markup removed the important information
This is what I meant (the input markers)
*\<input\>wait jobid=1\</input\>

PR #137, opened 2019-01-15
stored: do not treat read error as EoT by default

arogge:
Fixes #1034: Read error on tape may be misinterpreted as end-of-tape
Previously stored treated a read-error immediately
following an end-of-file mark on a tape as end-of-tape
instead of an error.
This patch makes stored raise an error in this case,
but allows to switch back to the previous behaviour on
a per-device basis using the new configuration option
"Eof On Error Is Eot".


src/stored/stored_conf.c
```diff
@@ -205,6 +205,7 @@ static RES_ITEM dev_items[] = {
    { "AutoDeflateLevel", CFG_TYPE_PINT16, ITEM(res_dev.autodeflate_level), 0, CFG_ITEM_DEFAULT, "6", "13.4.0-", NULL },
    { "AutoInflate", CFG_TYPE_IODIRECTION, ITEM(res_dev.autoinflate), 0, 0, NULL, "13.4.0-", NULL },
    { "CollectStatistics", CFG_TYPE_BOOL, ITEM(res_dev.collectstats), 0, CFG_ITEM_DEFAULT, "true", NULL, NULL },
+   { "EofOnErrorIsEot", CFG_TYPE_BOOL, ITEM(res_dev.eof_on_error_is_eot), 0, CFG_ITEM_DEFAULT, NULL, NULL, NULL },
```
joergsteffens: The 7. parameter should contain the number, where this directive get introduced. "17.2.8-".
The 8. parameter should contain a short description about the directive. This will also been shown in the documentation.

franku:
Merged with Bareos-17.2 with commit 22b8eca5dadf9186a37449a804efaccb27de6ef7

PR #138, opened 2019-01-15
stored: do not treat read error as EoT by default

arogge:
Fixes #1034: Read error on tape may be misinterpreted as end-of-tape
Previously stored treated a read-error immediately
following an end-of-file mark on a tape as end-of-tape
instead of an error.
This patch makes stored raise an error in this case,
but allows to switch back to the previous behaviour on
a per-device basis using the new configuration option
"Eof On Error Is Eot".


franku:
Merged into Bareos 18.2 with commit c9f3c364c77c51a748783e581158585d339c33de

PR #140, opened 2019-01-22
Dev/stephand/bareos 18.2/vmware fixes utf8

sduehr:
Fixed VMWare plugin and documentation related to use of non-ascii characters.


pstorz:
branch was merged into bareos-18.2

PR #141, opened 2019-01-24
core: make sure jcr->stime is set correctly (17.2)

arogge:
Fixes #951: Differential backup compute date from last incremental
Previously job.c did not recalculate jcr->stime when upgrading an
incremental job to a differential job due to MaxDiffInterval, thus
backing up only files like an incremental job would do.
This patch sets jcr->stime to the start time of the last full or
differential job in this case.


arogge:
This needs work. I'll reopen a pull-req once I get it correct.

PR #145, opened 2019-02-05
webui: webui login problem with tls enabled

fbergkemper:
STREAM_CRYPTO_METHOD_TLSv1_2_CLIENT was introduced with PHP version
5.6.0. We need to care that calling stream_socket_enable_crypto method works
with versions < 5.6.0 as well.

Fixes #1045: webui login problem


franku:
merged into bareos-18.2 with commit 570d124e8f142fed0f8c5645f567155f2f19cc8b

PR #148, opened 2019-02-12
Preprocessor cleanup

arogge:
Preprocessor cleanup to remove dead code
===================================
This rather large set of patches removes lots of code that was either never compiled or hasn't been used for a long time.

Care has been taken so that there are no unintended functional changes in the code. The two known changes that are not just a change in (ignored) whitespace are the following:
- dird/scheduler.cc: removes empty function dump_job() and calls to it
- dird/ua_status.cc and stored/status.cc: remove indication of DEVELOPER_MODE from status message

How you can check the changes by yourselves
-----------------------------------------------------------
1. checkout master and this branch to different directories
2. fixup unstable preprocessor stuff like this:
```
find . \( -name '*.c' -o -name '*.cc' -o -name '*.h' \) -print0 |
xargs --null sed -i.bak -e '1i# line 0 "noname"
s/__LINE__/123/g;s/__FILE__/"FILE"/g'

find . -name 'label.cc' -print0 |
xargs --null sed -i.bak2 -e 's/__DATE__/"DATE"/g;s/__TIME__/"TIME"/g'
```
3. run `cmake` in top git directory with `-DCMAKE_EXPORT_COMPILE_COMMANDS=ON`
4. run `make -C core/src/ndmp` to generate the ndmp header files
5. run the following python script to run only the preprocess step of each compile:
```
#!/usr/bin/env python
from __future__ import print_function, division, unicode_literals
import json
from os import chdir, system

with open('compile_commands.json') as f:
    commands = json.loads(f.read())

for command in commands:
    chdir(command['directory'])
    system("%s -E" % (command['command']))
```
6. copy all generated ".o" files that contain the preprocessed sources to a different directory tree using `find . -name '*.o' -print0 | cpio -p --null --make-directories /target/path`
7. remove file/line information and empty lines from the preprocessed code using `find /tmp/$1 -type f -print0 | xargs --null sed -i -e '/^# /d;/^$/d;'`
8. compare both prepared preprocessed directory-trees using `diff -wur /path/to/master-output /path/to/mybranch-output`


arogge:
added another functional change in 1823c57 that removes the function Device::SetMutexPriorities()

arogge:
Removed commit 4088d36 that changed lmdb code. As this is not our code, we shouldn't change it.

PR #154, opened 2019-03-07
Add basic unit-testing for reservation in storage daemon

arogge:


core/src/tests/sd_reservation.cc
```diff
@@ -0,0 +1,308 @@
...
+using ::testing::DoAll;
+using ::testing::Return;
+
+template <typename T, typename... Args>
+std::unique_ptr<T> sm_make_unique(Args&&... args) {
+  return std::unique_ptr<T>(new ("unique_ptr", 123)
+                                T(std::forward<Args>(args)...));
+}
+
+namespace storagedaemon {
```
franku: using namespace storagedaemon;

PR #156, opened 2019-03-08
docs: make diagrams work with newer plantuml

arogge:
Previously the diagrams for docs.bareos.org have
been built using plantuml 8033 (which even
predates the modern versioning scheme). This patch
updates some of the diagrams to still look nice
when built using plantuml 1.2019.2 (which is the
latest version as of now).

Applying this pull-request should be coordinated with an upgrade of plantuml on docs.bareos.org. The new JAR is already in place, but the configuration has not been changed yet.


joergsteffens:
As i'm not involved in PlantUML images, I removed myself from the reviewers and assigned the ticket to @pstorz.

PR #160, opened 2019-03-15
dird: redesign GetNdmpEnvironmentString() API

arogge:
Fixes #1056: NDMP restore on 16.2.5 and above does not fill NDMP environment correctly

Previously one overload of the function GetNdmpEnvironmentString() wanted a JobDbRecord* and expected jr->VolSessionId and jr->VolSessionTime to contain the values for the volume from which the
restore happens. These had to be filled manually before calling GetNdmlEnvironmentString() which had not been done since 16.2.5 resulting in #1056.

This patch now redesigns the API for all overloads of GetNdmpEnviromentString() to make it harder
to misuse.

We also add a new struct VolumeSessionInfo to wrap a pair of VolumeSessionId and VolumeSessionTime. These two numbers are only meaningful together, so they now have their own container.

### DIFF
--- a/core/src/cats/cats.h
+++ b/core/src/cats/cats.h
@@ -35,8 +35,6 @@
 #ifndef BAREOS_CATS_CATS_H_
 #define BAREOS_CATS_CATS_H_ 1
 
-#include "lib/volume_session_info.h"
-
 /* import automatically generated SQL_QUERY_ENUM */
 #include "bdb_query_enum_class.h"
 
@@ -48,6 +46,8 @@
  */
 
 #define faddr_t long
+
+struct VolumeSessionInfo;
 
 /**
  * Generic definitions of list types, list handlers and result handlers.
@@ -823,7 +823,7 @@ class BareosDb
   int GetNdmpLevelMapping(JobControlRecord* jcr,
                           JobDbRecord* jr,
                           char* filesystem);
-  bool GetNdmpEnvironmentString(const VolumeSessionInfo vsi,
+  bool GetNdmpEnvironmentString(const VolumeSessionInfo& vsi,
                                 const int32_t FileIndex,
                                 DB_RESULT_HANDLER* ResultHandler,
                                 void* ctx);
--- a/core/src/cats/cats.h
+++ b/core/src/cats/cats.h
@@ -35,6 +35,8 @@
 #ifndef BAREOS_CATS_CATS_H_
 #define BAREOS_CATS_CATS_H_ 1
 
+#include "lib/volume_session_info.h"
+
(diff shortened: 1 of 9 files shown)


core/src/cats/cats.h
```diff
@@ -821,12 +823,15 @@ class BareosDb
   int GetNdmpLevelMapping(JobControlRecord* jcr,
                           JobDbRecord* jr,
                           char* filesystem);
-  bool GetNdmpEnvironmentString(JobControlRecord* jcr,
-                                JobDbRecord* jr,
+  bool GetNdmpEnvironmentString(const VolumeSessionInfo vsi,
```
franku:
```suggestion
  bool GetNdmpEnvironmentString(const VolumeSessionInfo& vsi,
```

core/src/cats/cats.h
```diff
@@ -35,6 +35,8 @@
 #ifndef BAREOS_CATS_CATS_H_
 #define BAREOS_CATS_CATS_H_ 1
 
+#include "lib/volume_session_info.h"
```
franku:
```suggestion
class VolumeSessionInfo;
```

core/src/cats/sql_get.cc
```diff
@@ -1746,48 +1761,29 @@ bool BareosDb::GetNdmpEnvironmentString(JobControlRecord* jcr,
  * Returns false: on failure
  *         true: on success
  */
-bool BareosDb::GetNdmpEnvironmentString(JobControlRecord* jcr,
-                                        JobDbRecord* jr,
+bool BareosDb::GetNdmpEnvironmentString(const VolumeSessionInfo vsi,
```
franku:
```suggestion
bool BareosDb::GetNdmpEnvironmentString(const VolumeSessionInfo& vsi,
```

core/src/lib/volume_session_info.h
```diff
@@ -0,0 +1,35 @@
...
+
+#ifndef BAREOS_LIB_VOLUME_SESSION_INFO_H_
+#define BAREOS_LIB_VOLUME_SESSION_INFO_H_ 1
+
+struct VolumeSessionInfo {
+  uint32_t id;
+  uint32_t time;
+
+  /* explicit constructor disables default construction */
+  VolumeSessionInfo(uint32_t t_id, uint32_t t_time) : id(t_id), time(t_time) {}
```
franku:
```suggestion
  VolumeSessionInfo(const uint32_t t_id, const uint32_t t_time) : id(t_id), time(t_time) {}
```

arogge:
This is currently broken, because the DB api behaves quite different than expected.

PR #163, opened 2019-03-19
stored: don't touch devices that were not acquired

arogge:
when one of acquire_device_for_read() or
acquire_device_for_append() in do_mac_run()
returned false, the code still touched the
devices eventually failing another job that
had the device currently acquired.
This patch makes sure do_mac_run() does not
touch any device it couldn't aquire.


arogge:
Can someone please take a really good look at this? I cannot write a test as reservation usually handles this and it breaks only sometimes under high load.

PR #168, opened 2019-04-17
vss crossbuild: Eliminated an incompatible type conversion and a compiler error 

franku:
- in class VSSClient for GetWriterState replaced alists by a std::vector
- in lex_add(..) added nullpointer guard for filename to savely use strdup

### DIFF
--- a/core/src/win32/filed/vss.cc
+++ b/core/src/win32/filed/vss.cc
@@ -104,17 +104,6 @@ void VSSInit(JobControlRecord* jcr)
 }
 
 /*
- * Constructor
- */
-VSSClient::VSSClient()
-{
-  memset(this, 0, sizeof(VSSClient));
-  pAlistWriterState_ = New(alist(10, not_owned_by_alist));
-  pAlistWriterInfoText_ = New(alist(10, owned_by_alist));
-  uidCurrentSnapshotSet_ = GUID_NULL;
-}
-
-/*
  * Destructor
  */
 VSSClient::~VSSClient()
@@ -129,8 +118,6 @@ VSSClient::~VSSClient()
   }
 
   DestroyWriterInfo();
-  delete pAlistWriterState_;
-  delete pAlistWriterInfoText_;
 
   /*
    * Call CoUninitialize if the CoInitialize was performed successfully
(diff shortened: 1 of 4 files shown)


core/src/win32/filed/vss_generic.cc
```diff
@@ -1128,7 +1128,7 @@ bool VSSClientGeneric::CheckWriterStatus()
     return false;
   }
 
-  int nState;
+  int nState = 0;
```
arogge: this doesn't make sense here. Move after 1149 or even better wrap lines 1150 to 1180 into an IIFE.

core/src/win32/filed/vss.cc
```diff
@@ -223,44 +210,38 @@ bool VSSClient::GetShadowPathW(const wchar_t* szFilePath,
...
+    return writer_info_.at(nIndex).state_;
+  }
+  return 0;
 }
 
 void VSSClient::AppendWriterInfo(int nState, const char* pszInfo)
 {
-  pAlistWriterInfoText_->push(bstrdup(pszInfo));
-  pAlistWriterState_->push((void*)nState);
+  WriterInfo info;
```
arogge: if WriterInfo had a ctor(int, const char*) this function could just do
``writer_info_.push_back(WriterInfo{nState, pszInfo});``

core/src/win32/filed/vss.cc
```diff
@@ -223,44 +210,38 @@ bool VSSClient::GetShadowPathW(const wchar_t* szFilePath,
...
-  return pAlistWriterInfoText_->size();
+  return writer_info_.size();
 }
 
-const char* VSSClient::GetWriterInfo(int nIndex)
+const char* VSSClient::GetWriterInfo(size_t nIndex)
 {
-  return (char*)pAlistWriterInfoText_->get(nIndex);
+  if ( nIndex < writer_info_.size() ) {
+    return writer_info_.at(nIndex).info_text_.c_str();
```
arogge: you just did the bounds check the line before, so you can use [] instead of at().
```suggestion
    return writer_info_[nIndex].info_text_.c_str();
```

core/src/lib/lex.cc
```diff
@@ -221,10 +221,10 @@ static inline LEX* lex_add(LEX* lf,
                            LEX_ERROR_HANDLER* ScanError,
                            LEX_WARNING_HANDLER* scan_warning)
 {
-  LEX* nf;
+  if (!filename) { return nullptr; }
```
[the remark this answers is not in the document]
franku: This fixes a compiler warning that shows up a few lines down at strdup(filename) The warning disappeared after guarding the nullptr. 

As this fix just has a few lines I would not refactor all the code around.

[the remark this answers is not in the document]
franku: Good idea. Will try that.

core/src/win32/filed/vss_generic.cc
franku: Will remove this change.

core/src/lib/lex.cc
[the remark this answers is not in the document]
franku: Will remove this change.

franku:
merged into master with commit 37cf7809cb528cc695b5c6d6e52b8596568840c9

PR #169, opened 2019-04-18
dird: Fix pthread issues

bratkartoffel:
Fixes two double pthread_detach() calls and one pthread_getspecific()
without initialization, all leading to a SEGFAULT with musl libc and
undefined behaviour on other libc implementations.

This closes Bug #1073
Relates to https://bugs.alpinelinux.org/issues/10156

### DIFF
--- a/core/src/dird/dird.cc
+++ b/core/src/dird/dird.cc
@@ -650,6 +650,18 @@ bail_out:
   is_reloading = false;
   return reloaded;
 }
+
+void DetachIfNotDetached(pthread_t thr) {
+   /* only detach if not yet detached */
+   int _detachstate;
+   pthread_attr_t _gattr;
+   pthread_getattr_np(thr, &_gattr);
+   pthread_attr_getdetachstate(&_gattr, &_detachstate);
+   pthread_attr_destroy(&_gattr);
+   if(_detachstate != PTHREAD_CREATE_DETACHED) {
+      pthread_detach(thr);
+   }
+}
 } /* namespace directordaemon */
 
 /*
--- a/core/src/dird/dird.h
+++ b/core/src/dird/dird.h
@@ -246,6 +246,7 @@ struct runtime_job_status_t {
 #define FD_VERSION_54 54
 
 bool DoReloadConfig();
+void DetachIfNotDetached(pthread_t thr);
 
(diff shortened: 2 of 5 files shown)


core/src/dird/dird.cc
```diff
@@ -653,6 +653,18 @@ bool DoReloadConfig()
...
+
+void DetachIfNotDetached(pthread_t thr) {
+   /* only detach if not yet detached */
+   int _detachstate;
+   pthread_attr_t _gattr;
+   pthread_getattr_np(pthread_self(), &_gattr);
+   pthread_attr_getdetachstate(&_gattr, &_detachstate);
+   pthread_attr_destroy(&_gattr);
+   if(_detachstate != PTHREAD_CREATE_DETACHED) {
+      pthread_detach(pthread_self());
```
arogge:
```suggestion
      pthread_detach(thr);
```

bratkartoffel:
@arogge I've rebased the patch on current master, can you please take a look at it again?

core/src/dird/dird.cc
```diff
@@ -650,6 +650,18 @@ bool DoReloadConfig()
   is_reloading = false;
   return reloaded;
 }
+
+void DetachIfNotDetached(pthread_t thr) {
+   /* only detach if not yet detached */
+   int _detachstate;
+   pthread_attr_t _gattr;
+   pthread_getattr_np(thr, &_gattr);
```
arogge: Building on FreeBSD fails with your patch applied:
```
dird.cc:658:4: error: use of undeclared identifier 'pthread_getattr_np'
   pthread_getattr_np(thr, &_gattr);
   ^
1 error generated.
```

bratkartoffel:
As far as I can see (from googling), the function has a different name in freebsd:
pthread_attr_get_np

I'm not familiar with freebsd, I think that this issue can be resolved by some precompiler if-else-stuff.
Can you please change this?

arogge:
I have a candidate branch dev/arogge/master/fix-1073 with the #ifdef stuff for FreeBSD.
Could you make sure that branch works for your use-case?
I'm waiting for  #212 to be approved and could then create a new PR with your and my changes combined.

arogge:
Change has been implemented with PR #220

PR #174, opened 2019-05-14
FreeBSD: fix lmdb build on FreeBSD 12

pstorz:
Defining -D_WANT_SEMUN=1 enables the definition of "union semun"
in sys/sem.h on FreeBSD 12.

This fixes the following compile error:

  /src/lmdb/mdb.c:4832:14: error: variable has incomplete type 'union semun'
        union semun semu;
                    ^
  /core/src/lmdb/mdb.c:4832:8: note: forward declaration of 'union semun'
        union semun semu;
              ^
1 error generated.


franku:
merged into master with commit f3ccf82b62749cf426cfebccc080012d4de44de7

PR #178, opened 2019-06-06
Univention (UCS): change umask only temporary improvement

joergsteffens:
Do a try/finally exception handling around modifying the umask.


franku:
merged into master: 
75a3f9b29 -> fe2c74dad

PR #182, opened 2019-06-16
Fix responsive body padding

thebellenik:


fbergkemper:
PR #180 already fixes this issue which occurs on devices with a resolution < 1330px in a better way.

See: https://github.com/bareos/bareos/pull/180

Your patch would not fix the problem with the overflowing navbar element, we would like to not have the navbar breaking into a second row.

But thanks for your efforts!

fbergkemper:
Rejected

PR #188, opened 2019-06-24
webui: topnavbar follow up fix to commit 637095dee

fbergkemper:
Unfortunately, some languages have pretty long words, so we need to adjust the previous
introduced media query a bit.

The collapse happens at 1480px now to avoid the top navigation bar overflowing issue
in any language currently available for the webui.


arogge:
when I look at the history in 6 months I will not be able to figure out what this was supposed to fix.

PR #191, opened 2019-06-25
Change doc.bareos.org references to docs.bareos.org

arogge:
This changes the remaining references to doc.bareos.org in the repository to the new urls in docs.bareos.org.


arogge:
@fbergkemper can you check that the translation changes for the webui actually work? Thank you!

PR #207, opened 2019-07-03
Dev/pstorz/master/partitioned tables

pstorz:
This is a pull request for partioned tables


arogge:
as discussed in dev-meeting:
We need to make sure the changes are still compatible with MySQL, SQLite and PostgreSQL without partitioning.

PR #209, opened 2019-07-04
systemtests: cleanup

joergsteffens:
Remove unused files.
Partly these files are duplicate to the corresponding *.in files
and are overwritten.
bareos-dir resources:
  * removed fileset LinuxAll, as systemtests do not backup a full system
  * removed fileset Windows All Drives, as systemtests do not run on Windows
  * removed all schedules, as they are not required during systemtests
  * message resource: remove email configuration, as systemtests are not required to send emails

Other changes:
  * bin/bconsole now accepts parameter


systemtests/bin/bconsole
```diff
@@ -1,4 +1,5 @@
 #!/bin/bash
 # wrapper script to start/stop bconsole for each individual test
 . ./environment
-${bin}/bconsole -c ${conf}
+${bin}/bconsole -c ${conf} $*
```
arogge: Using $* will apply word-splitting to your parameters, this is usually not what you want.
```suggestion
${bin}/bconsole -c ${conf} "$@"
```

PR #211, opened 2019-07-08
docs: fixed some typos

huhnix:
Webui section of the docs

### DIFF
--- a/docs/manuals/source/IntroductionAndTutorial/InstallingBareosWebui.rst
+++ b/docs/manuals/source/IntroductionAndTutorial/InstallingBareosWebui.rst
@@ -29,7 +29,7 @@ Features
 
 -  Individual accounts and ACL support via Bareos restricted named consoles
 
--  Tape Autochanger management with the possibility to label, import/export media, and update your autochanger slot status
+-  Tape Autochanger management with the possibility to label, import/export media and update your autochanger slot status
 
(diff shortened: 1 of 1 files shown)


franku:
Would you mind changing the topic of the commit to "docs: fixed some typos"? Thanks.

joergsteffens:
I've been confused by the added commas (",") before the "and". My understanding is, that these commas are called serial commas or Oxford commas and are optional (see https://www.grammarly.com/blog/comma-before-and/).
As they are not (or only seldom) used in the remaining of the documentation, I propose not to use them at all.
Or we have to define, that they should be used from now on throughout the document.

huhnix:
> I've been confused by the added commas (",") before the "and". My understanding is, that these commas are called serial commas or Oxford commas and are optional (see https://www.grammarly.com/blog/comma-before-and/).

If the docs are in American English, then we should use them. If we prefer British English, then they're optional, agreed. :)

> Or we have to define, that they should be used from now on throughout the document.

Up to you – let me know what you decide. I'm happy either way. :)

franku:
Thank you for mentioning. We prefer no serial comma before 'and' however this is American or optional British English. Therefore my comments above apply.

franku:
merged into master with commit 7396a9c21ddf3bb232fb9e3b80bd3e525e1ff86c

PR #212, opened 2019-07-08
core: make systemtests work on FreeBSD

arogge:
The scripts used by the systemtests for the bareos startup/shutdown were flawed. This PR changes the pgrep behaviour so the running processes are actually found.


franku:
ok

PR #214, opened 2019-07-08
webui: Fixed always active debug messages in error.log

astoorangi:
Fixes #1059: Webui spams Apache error_log with bconsole messages


webui/module/Auth/src/Auth/Controller/AuthController.php
```diff
@@ -88,7 +88,7 @@ public function loginAction()
             $this->bsock = $this->getServiceLocator()->get('director');
             $this->bsock->set_config($config['directors'][$director]);
             $this->bsock->set_user_credentials($username, $password);
-            error_log($username);
+            //error_log($username);
```
arogge: don't comment out code that isn't required anymore. If you're concerned about the previous version: that's what git is for.
```suggestion
```

PR #224, opened 2019-07-12
docs: re-add information

joergsteffens:
Re-add some information to the documentation that did get lost during cleanup after the migration from LaTex to Sphinx


joergsteffens:
Discussed with @franku and accepted by him.

PR #228, opened 2019-07-14
Bugfix for #1072 (Newer versions of Fedora use stricter code checking)

tuxmaster5000:
https://bugs.bareos.org/view.php?id=1072

### DIFF
--- a/core/CMakeLists.txt
+++ b/core/CMakeLists.txt
@@ -42,6 +42,16 @@ if (${compiler_will_suggest_override})
    set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wsuggest-override")
 endif()
 
+# make format-security issues a compiler-error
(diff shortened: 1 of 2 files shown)


arogge:
Hi, thanks for the PR!
I'd really like to fix this problem. However, the line you propose the change for is probably better off by replacing it with strncpy().
The real issue is that our cmake config doesn't enable -Werror=format-security by default. We should fix that one!

tuxmaster5000:
Exhange the patched.

arogge:
Sorry, that one doesn't (yet) build on MacOS. The Xcode version we're using flags gettext strings as "not a string literal" which then triggers the format-security error.

arogge:
I guess we will need to upgade the Xcode on our Mac build system before we can merge this.
This might take some time. Sorry.

arogge:
I finally managed to get the MacOS builds working with a newer XCode version.
Not all tests have finished yet, but I think we will be able to merge this next week.

PR #231, opened 2019-07-18
Dev/franku/master/story 3081

franku:
Compare with 39e217eb9ebf313f080142ab48713ed24217dde9


franku:
It is a backport from an old 17.2 Bugfix and will improve in the next months.

PR #232, opened 2019-07-19
Fix bogus errors in TLS sessions

arogge:
Previously if a TLS session failed, the next TLS operation in the same thread saw the error of that failing session and was also terminated.
This PR fixes this behaviour by clearing the error-queue.

As the original issue that lead to this behaviour occured during a status client command to
a) a 17.2 or older client
b) a client with a mismatching password
we also add a systemtest that checks the correct behaviour when trying to status a client with a mismatching password (which eventually runs into a TLS handshake error)

### DIFF
--- a/core/src/lib/tls_openssl.cc
+++ b/core/src/lib/tls_openssl.cc
@@ -333,6 +333,14 @@ void TlsOpenSsl::TlsBsockShutdown(BareosSocket* bsock)
 
   int ssl_error = SSL_get_error(d_->openssl_, err_shutdown);
 
+  /*
+   * There may be more errors on the thread-local error-queue.
+   * As we just shutdown our context and looked at the errors that we were
+   * interested in we clear the queue so nobody else gets to read an error
+   * that may have occured here.
+   */
+  ERR_clear_error(); // empties the current thread's openssl error queue
+
   SSL_free(d_->openssl_);
   d_->openssl_ = nullptr;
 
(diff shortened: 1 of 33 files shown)


core/src/lib/tls_openssl.cc
```diff
@@ -333,6 +333,14 @@ void TlsOpenSsl::TlsBsockShutdown(BareosSocket* bsock)
 
   int ssl_error = SSL_get_error(d_->openssl_, err_shutdown);
 
+  /*
+   * There may be more errors on the thread-local error-queue.
+   * As we just shutdown our context and looked at the errors that we were
+   * interested in we clear the queue so nobody else gets to read an error
+   * that may have occured here.
+   */
+  ERR_clear_error();
```
franku:
```suggestion
  ERR_clear_error(); // empties the current thread's error queue
```

systemtests/tests/bconsole-status-client/testrunner
```diff
@@ -0,0 +1,42 @@
...
+
+@$out $tmp/log2.out
+messages
+quit
+END_OF_DATA
+
+run_bareos
+stop_bareos
+
+# if log2.out has not been written, bconsole crashed or disconnected
```
franku: Does this assumption always apply?

arogge: it does.
Either the bconsole session could execute
```
@$out $tmp/log2.out
messages
```
then the file will exist.
Or bconsole crashed before (at the status command), then the file will not exist.

franku:
You found that! Terrific!

Should merge into master and 18.2.

PR #234, opened 2019-07-19
Backport to 18.2: Fix bogus errors in TLS sessions #232

arogge:
If during or before SSL_shutdown() errors occured that were saved in the
thread-local openssl error queue, these were not read and thus retained
on the queue until another tls session read them.
This lead to bogus error detection. Now we simply clear all errors from
the queue after we're done with our openssl context, so the errors do
not "leak" into another context anymore.

(cherry picked from commit cf3e1064f495b0592e0202edb2cec7e358c5b378)


franku:
Successful on my local buildsystem on a Fedora 29.

PR #243, opened 2019-08-16
webui: Notify user about Bvfs cache update

fbergkemper:
Display an bvfs update cache notification in the restore file tree
container while loading the tree and in case of an error after tree
loading failed.

The tree node loading timeout is now configurable by the
node_loading_timeout parameter in restore section of the
configuration.ini file. The default of the parameter is
set to 120 seconds.

[restore]
; Restore file tree node loading timeout after n milliseconds
; Default: 120000 milliseconds
node_loading_timeout=120000


franku:
Pls. see comments on PR#240

PR #246, opened 2019-08-21
core: replace crc32 implementation

arogge:
The previous crc32 implementation was not optimal from a performance
point-of-view.
This patch introduces Stephan Brumme's Fast CRC32 which is a lot faster
than the old algorithm.


arogge:
I'm not sure the files are in the right location. Can somebody please review?

joergsteffens:
Additional licenses must also be described in https://github.com/bareos/bareos/blob/master/core/LICENSE and https://github.com/bareos/bareos/blob/master/core/debian/copyright.footer

arogge:
Licensing info will be added once we've finally agreed on the file locations, because the paths to the sources will be coded into the licensing info.

core/src/tests/CMakeLists.txt
```diff
@@ -290,3 +290,13 @@ IF(HAVE_EXECINFO_H AND HAVE_BACKTRACE AND HAVE_BACKTRACE_SYMBOLS)
   gtest_discover_tests(test_backtrace TEST_PREFIX gtest:)
 ENDIF()
 
+####### test_backtrace #####################################
```
franku:
```suggestion
####### test_crc32 #####################################
```

stbrumme:
Hi, I'm the author of the CRC32 library.
My lib uses the zlib license (see https://github.com/stbrumme/crc32/blob/master/LICENSE ).
Quoting from the license: "an acknowledgment in the product documentation would be appreciated but is not required".

Have a nice day,
Stephan

PR #253, opened 2019-08-27
systemtests: Fix build without Python support

fbergkemper:
Fix CMake Error: "get_target_property() called with non-existent target"
if Python support is FALSE.


systemtests/CMakeLists.txt
```diff
@@ -99,10 +99,13 @@ set(mon_sd_password "mon_sd_password")
 set(python_plugin_module_src_dir ${CMAKE_SOURCE_DIR}/core/src/plugins)
 set(plugindirtmp ${PROJECT_BINARY_DIR}/plugindirtmp)
 set(rscripts ${PROJECT_BINARY_DIR}/scripts)
-get_target_property(fd_plugin_binary_path python-fd BINARY_DIR)
-if(NOT client-only)
-  get_target_property(dir_plugin_binary_path python-dir BINARY_DIR)
-  get_target_property(sd_plugin_binary_path python-sd BINARY_DIR)
+
+if(PYTHONLIBS_FOUND)
```
joergsteffens: Other CMakeLists.txt files use 
IF(HAVE_PYTHON)
instead.

fbergkemper: Good point, I will adapt it to the most used.

fbergkemper: Reverted, as HAVE_PYTHON does not mean libraries and header files for Python development are available.

PR #260, opened 2019-08-30
systemtests: run python-fd-plugin-local-fileset-test only if python-fd is build

joergsteffens:
Previous the PYTHONLIBS_FOUND has been checked,
which wasn't always defined.


franku:
Going to merge after all tests passed.

PR #261, opened 2019-08-30
workq: refactor code or remove unneded code for worker threads

franku:
The workq functionality was overengineered and did not do what it
  probably should do.

  The names workq or workqueue did not explain what the code
  actually did and it was not reliable.

  Therefore refactored the complete code:

- ThreadList is now a class that contains the calls to create
  worker-threads who will call a pre-defined handler routine
- ThreadListPrivate contains the implementation and is hidden from
  the public interface
- using ThreadList with RAII in the daemon code, accordingly
- the new functionality consists mainly of a list where new threads
  are added to, therefore the name of all files and variables
  are changed into names containing "thread list" or similar
- used std::thread instead of pthread
- used std::set instead of a linked list
- worker threads start detached and cleanup resources by themself

- removed or renamed variables
- removed unused code in the thread start and at the end
- previous return type of int was not used as intended,
  therefore a simple bool is enough
- upon shutdown threads are not joined but the parent thread will
  wait until the thread list is empty or a timeout occurs
- added a user callback for each worker-thread that is called from
  the parent thread before it enters the waiting state

- ua_server: close user_agent_socket only in FreeUaContext
- synchronize shutdown of worker threads on daemon shutdown
- updated code that calls the new functions accordingly

- fd and dir: added a worker-thread shutdown handler to each
  socket server
- sd: do not cleanup jcr memory of a console connection in the
  terminate handler
- dird: StopSocketServer is now called before StopWatchdog so
  socket operations still can time out

- added a unit test


franku:
Should improve again. Not yet ready to review.

PR #263, opened 2019-09-02
dir: add new resource User to config

franku:
Up to now a User resource in the director config was an alias for the Console resource and it was only used for pam authentication. 

Actually the User resource does not need a password and most of all other directives from Console Resource. Therefore this PR introduces a dedicated User resource. 

Implemented: 

- separated struct user_acl contained in UserResource and ConsoleResource
- only referenced when using pam authentication
- corrected pam documentation

**Edit after review:** 
- enhanced bconsole-pam test by configure add user name=user3
- rewrite macro USER_ACL to use in all relevant initializer lists (RunACL was missing in the profile list)
- removed a potential wrong positive from bconsole-pam test
- updated docs
- rebuilt and checked in resource json files
- package installer adds user subdirectory for bareos-dir.d


core/src/dird/dird_conf.cc
```diff
@@ -186,31 +187,42 @@ static ResourceItem profile_items[] = {
...
+  { "PoolACL", CFG_TYPE_ACL, ITEM(resource, user_acl.ACL_lists), Pool_ACL, 0, NULL, NULL, NULL },\
+  { "CommandACL", CFG_TYPE_ACL, ITEM(resource, user_acl.ACL_lists), Command_ACL, 0, NULL, NULL, NULL },\
+  { "FileSetACL", CFG_TYPE_ACL, ITEM(resource, user_acl.ACL_lists), FileSet_ACL, 0, NULL, NULL, NULL },\
+  { "CatalogACL", CFG_TYPE_ACL, ITEM(resource, user_acl.ACL_lists), Catalog_ACL, 0, NULL, NULL, NULL },\
+  { "WhereACL", CFG_TYPE_ACL, ITEM(resource, user_acl.ACL_lists), Where_ACL, 0, NULL, NULL, NULL },\
+  { "PluginOptionsACL", CFG_TYPE_ACL, ITEM(resource, user_acl.ACL_lists), PluginOptions_ACL, 0, NULL, NULL, NULL },\
+  { "Profile", CFG_TYPE_ALIST_RES, ITEM(resource, user_acl.profiles), R_PROFILE, 0, NULL, "14.2.3-",\
+     "Profiles can be assigned to a Console. ACL are checked until either a deny ACL is found or an allow ACL. "\
+     "First the console ACL is checked then any profile the console is linked to." }
+
```
joergsteffens: How about using the USER_ACL (except "Profile") also for profile_items?

joergsteffens: Could you please verify, that "configure add user name=test ..." works?

franku: > Could you please verify, that "configure add user name=test ..." works?
Looks like. But it does not create a folder "User" if it does not already exist. Is this the itended behaviour?

franku: > How about using the USER_ACL (except "Profile") also for profile_items?

I would deny. They do not share the same items (RunACL, profiles) and in the resource objects they refer to different C-structures. 

The USER_ACL Macro was introduced to bring together console and user resource because they must be able to use the same interface to access the data.

joergsteffens: > > Could you please verify, that "configure add user name=test ..." works?
> 
> Looks like. But it does not create a folder "User" if it does not already exist. Is this the itended behaviour?

Yes, the user folder should be generated by the package.

franku: - enhanced using bconsole-pam test by configure add user name=user3
- rewrite macro USER_ACL to use in all relevant initializer lists (RunACL was missing in the one mentioned above)

franku:
We need a description for "RunACL", any ideas regarding the wording?

systemtests/tests/bconsole-pam/etc/bareos/bareos-dir.d/profile/operator.conf
```diff
@@ -3,7 +3,7 @@ Profile {
    Description = "Profile allowing normal Bareos operations."
 
    Command ACL = !.bvfs_clear_cache, !.exit, !.sql
-   Command ACL = !configure, !create, !delete, !purge, !prune, !sqlquery, !umount, !unmount
+   Command ACL = configure, !create, !delete, !purge, !prune, !sqlquery, !umount, !unmount
```
joergsteffens: You don't want to allow "configure". If you allow it, a user could create a new console with no ACL restrictions.
And if you really want to allow it, there is no need to explicitly specify it, as the last entry is "*all*", which allows all commands not explicitly denied.

franku: The "configure" command is needed for this test; "user3" will be added using "configure add" as requested from you.
There are no side effects to production systems because this is only used for this test, thus no concerns.

PR #274, opened 2019-09-23
python-bareos and dird commands improvments

joergsteffens:
Added systemtests for python-bareos module.

Improve exception handling in python-bareos

Also fixes and improves some console commands in the Bareos Director.

### DIFF
--- a/systemtests/tests/python-bareos-test/testrunner
+++ b/systemtests/tests/python-bareos-test/testrunner
@@ -28,30 +28,23 @@ cat <<END_OF_DATA >$tmp/bconcmds
 @$out /dev/null
 messages
 @$out $tmp/log1.out
-setdebug level=100 storage=File
-label volume=TestVolume001 storage=File pool=Full
-run job=$JobName yes
-status director
-status client
-status storage=File
-wait
-messages
-@#
-@# now do a restore
-@#
-@$out $tmp/log2.out
-wait
-restore client=bareos-fd fileset=SelfTest where=$tmp/bareos-restores select all done
-yes
-wait
+configure add console=admin password=secret profile=all
 messages
 quit
 END_OF_DATA
 
 run_bareos
+
+export PYTHONPATH=${PROJECT_SOURCE_DIR}/../python-bareos/
+print_debug "PYTHONPATH=$PYTHONPATH"
+if PYTHONRESULT=$(./python-bareos-unittest.py -v); then
+    print_debug "$PYTHONRESULT"
+else
+    set_error "$PYTHONRESULT"
+fi
+
 check_for_zombie_jobs storage=File
+
 stop_bareos
 
-check_two_logs
-check_restore_diff ${BackupDirectory}
 end_test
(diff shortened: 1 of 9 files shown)


core/src/cats/sqlite_queries.inc
```diff
@@ -949,14 +949,46 @@ const char *BareosDbSqlite::query_definitions[] = {
...
-          "VolCapacityBytes,VolStatus,Enabled,Recycle,VolRetention, "
-          "VolUseDuration,MaxVolJobs,MaxVolFiles,MaxVolBytes,InChanger, "
-          "EndFile,EndBlock,LabelType,StorageId,DeviceId, "
-          "LocationId,RecycleCount,InitialWrite,ScratchPoolId,RecyclePoolId, "
-          "Comment,Name AS Storage "
+"SELECT    Media.MediaId, "
+          "Media.VolumeName, "
+          "Media.Slot, "
+          "Media.PoolId, "
+          "(SELECT Name FROM Pool WHERE Pool.PoolId=Media.PoolId) AS Pool, "
```
arogge: Can you really just change the query without changing the code that uses it?

systemtests/tests/python-bareos-test/testrunner
```diff
@@ -28,30 +28,24 @@ cat <<END_OF_DATA >$tmp/bconcmds
...
+    print_debug "$PYTHONRESULT"
+else
+    set_error "$PYTHONRESULT"
+fi
+
 stop_bareos
 
-check_two_logs
-check_restore_diff ${BackupDirectory}
+#check_two_logs
```
arogge: delete lines you don't need

systemtests/tests/python-bareos-test/testrunner
```diff
@@ -28,30 +28,24 @@ cat <<END_OF_DATA >$tmp/bconcmds
...
+#check_restore_diff ${BackupDirectory}
```
arogge: delete lines you don't need

systemtests/tests/python-bareos-test/testrunner
```diff
@@ -28,30 +28,24 @@ cat <<END_OF_DATA >$tmp/bconcmds
...
-yes
-wait
+configure add console=admin password=secret profile=operator
 messages
 quit
 END_OF_DATA
 
 run_bareos
-check_for_zombie_jobs storage=File
+#check_for_zombie_jobs storage=File
```
arogge: delete lines you don't need

python-bareos/bareos/bsock/lowlevel.py
```diff
@@ -5,12 +5,13 @@
 # Authentication code is taken from
 # https://github.com/hanxiangduo/bacula-console-python
 
-from   bareos.exceptions import *
+import bareos
```
arogge: maybe `import bareos.exceptions` instead of `import bareos`. Otherwise the `from` imports below should probably be converted too

core/src/cats/sqlite_queries.inc
joergsteffens: It adds extra output to the "llist media all" command. However, this should not hurt,

PR #276, opened 2019-09-24
Fix nullptr cornercase in mtx-changer parser

arogge:
Previously when mtx-changer provided text with no colons in it, field2 was a nullptr. This nullptr wasn't checked correctly and lead to a crash.
The code did not match the comment, because it didn't check that f1 and f2 were set, but that f1 *or* f2 were set.
This patch now makes the code behave as described in the comment above.


arogge:
As discussed this will be backported to 18.2 and 17.2

PR #279, opened 2019-09-25
internalize fastlz

arogge:
Previusly libfastlz has been an external dependency to bareos.
However, as this were only a few source-files and nobody seems to care about the upstream project anymore I propose we make this internal to bareos.


arogge:
Currently packaging changes for debian haven't been done yet.

core/CMakeLists.txt
```diff
@@ -692,7 +691,6 @@ MESSAGE("   OpenSSL support:              ${OPENSSL_FOUND} ${OPENSSL_VERSION} ${
 MESSAGE("   PAM support:                  ${PAM_FOUND} ${PAM_LIBRARIES} ${PAM_INCLUDE_DIRS} ")
 MESSAGE("   ZLIB support:                 ${ZLIB_FOUND} ${ZLIB_LIBRARIES} ${ZLIB_INCLUDE_DIRS} ")
 MESSAGE("   LZO2 support:                 ${LZO2_FOUND} ${LZO2_LIBRARIES} ${LZO2_INCLUDE_DIRS} ")
-MESSAGE("   FASTLZ support:               ${FASTLZ_FOUND} ${FASTLZ_LIBRARIES} ${FASTLZ_INCLUDE_DIRS} ")
```
pstorz: Maybe we could put a message like "FASTLZ: yes, using internal library" or something like that

arogge: I wouldn't really mention that. It should be mentioned in the release notes for 19 and that's it. If we really need a message, I'd propose "FASLZ support: integrated".

arogge:
FreeBSD packaging still needs work.

arogge:
FreeBSD build issues have been resolved. Documentation has been added.

PR #284, opened 2019-09-30
Make embedded paths reproducible

arogge:
Using cmake __FILE__ will point to the absolute path of the source.
This leads to two issues:
- object files change depending on where you checkout the sources
- local paths of the build-system leak into the resulting binaries

This PR adds and applies script that can add a preprocessor directive to every source file, so __FILE__ evaluates to a relative path.

### DIFF
--- a/core/CMakeLists.txt
+++ b/core/CMakeLists.txt
@@ -48,6 +48,7 @@ if (${CMAKE_SYSTEM_NAME} MATCHES "SunOS")
   set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
 endif()
 
+include(CheckCCompilerFlag)
 include(CheckCXXCompilerFlag)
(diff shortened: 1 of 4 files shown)


franku:
We should check if debugging support still works as expected.

arogge:
Checking that debugging still works for:
- [ ] Source Build on CentOS
- [ ] Packages on Debian
- [ ] Packages on CentOS

arogge:
I checked this on CentOS, Debian and in my custom build from source. All still works as expected.

arogge:
The goal of this PR was to improve reproducible builds so we can leverage ccache better. I have found another (probably better) way to do this without touching the source files and will update the PR accordingly.

arogge:
The path is now set using `-ffile-prefix-map` if the compiler supports it.
Currently this is only the case for GCC 8.
So this should work ouf-of-the-box on Windows and all recent Distros. For RHEL/CentOS 6 and 7 a recent enough gcc is available through SCL (devtoolset-8), so we could at least fix this for most of our platforms.

PR #295, opened 2019-10-02
packaging: make specfiles compatible to docker

arogge:
This PR changes the specfiles so we can build bareos in simple docker containers.
Mostly this moves changes that were done in OBS project configuration into the specfiles.


arogge:
We should probably backport these to 18.2, 17.2 and maybe even 16.2 (if we want to be able to build that in docker)

PR #301, opened 2019-10-08
Fixes #1118: Incorrect and ugly default mail values

wolfaba:
Fixes #1118: Incorrect default value for smtp_host and ugly default values for job_email and dump_email in BareosSetVariableDefaults.cmake

This patch fixes default mail values for smtp_host, job_email and dump_email.


wolfaba:
I hope now is better :-)

arogge:
@pstorz if you don't have any objections I'd like to merge this.

pstorz:
@arogge :Obvious rhese values make sense so I agree top merge.

PR #303, opened 2019-10-14
Systemtest python plugin dir

astoorangi:
The test checks if a python module logs events from the director correctly.

Added "python-modules" functionality

Every system test with a "python-module" folder will get gets the default python modules from bareos


systemtests/tests/python-dir-plugin-test/testrunner
```diff
@@ -43,23 +43,15 @@ status client
...
 run_bareos
 check_for_zombie_jobs storage=File
 stop_bareos
 
-check_two_logs
-list="$(find ${BackupDirectory} -type f)"
-check_restore_only_files_diff $list
+if ! grep -Eq "^[bDirEventJobStart,bDirEventJobInit,bDirEventJobRun,bDirEventJobEnd]${1}" ${tmp}/test-plugin.log; then
+  estat=1
+fi
```
joergsteffens: Your grep statement looks wrong. How about the following one:

```suggestion
for i in bDirEventJobStart bDirEventJobInit bDirEventJobRun bDirEventJobEnd; do
  if ! grep "$i" ${tmp}/test-plugin.log; then
    set_error "Failed to find logged event $i"
  fi
done
```

joergsteffens: Beside the simple change to the "grep" line, I agree to this PR, but did ask @franku to also review this.
One other thing: please reformulate your 2. commit message summary to something like:

systemtests: added python-dir-plugin-test

systemtests/tests/python-dir-plugin-test/python-modules/bareos-dir-test.py
```diff
@@ -0,0 +1,74 @@
...
+
+  elif event == bDirEventType['bDirEventJobRun']:
+    toFile("bDirEventJobRun\n")
+
+  return bRCs['bRC_OK']
+
+def toFile(text):
+  doc = open(outputfile,"a")
+  doc.write(text)
+  doc.close()
```
arogge: while most things in Bareos are indented with 2 spaces, python has an official style guide (PEP-8) that has been applied to the rest of our Python code.
Consider reformatting using `black` or at least make sure your indentation uses 4 spaces.

systemtests/scripts/functions
```diff
@@ -288,7 +288,6 @@ set_error()
 {
     estat=9
     echo "ERROR: $*" >> "$tmp/err.log"
-    echo "ERROR: $*"
```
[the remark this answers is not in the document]
astoorangi: end_test(), which is called at the end of each test, prints the "$tmp/err.log" file, so the error would be shown twice.

systemtests/tests/python-dir-plugin-test/python-modules/bareos-dir-test.py
astoorangi: I adopted the indentation from the templates in the core. Should I reformat them with black too?

https://github.com/bareos/bareos/blob/master/core/src/plugins/dird/bareos-dir.py.template

systemtests/CMakeLists.txt
```diff
@@ -97,7 +97,6 @@ set(mon_dir_password "mon_dir_password")
 set(mon_fd_password "mon_fd_password")
 set(mon_sd_password "mon_sd_password")
 set(python_plugin_module_src_dir ${CMAKE_SOURCE_DIR}/core/src/plugins)
-set(plugindirtmp ${PROJECT_BINARY_DIR}/plugindirtmp)
```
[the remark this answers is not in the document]
astoorangi: No, it is not required. I re-added the variable

systemtests/tests/python-dir-plugin-test/python-modules/bareos-dir-test.py
joergsteffens: You could do so, but please use a separate commit or even a separate PR.

systemtests/scripts/functions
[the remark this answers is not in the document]
arogge: While we shouldn't print the message twice, we also should not delay its output till the test is finished.
For now I'd like to keep the old behaviour. Maybe we can discuss improvements to the systemtest logging in general later.

systemtests/tests/python-dir-plugin-test/python-modules/bareos-dir-test.py
arogge: A separate commit would be good for that, yes.

astoorangi: Should I reformat all python scripts in the repository or just the templates for the plugins?

astoorangi: PR: #322

joergsteffens:
I agree to merging this PR.
@arogge will you do the merging?

PR #313, opened 2019-10-30
Segfaults in CrlReloaderGetBySubject

adrb:
In case none of the CRLs match the name and less than MAX_CRLS  were loaded, we will got segfault in X509_NAME_cmp because  data->crls[cnt] is null

Signed-off-by: Adrian Brzezinski <adrian.brzezinski@eo.pl>

### DIFF
--- a/core/src/lib/tls_openssl_crl.cc
+++ b/core/src/lib/tls_openssl_crl.cc
@@ -271,6 +271,9 @@ static int CrlReloaderGetBySubject(X509_LOOKUP* lookup,
   ret->type = 0;
   ret->data.crl = NULL;
   for (cnt = 0; cnt < MAX_CRLS; cnt++) {
+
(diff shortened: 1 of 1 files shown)


arogge:
@franku can you take a quick look at this?

franku:
However everything else looks fine, please correct the commit message to: 

**tls: avoid a nullpointer dereference**

- If none of the CRLs match the name and less than MAX_CRLS were loaded, it comes to a segmentation fault because data->crls[cnt] is null

franku:
restarted jenkins CD process

adrb:
From what I checked, this error goes back till version 14.2 where it was in file _src/lib/tls_openssl.c_ and _crl_reloader_get_by_subject_ function, so you may consider backporting it.

franku:
Added Adrian Brzezinski to AUTHORS, thank you for your contribution!

PR #314, opened 2019-10-31
Enhancements to python-bareos

joergsteffens:
Enhancements to python-bareos:
  * Improve error handling.
  * Implement Console Protocol of Bareos >= 18.2.4. Before connection did use the Bareos Director  compat mode.
  * Implement PAM authentication.
  * build python-bareos for Python 2 and Python3 (DEB and RPM)
  * Test with Python 2.6 (RHEL/Centos 6).
  * Add a lot of systemtests of the different console connections.

Tested:
  * Local systemtests succeed.


joergsteffens:
Verified that for Debian based distribution python-bareos and python3-bareos are build.

joergsteffens:
As discussed with @franku, integrate the fix commits into the original ones.

joergsteffens:
Added commit for Python 2 and Python 3 RPMs. Rebased on current master.

python-bareos/bareos/bsock/lowlevel.py
```diff
@@ -239,14 +307,49 @@ def send(self, msg=None):
...
+                raise bareos.exceptions.ConnectionLostError(errormsg)
+            length -= len(submsg)
+            msg += submsg
+        return msg
+
+
     def recv(self):
-        '''will receive data from director '''
+        '''
+        Receive a single Director message.
```
franku:
```suggestion
        Receive a single message.
```

python-bareos/debian/control
```diff
@@ -1,13 +1,20 @@
 Source: python-bareos
-Maintainer: Joerg Steffens <joerg.steffens@bareos.com>
+Maintainer: Bareos Team <packager@bareos.com>
```
franku:
```suggestion
Maintainer: Bareos GmbH & Co. KG <packager@bareos.com>
```

systemtests/scripts/functions
```diff
@@ -473,6 +473,28 @@ stop_bareos()
    fi
 }
 
+
+run_python_unittests()
```
franku:
```suggestion
run_python_unittests()
```
Perhaps Philipp should review this function because he is currently reviewing the unit-test code, especially the path arguments.

systemtests/tests/bconsole-pam/python-bareos-unittest.py
```diff
@@ -0,0 +1,196 @@
...
+import unittest
+
+import bareos.bsock
+import bareos.exceptions
+from   bareos.bsock.protocolversions import ProtocolVersions
+
+
+class PythonBareosBase(unittest.TestCase):
+    director_address = 'localhost'
+    director_port = 9101
```
franku: The port number should be dynamic. This is mandatory when running tests parallel. 
-> pstorz, arogge

systemtests/tests/bconsole-pam/python-bareos-unittest.py
```diff
@@ -0,0 +1,196 @@
...
+            director = bareos.bsock.DirectorConsole(
+                address=self.director_address,
+                port=self.director_port,
+                name=self.console_pam_username,
+                password=bareos_password,
+                pam_username=pam_username,
+                pam_password=pam_password)
+
+
+    def test_login_with_not_wrong_password(self):
```
franku:
```suggestion
    def test_login_with_right_password(self):
```

systemtests/tests/bconsole-pam/python-bareos-unittest.py
franku: This is done below by reading the BAREOS_DIRECTOR_PORT environment variable.

python-bareos/debian/control
joergsteffens: Bareos Team is used throughout similar settings, in SPEC files, DEB, ...

systemtests/tests/bconsole-pam/python-bareos-unittest.py
joergsteffens: In fact, the method name has been wrong. I changed it to "test_login_with_wrong_password".

python-bareos/bareos/bsock/lowlevel.py
```diff
@@ -123,79 +177,92 @@ def __connect_tls_psk(self):
...
+            raise bareos.exceptions.AuthenticationError('Received unexcepted signal: {0}'.format(str(e)))
         if not result:
             raise bareos.exceptions.AuthenticationError("failed (in response)")
         if not self._cram_md5_challenge(clientname=self.name, password=self.password.md5(), tls_local_need=0, compatible=True):
             raise bareos.exceptions.AuthenticationError("failed (in challenge)")
-        self.recv_msg(self.auth_success_regex)
-        self.auth_credentials_valid = True
-        return True
+
+        self.get_and_evaluate_auth_responses()
```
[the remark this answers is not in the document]
joergsteffens: It checks for a series of Protocol Message (IDs). If requested, sends PAM credentials.
Suggestions for a better method name? How is this part of authentication named? I found no hint in https://github.com/bareos/bareos/blob/master/core/src/console/console.cc#L1078

python-bareos/bareos/bsock/lowlevel.py
```diff
@@ -39,42 +42,77 @@ class LowLevel(object):
...
     def __connect(self):
         connected = False
-        if self.tls_psk_require and not self.is_tls_psk_available():
-            raise bareos.exceptions.ConnectionError(u'TLS-PSK is required, but sslpsk module not loaded/available.')
+        connected_plain = False
+        auth = False
+        if self.tls_psk_require:
+            if not self.is_tls_psk_available():
+                raise bareos.exceptions.ConnectionError(u'TLS-PSK is required, but sslpsk module not loaded/available.')
+            if not self.tls_psk_enable:
```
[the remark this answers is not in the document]
joergsteffens: Thanks for pointing this out. I considered this. However, in a default installation of python-bareos  TLS-PSK is not available as the Python module sslpsk required for this functionality is not available as package in the current Linux distributions and the Bareos project decided not to ship the module in this own repositories.. While it can be installed using pip, in my environment the module has only been usable with Python 2, not with Python 3 (library problem with openssl).
So considering this, I assume that many persons will use python-bareos without TLS-PSK, maybe even in compat (protocol <= 18.2.4) mode.
Therefore in kept the possibility to disable TLS-PSK on class initiation. By this, the already implemented systemtests can control if they want to test with TLS-PSK enabled or disabled, even if the sslpsk module is installed.
However, as I agree that most people do not want to manually disable TLS-PSK, I removed the command line option --no-tls-psk in an extra commit.

python-bareos/bareos/bsock/lowlevel.py
```diff
@@ -86,28 +124,44 @@ def __connect(self):
...
 
-        return connected
+        if connected:
+            try:
+                auth = self.auth()
+            except bareos.exceptions.PamAuthenticationError:
+                raise
+            except bareos.exceptions.AuthenticationError:
+                if self.connection_type == ConnectionType.DIRECTOR and self.protocolversion > ProtocolVersions.bareos_12_4:
+                    # reconnect and try old protocol
```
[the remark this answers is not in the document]
joergsteffens: You understand the code correct.
My comments from above do also apply here.
However, I changed the behavior a bit:
- when it fallbacks to an old protocol, it issues a warning (has been a debug message before).
- when the user requests a specific protocol (last), there will be no fallback.

python-bareos/bareos/bsock/lowlevel.py
```diff
@@ -208,10 +275,10 @@ def call(self, command):
         '''
         if isinstance(command, list):
             command = " ".join(command)
-        return self.__call(command, 0)
+        return self.__call(command)
 
 
-    def __call(self, command, count):
+    def __call(self, command):
```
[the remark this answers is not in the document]
joergsteffens: ok.

joergsteffens:
Added a commit introducing the test class PythonBareosJsonRunScriptTest.
This class checks if RunScripts are executed in the correct context.
Some of these tests will fail, until #358 gets merged.

python-bareos/bareos/bsock/lowlevel.py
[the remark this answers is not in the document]
franku: As you are presenting me your code, what name would you like to suggest?

systemtests/scripts/functions
pstorz: If the sourcecode is checked out in a long subpath, the tests partly failed because of the limited length of some resources in bareos. To avoid these problems, absolute paths have been changed to relative paths where this problem showed up. 

As the pathes used here seem only to be stored in the environment I don't think that we will have these problems.

python-bareos/bareos/bsock/lowlevel.py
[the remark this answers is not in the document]
joergsteffens: renamed to finalize_authentication().

joergsteffens:
Applied discussed changed and rebased on master.

PR #315, opened 2019-10-31
lib: IsNameValid(): disallow leading and trailing spaces

pstorz:
Names with leading and trailing space are now illegal.
Also added a unit test that checks that IsNameValid() works
as expected.


arogge:
If we added a second parameter `const char* accept_additional` that defaults to `""` we could check this with `strchr()` in addition to `accept` so the caller could pass additional valid characters.

Then it should be possible to implement IsAclEntryValid() like this
```
bool IsAclEntryValid(const char* acl, char* msg) {
  if (acl[0] == '!') ++acl; // negated resource name/pattern, skip negation sign
  return IsNameValid(acl, msg, '.*')
}
```

PR #319, opened 2019-11-05
python-bareos: update RPM spec file to support Python2 and Python3

joergsteffens:
This is a draft, required to test the build process for all supported distribution. Currently only tested on SLES on OBS.


joergsteffens:
Revoke this PR, until issues with RHEL are solved.

PR #322, opened 2019-11-06
python: reformatted all python scripts with black

astoorangi:


astoorangi:
Should I reformat all python scripts in the repository or just the templates for the plugins?

joergsteffens:
Replacing all single quotes (') by double quotes (") is neither required nor helpful.
I've used yapf to reformat some Python scripts in the past, and yagf formats the code differently to black. We should use only one tool for the reformatting.

I also do not like the changes of python-bareos, as some of the changes make the code less readable and there are some outstanding PRs for that code, so this PR would prevent that the other PRs can be merged.

pstorz:
I think having a common code formatter makes sense.
We have very good experiences doing this on the c and c++ source code.

sduehr:
I was skeptical first, I've always used flake8 and fixed the code manually.
Quick comparision of the yapf and black result for a simple example:

```python
def myfunc(templatefile, path, enable, password, client_name, bareos_gid):
    with open(templatefile, 'r') as f:
        content = f.read()

    t = string.Template(content)
    with os.fdopen(os.open(path, os.O_CREAT | os.O_WRONLY, stat.S_IRUSR | stat.S_IWUSR | stat.S_IRGRP), 'w') as f:
        f.write(t.substitute(enable=enable, password=password, client_name=client_name))
    os.chown(path, -1, bareos_gid)
```
yapf result:

```python
def myfunc(templatefile, path, enable, password, client_name, bareos_gid):
    with open(templatefile, 'r') as f:
        content = f.read()

    t = string.Template(content)
    with os.fdopen(
            os.open(path, os.O_CREAT | os.O_WRONLY,
                    stat.S_IRUSR | stat.S_IWUSR | stat.S_IRGRP), 'w') as f:
        f.write(
            t.substitute(enable=enable,
                         password=password,
                         client_name=client_name))
    os.chown(path, -1, bareos_gid)
```

black result:

```python
def myfunc(templatefile, path, enable, password, client_name, bareos_gid):
    with open(templatefile, "r") as f:
        content = f.read()

    t = string.Template(content)
    with os.fdopen(
        os.open(
            path, os.O_CREAT | os.O_WRONLY, stat.S_IRUSR | stat.S_IWUSR | stat.S_IRGRP
        ),
        "w",
    ) as f:
        f.write(t.substitute(enable=enable, password=password, client_name=client_name))
    os.chown(path, -1, bareos_gid)
```
I think the black result looks better, in this example it's more obvious that the "w" is a parameter of os.fdopen() and not os.open().

As of https://github.com/psf/black#strings there are some good reasons to standardize on double quotes, however, we could use `black --skip-string-normalization` first.

I also like that as of https://github.com/psf/black#line-length black defaults to 88 characters per line, personally I'd prefer even longer lines, but there are good reasons to agree on something below 100.

astoorangi:
I also think we should have a formatting standard, I would prefer black for its consistency. Therefore I think the standardization of the quotes makes sense, the reasoning of black for double quotes sounds plausible to me.

joergsteffens:
Hi @astoorangi, after some internal discussion, we decided to use the Python formatter black from now on for all Python code in Bareos.
However, I would prefer if you do this pull request again, containing only the code in core, vmware and webui.
I'll take care about the remaining code as soon as #314 is merged.

astoorangi:
Also in docs and regress are some Python scripts (docs/manuals/scripts|source/* and regress/starttime|endtime). Should I reformat them or do you want to do that, @joergsteffens ?

PR #330, opened 2019-11-09
Fix crash on start on Fedora 30

tuxmaster5000:
BAREOS interrupted by signal 11: Segmentation violation
Which will happens when not all interfaces are ready.

### DIFF
--- a/core/platforms/systemd/bareos-dir.service.in
+++ b/core/platforms/systemd/bareos-dir.service.in
@@ -14,8 +14,8 @@
 [Unit]
 Description=Bareos Director Daemon service
 Documentation=man:bareos-dir(8)
-Requires=nss-lookup.target network.target remote-fs.target time-sync.target
-After=nss-lookup.target network.target remote-fs.target time-sync.target postgresql.service mysql.service mariadb.service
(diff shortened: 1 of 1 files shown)


pstorz:
Hello, 
thank you very much for your PR. However, I would like to fix the problem,
while your PR looks more like a workaround.

Could you please provide a descriptino how to reproduce the problem?
Then we probably can fix the problem itself.

Thank you very much.

tuxmaster5000:
The problem will occurs on system boot. Starting the director after by hand via systemctl works. Looking in the journal shows, that using network as target will start the daemon to early, because on that stage the interfaces are up, but not configured. On F30 the interfaces are configured(the ip's are set) after the network-online target. So the director can't bind on the ip addresses, on which it must listen.

pstorz:
What version of bareos are you exactly using? 
If it is not master, can you verify that the problem still exists with the current master packages here:
https://download.bareos.org/bareos/experimental/nightly/Fedora_30/

tuxmaster5000:
I use the release for 18. (18.2.6)

pstorz:
My tests show that this problem does not exist in the current master branch.

I created a bug report in our bugtracker:
https://bugs.bareos.org/view.php?id=1136

And will close this PR.

PR #331, opened 2019-11-11
dev/franku/master/threadlist fix rebased

franku:
- refactor the thread list in order to have an atomic insertion into the list 
- detach the created thread in a try-catch block
- fixed some clang-tidy issues
- added another test
- renamed variables and functions more meaningful
- refactore the jcr-set/get thread specific data code to reduce #ifdef hell

### DIFF
--- a/core/src/lib/thread_list.cc
+++ b/core/src/lib/thread_list.cc
@@ -20,39 +20,46 @@
 */
 
 #include "include/bareos.h"
-#include "thread_list.h"
 #include "include/jcr.h"
 #include "include/make_unique.h"
 #include "lib/berrno.h"
 #include "lib/bsock.h"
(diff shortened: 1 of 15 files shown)


core/src/lib/thread_list.cc
```diff
@@ -20,39 +20,44 @@
...
-  std::size_t maximum_thread_count_ = 0;
+class ThreadListPrivate {
+  friend class ThreadList;
+  friend class ThreadGuard;
+
+ public:
+  using ThreadListUniqueLock = std::unique_lock<std::mutex>;
+
+ private:
+  std::size_t maximum_thread_count_{};
```
arogge:
```suggestion
  std::size_t maximum_thread_count_{0};
```

PR #335, opened 2019-11-13
bnet: check quit variable every second

franku:
**bnet server tcp:** 
- poll and select will timeout every second so the surrounding loop can break
- do not send a TIMEOUT_SIGNAL to the task


franku:
Manually retriggered CI Tests, which in turn succeed.

PR #337, opened 2019-11-15
fd-plugins: Check return value of endRestoreFile

sduehr:
When omitting return bRCs['bRC_OK'] in the end_restore_file method
of a Python plugin, it resulted in seemingly unrelated Python
errors like "TypeError: an integer is required" in subsequent calls
of Python funtions.

Now the return value is checked and if it is invalid, an appropriate
error message will be generated.

### DIFF
--- a/core/src/filed/fd_plugins.cc
+++ b/core/src/filed/fd_plugins.cc
@@ -1113,6 +1113,7 @@ bool PluginNameStream(JobControlRecord* jcr, char* name)
   char* cmd;
   char* p = name;
   bool start;
+  bool retval = true;
   bpContext* ctx = nullptr;
   alist* plugin_ctx_list;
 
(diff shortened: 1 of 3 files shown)


core/src/filed/fd_plugins.cc
```diff
@@ -1140,7 +1141,15 @@ bool PluginNameStream(JobControlRecord* jcr, char* name)
       Dmsg2(debuglevel, "End plugin data plugin=%p ctx=%p\n", plugin,
             jcr->plugin_ctx);
       if (b_ctx->restoreFileStarted) {
-        PlugFunc(plugin)->endRestoreFile(jcr->plugin_ctx);
+        /* PlugFunc(plugin)->endRestoreFile(jcr->plugin_ctx); */
+        bRC ret = PlugFunc(plugin)->endRestoreFile(jcr->plugin_ctx);
+        Dmsg1(0, "endRestoreFile ret: %d\n", ret);
+        if (ret < 0) {
```
franku: According to the bRC enum in lib/plugins.h this value cannot be less than zero.

sduehr: Changed as suggested.

PR #338, opened 2019-11-20
dird: Propagate RunScripts from JobDefs

adrb:
Copy RunScripts when propagate settings to JobDef from other JobDef.

### DIFF
--- a/core/src/dird/dird_conf.cc
+++ b/core/src/dird/dird_conf.cc
@@ -2605,7 +2605,7 @@ bool PropagateJobdefs(int res_type, JobResource* res)
    */
   if (res->jobdefs == res) { return false; }
 
-  if (res_type == R_JOB) {
+  if (res_type == R_JOB || res_type == R_JOBDEFS) {
     jobdefs = res->jobdefs;
 
     /*


arogge:
Hi,
thank you very much for your effort.

Could you describe what problem you're facing that this PR fixes? Is there an issue on bugs.bareos.org for that problem?

Best Regards,
Andreas

joergsteffens:
Is it related to https://bugs.bareos.org/view.php?id=886 ?

adrb:
Hi,

No it's not related to any bug report I know of. I have bunch of separate environments with similar configuration. So I create "two layer" of JobDefs definitions where on the first "layer" I have settings common for all of those environments and RunScripts is one of them. I notice that it wasn't propagated down to actual Job

arogge:
We're going to accept this. However, we will need an automated test for the behaviour first and it might take some time until someone can do this, so please be patient.

adrb:
Sounds great, thanks for the update :)

joergsteffens:
I implemented a systemtest for this in https://github.com/joergsteffens/bareos/tree/dev/joergs/master/systemtests-jobdefs-runscripts

However, while I see that your change does improve the handling in some cases, I found out, that JobDefs inheritance itself only works partly. I created a ticket for this, see https://bugs.bareos.org/view.php?id=1164

So, before integrating your PR, we first have to solve the general JobDefs problem.

arogge:
Hi!
your PR has been on hold for quite a long time. We have discussed how we want to handle the issue you're facing and decided that we will need to redesign a lot of the current code that does the propagation to do it right.

Therefore I would like to thank you for making us look into this problem and realizing that there is a lot of related issues surrounding yours. Anyway, I'm going to close the PR, as it doesn't provide the solution we are looking into right now.

Stay safe and healthy!

Best Regards,
Andreas

PR #347, opened 2019-11-26
docs: Remove version information from conf.py (backport to 18.2)

arogge:
This doesn't work with the new versioning scheme and never worked
completely accurately when using sphinx-versioning.


arogge:
I'm going to close the PR, because something went wrong on github with this.

PR #352, opened 2019-11-27
Fix restore objects with sqlite

sduehr:
Due to wrong calculation of the base64 encoded length, the last
byte of each restore object was overwritten with null. A function
was added that returns the correct length now.

Additionally the system test python-fd-plugin-local-fileset-test was
enhanced so that it now also tests if restore objects work correctly.

### DIFF
--- a/core/src/lib/base64.cc
+++ b/core/src/lib/base64.cc
@@ -226,7 +226,7 @@ int Base64ToBin(char* dest, int dest_size, char* src, int srclen)
  *
  *  Returns: Length of unpadded base64
  */
-int Base64LengthUnpadded(const int source_length)
+int Base64LengthUnpadded(int source_length)
 {
   if (source_length == 0) { return 0; }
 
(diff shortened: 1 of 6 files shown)


core/src/lib/base64.cc
```diff
@@ -219,3 +219,22 @@ int Base64ToBin(char* dest, int dest_size, char* src, int srclen)
...
+ */
+int Base64LengthUnpadded(const int source_length)
+{
+  if (source_length == 0) { return 0; }
+
+  int quotient = source_length / 3;
+  const int remainder = source_length % 3;
+  if (remainder > 0) { ++quotient; }
+  int unpadded_length = 4 * quotient;
+  if (remainder > 0) { unpadded_length -= 3 - remainder; }
```
arogge: While the operator precedence is  well defined, `unpadded_length -= (3 - remainder)` would make it unambiguous to the reader at first sight (and not only the compiler)

core/src/lib/base64.h
```diff
@@ -36,3 +36,4 @@ int ToBase64(int64_t value, char* where);
 int FromBase64(int64_t* value, char* where);
 int BinToBase64(char* buf, int buflen, char* bin, int binlen, bool compatible);
 int Base64ToBin(char* dest, int destlen, char* src, int srclen);
+int Base64LengthUnpadded(const int source_length);
```
arogge: The function is passed a copy of the int anyway, so making it `const` only affects the implementation, so we usually don't qualify by-value parameters with `const`.
```suggestion
int Base64LengthUnpadded(int source_length);
```

PR #356, opened 2019-12-02
Remove public-key for Univention from Repository

arogge:
This PR renames the bareos-com-release.key to bareos-release.key. The previous wording was misleading, because there is no difference between the keys for the bareos.org and the bareos.com release.
The PR also removes the bareos-release.key file as it must contain the public-key with which the binaries are going to be signed. This file is now injected by Jenkins during the build process. People who build their own packages will probably need to change the repository URL and the key anyway, so this should not hurt too much.


core/platforms/univention/CMakeLists.txt
```diff
@@ -57,7 +57,7 @@ install(
 )
 
 install(
-  FILES "bareos-com-release.key"
+  FILES "bareos-release.key"
```
joergsteffens:
```suggestion
  # This file contains the Bareos repository signature public.
  # During the automatic build process, It will be added.
  # For manual builds, download the file from 
  # http://download.bareos.org/bareos/release/<RELEASE>/<DISTRIBUTION>/Release.key
  FILES "bareos-release.key"
```

arogge:
That's a really good point.
Do you have any idea where we could put that information so people find it?

PR #358, opened 2019-12-02
dird: fix RunScript parsing

joergsteffens:
This commit fixes a bug introduced during the recent months.
Without this patch, all RunScripts are configured to "Runs On Client = Yes".
Because of this, Admin and Console RunScripts did not work and are silently ignored.

With this change, RunScripts can by set to run locally or on the Client (Runs On Client = Yes/No).

A systemtest for this is included in PR #314 .

### DIFF
--- a/systemtests/tests/python-bareos-test/python-bareos-unittest.py
+++ b/systemtests/tests/python-bareos-test/python-bareos-unittest.py
@@ -564,7 +564,7 @@ class PythonBareosJsonBase(PythonBareosBase):
                 u'Failed to find resource {} in {}.'.format(
                     resourcename, resourcesname))
 
-    def run_job(self, director, jobname, level, wait=False):
+    def run_job(self, director, jobname, level=None, wait=False):
         logger = logging.getLogger()
         run_parameter = ['job={}'.format(jobname), 'yes']
         if level:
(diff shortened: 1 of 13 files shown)


franku:
I cannot see why or where there is a bug introduced. The comment in the source that mentions the behaviour "Run on client by default" and the according line of code is present at least since the Bareos 12.4.1 release. This is what your PR changes. 

Since we handle long established default behaviour very seriously, I do not recommend to merge your PR. 

However, thanks for your contribution. What we should do next is:
- please come to our developer meeting 
- tell why you think this is a bug and exactly when it was introduced
- if we agree this is a bug we're going to fix it

joergsteffens:
Well, nevertheless it is a bug. I see, that you don't approve it, but why closing this PR already?
In the past, the line `if (res_runscript->target.empty()) ` have been `if (!res_runscript->target) ` and target have been a pointer, not a `std::string`. So there have been a difference between nullprt, emty string ("") or string with content. Also I've added a systemtest to check this behavior in PR #314, to verify this behavior.

joergsteffens:
As discussed: rebased on master, add systemtest.

systemtests/tests/python-bareos-test/python-bareos-unittest.py
```diff
@@ -640,6 +640,31 @@ def _test_list_with_invalid_jobid(self, director, jobid):
...
+
+        result = director.call('list joblog jobid={}'.format(jobId))
+        joblog_list = result['joblog']
+
+        found = False
+        for pattern in pattern_dict:
+            for logentry in joblog_list:
+                if re.search(pattern, logentry['logtext']):
+                    pattern_dict[pattern] = True
+            self.assertTrue(pattern_dict[pattern], 'Failed to find pattern "{}" in Job Log of Job {}.'.format(pattern, jobId))
```
franku: The test (also, as many others) relies on the fact that a job message in the source code has this expected pattern. During a code refactoring this is not always obvious. Therefore, would you like to add a comment to that part in the code to point out the fact?

joergsteffens: Thank you for your review. IMHO this bugs affects almost all users, as the CatalogBackup job of the Bareos default installation uses local RunScripts. Also the Bareos documentation delivers examples for Admin jobs for some tasks.

franku: Yes, I understand it should run on the director. 

As mentioned above, would you add an additional comment in the source code that mentions the dependency with the test?

joergsteffens:
Implemented the requested changes and rebased on master.

PR #359, opened 2019-12-02
dird: add Job warning if Admin job contains invalid RunScripts

joergsteffens:
Currently, RunScripts in Admin Jobs configured with "RunsOnClient = Yes" (or no RunsOnClient setting at all, as this is the default) have been silently ignored.
It has been hard, to figure out if the RunScript did run or not, as there has been no hint in the job log.

With this change, an error job message will be generated
and the job will be set to "Admin Error".

### DIFF
--- a/core/src/lib/runscript.cc
+++ b/core/src/lib/runscript.cc
@@ -134,40 +134,48 @@ int RunScripts(JobControlRecord* jcr,
           NSTDPRNT(script->command));
     runit = false;
 
-    if ((script->when & SCRIPT_Before) && (when & SCRIPT_Before)) {
-      if ((script->on_success &&
-           (jcr->JobStatus == JS_Running || jcr->JobStatus == JS_Created)) ||
-          (script->on_failure &&
-           (JobCanceled(jcr) || jcr->JobStatus == JS_Differences))) {
-        Dmsg4(200, "runscript: Run it because SCRIPT_Before (%s,%i,%i,%c)\n",
-              script->command.c_str(), script->on_success, script->on_failure,
-              jcr->JobStatus);
-        runit = true;
+    if (!script->IsLocal()) {
+      if (jcr->is_JobType(JT_ADMIN)) {
+        Jmsg(jcr, M_WARNING, 0,
+             "Invalid runscript definition (command=%s). Admin Jobs only "
+             "support local runscripts.\n",
+             script->command.c_str());
+      }
+    } else {
+      if ((script->when & SCRIPT_Before) && (when & SCRIPT_Before)) {
+        if ((script->on_success &&
+             (jcr->JobStatus == JS_Running || jcr->JobStatus == JS_Created)) ||
+            (script->on_failure &&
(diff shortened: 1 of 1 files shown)


core/src/lib/runscript.cc
```diff
@@ -134,41 +134,49 @@ int RunScripts(JobControlRecord* jcr,
...
-    if ((script->when & SCRIPT_Before) && (when & SCRIPT_Before)) {
-      if ((script->on_success &&
-           (jcr->JobStatus == JS_Running || jcr->JobStatus == JS_Created)) ||
-          (script->on_failure &&
-           (JobCanceled(jcr) || jcr->JobStatus == JS_Differences))) {
-        Dmsg4(200, "runscript: Run it because SCRIPT_Before (%s,%i,%i,%c)\n",
-              script->command.c_str(), script->on_success, script->on_failure,
-              jcr->JobStatus);
-        runit = true;
+    if (!script->IsLocal()) {
```
franku: As far as I can see, currently, every runscript is by default not local (runscript->target is not empty). As a matter of fact this statement is always true and the following else won't be reached.

core/src/lib/runscript.cc
```diff
@@ -134,41 +134,49 @@ int RunScripts(JobControlRecord* jcr,
...
+             (JobCanceled(jcr) || jcr->JobStatus == JS_Differences))) {
+          Dmsg4(200, "runscript: Run it because SCRIPT_After (%s,%i,%i,%c)\n",
+                script->command.c_str(), script->on_success, script->on_failure,
+                jcr->JobStatus);
+          runit = true;
+        }
       }
     }
 
-    if (!script->IsLocal()) { runit = false; }
```
franku: As well as the code above removing this line changes the default behaviour.

franku:
As this PR is dependent on #358 and changes default behaviour of the way runscripts are handled I would not accept it without urgent need. Please understand that I cannot accept his change until we have a solution for #358. Thanks for contribution anyway.

joergsteffens:
The current incorrect behavior is also described in https://bugs.bareos.org/view.php?id=419.

joergsteffens:
This PR is not directly related to #358 and in essence only add one code line for a job message warning.

joergsteffens:
Adapted change as internal discussed. Do not set the Admin Job to error, just add a warning to the joblog.

core/src/lib/runscript.cc
joergsteffens: No, as discussed in PR #358.

joergsteffens:
As said before, in essence this change only adds one code line for a job message warning.
The reason why the diff show many modified lines is that the if statement "if (!script->IsLocal())" have been moved from being the last check to being the first check, to better represent the indented logic. Therefore the indentation of the code changed.

core/src/lib/runscript.cc
franku: This moved to the if clause in line 137. -> resolved

PR #360, opened 2019-12-04
cats: Update DB schema from 2171 to 2192

fbergkemper:
- Add an INDEX on Job.JobTDate to improve performance (MySQL only)
- Fix integer out of range error by changing BaseFiles.BaseId to BIGINT
- Add new DDL update scripts for each supported DB backend
    
Fixes #1088: Integer out of range when using large amounts of files with Base Jobs

Note:

Only the RENAME TABLE, ADD COLUMN, and RENAME COLUMN variants of the ALTER TABLE command are supported in SQLite. Other kinds of ALTER TABLE operations such as DROP COLUMN, ALTER COLUMN, ADD CONSTRAINT, and so forth are omitted.

As the SQLite DB backend is only for testing and development purposes and we do not want to fiddle around with such limitations, DDL update scripts for SQLite will no longer be provided.


joergsteffens:
I fear, that package updates on Debian with sqlite backend will fail, as there is no update schema script. I've not tested it, that this could make package updates impossible without fully deinstall Bareos and reinstalling it again.

sduehr:
> I fear, that package updates on Debian with sqlite backend will fail, as there is no update schema script. I've not tested it, that this could make package updates impossible without fully deinstall Bareos and reinstalling it again.

Considering that sqlite should be used for testing only, why wouldn't it be ok to reinstall?
Alternative would be to provide an sqlite schema update script but leave the BaseFiles.BaseId column as is (INTEGER), as a Bareos test installation with sqlite should never be used with such a lot of big basejobs and for such a long time that BaseId values would overflow INTEGER.

fbergkemper:
> > I fear, that package updates on Debian with sqlite backend will fail, as there is no update schema script. I've not tested it, that this could make package updates impossible without fully deinstall Bareos and reinstalling it again.
> 
> Considering that sqlite should be used for testing only, why wouldn't it be ok to reinstall?
> Alternative would be to provide an sqlite schema update script but leave the BaseFiles.BaseId column as is (INTEGER), as a Bareos test installation with sqlite should never be used with such a lot of big basejobs and for such a long time that BaseId values would overflow INTEGER.

Good suggestion, I've also thought about this solution.

fbergkemper:
I've added the SQLite DDL update script core/src/cats/ddl/updates/sqlite3.2171_2192.sql .

PR #367, opened 2019-12-09
docs: Update Installing Bareos

fbergkemper:


docs/manuals/source/IntroductionAndTutorial/InstallingBareos.rst
```diff
@@ -104,7 +104,7 @@ Bareos :sinceVersion:`15.2.0: requires: jansson` requires the :ref:`Jansson libr
    # DIST=Fedora_26
    # DIST=Fedora_25
```
joergsteffens: These DISTs are not the distributions delivered with Bareos 19.2. Same applies to the other sections.

fbergkemper:
> Looks ok, but you might want to remove hints to older releases. Please see also the patch I sent you as private message with some minor changes on the text.

Applied your patch, thank you.

docs/manuals/source/IntroductionAndTutorial/InstallingBareos.rst
fbergkemper: I modified the shell script examples to use recent dist versions, thanks for the review.

PR #373, opened 2019-12-15
core: reword how to cancel a command

fbergkemper:


core/src/console/console.cc
```diff
@@ -1124,7 +1124,7 @@ int main(int argc, char* argv[])
 
   Dmsg0(40, "Opened connection with Director daemon\n");
 
-  ConsoleOutput(_("\nEnter a period to cancel a command.\n"));
+  ConsoleOutput(_("\nEnter a period (.) to cancel a command.\n"));
```
franku:
```suggestion
  ConsoleOutput(_("\nEnter a period \".\" to cancel a command.\n"));
```

fbergkemper: The format (.) was already used in another place.

See:

dird/ua_run.cc:      ua->SendMsg(_("Enter a period (.) to stop this test\n"));

PR #375, opened 2019-12-17
dev/franku/master/config reload test duplicate job

franku:
This is the first systemtest ported from the regression "reload-test" that tries to reload a config with a duplicated Job.


franku:
Will chop that test into two.

PR #376, opened 2019-12-17
dird: Change the volume selection dialog text

fbergkemper:
Change the volume selection dialog text,
to clarify how the asterisk is used.


core/src/dird/ua_select.cc
```diff
@@ -852,7 +852,8 @@ bool SelectMediaDbr(UaContext* ua, MediaDbRecord* mr)
     mr->PoolId = pr.PoolId;
     ua->db->ListMediaRecords(ua->jcr, mr, NULL, false, ua->send, HORZ_LIST);
 
-    if (!GetCmd(ua, _("Enter *MediaId or Volume name: "))) { goto bail_out; }
+    ua->SendMsg(_("Enter the volume name or MediaId of the volume prefixed with asterisk (*).\n"));
```
arogge:
```suggestion
    ua->SendMsg(_("Enter the volume name or MediaId of the volume prefixed with an asterisk (*).\n"));
```

PR #379, opened 2019-12-18
findlib: parent dirs should have consistent mode

arogge:
Previously when creating parent directories Bareos handled the deepest
parent directory special and set permissions on it.
This surfaced when selecting only files, but not directories to restore.
In that case the directory that contains the files got a mode that was
derived from the file mode of the file being restored when the parent
directory was created.
While this in itself is already inconsistent, the effective mode was
also calculated by adding S_IWUSR and S_IXUSR which did not take group
or other in account.

This patch now makes sure that no permission on parent directories is
set explicitly, so you just get what the effective uid, gid and umask
will produce.

### DIFF
--- a/core/src/findlib/mkpath.cc
+++ b/core/src/findlib/mkpath.cc
@@ -257,7 +257,7 @@ bool makepath(Attributes* attr,
   /*
    * Set for final component
    */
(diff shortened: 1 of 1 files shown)


arogge:
We'll need to discuss whether this is the right approach and to what previous releases this fix should be backported.

PR #381, opened 2019-12-18
python-bareos: add tls support without psk

bigzbigz:
A pull request is proposed to fix the issue [0001155](https://bugs.bareos.org/view.php?id=1155) :

- Add a 2 new parameters _tls_psk_enable_ and _tls_psk_require_ to bareos python module in order to enable the connection to a director when _TLS Key_ and _TLS Certificate_ are configured


bigzbigz:
it seems that the problem has been fixed in joergsteffens commits in last days

PR #385, opened 2019-12-30
copy value of seq to the stack to avoid race condition when generating Job names

WaryWolf:
Fixes an issue where two jobs that were created at the same time could share a unique job name, which caused various issues.


WaryWolf:
Sorry, I seem to have messed something up by force-pushing the amended commit with the updated AUTHORS file. I can't see the changes you requested - hopefully the amended commit is what you wanted.

Have a good weekend!

PR #388, opened 2020-01-08
systemtests: do not killall in systemtests

franku:
- otherwise systemtests cannot run in parallel


pstorz:
These changes are included in #392

PR #391, opened 2020-01-10
added `configure delete <resource>`

benibr:
This PR add the functionality to delete a resources and it's corresponding configfile via bconsole.
This is particulary useful if Bareos is confirgured automatically via "API".

If a resource should be deleted the code first tries to guess its config file name and then tries to remove() it, if that fails the command is aborted. Then the resource is removed from the current configuration. Of course, like if one removes a file manually, the config reload could be impossible due to a missing resource afterwards.

### DIFF
--- a/core/src/dird/dird_conf.cc
+++ b/core/src/dird/dird_conf.cc
@@ -1033,6 +1033,10 @@ const char* GetUsageStringForConsoleConfigureCommand()
       }
     }
     /*
+     * subcommand: delete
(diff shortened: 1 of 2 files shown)


franku:
As discussed on the phone: 
- this feature is a little "tricky" because it can easily produce an inconsistent config
- the current config parser will be replaced in the midterm to make this or a similar feature possible

Thank you again for your contribution.

PR #402, opened 2020-01-24
oVirt-Plugin: Credentials from File, include/exclude also on restore

sduehr:
The Plugin now allows to optionally use a config file for the
credentials instead of using the plugin options username and
password.

The include_disk_aliases and exclude_disk_aliases options are
now also usable on restore. On backup, the excluded disks are
already excluded from the snapshot.

### DIFF
--- a/core/src/plugins/filed/BareosFdPluginOvirt.py
+++ b/core/src/plugins/filed/BareosFdPluginOvirt.py
@@ -18,6 +18,7 @@ import io
 import time
 import uuid
 import json
+import ConfigParser as configparser
 
 import lxml.etree
 
@@ -58,7 +59,7 @@ class BareosFdPluginOvirt(BareosFdPluginBaseclass.BareosFdPluginBaseclass):
         if self.mandatory_options is None:
             self.mandatory_options = []
 
-        # self.mandatory_options.extend(['vmname','storage_domain'])
+        self.config = None
 
         self.ovirt = BareosOvirtWrapper()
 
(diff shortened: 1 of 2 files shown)


docs/manuals/source/TasksAndConcepts/Plugins.rst
```diff
@@ -627,6 +628,13 @@ When performing restores, the plugin can do one of the following:
 * Create a new VM with new disks
 * Overwrite existing disks of an existing VM
 
+Additionally it is possible to
+
+* Skip disks by alias names using include/exclude
+* Restoring the VM only without any disks is possible by
+  * Selecting to restore only the **.ovf** file
+  * Excluding all disks by alias
```
pstorz: This Line and the one before are mit rendered correct. See https://download.bareos.org/bareos/experimental/CD/PR-402/BareosMainReference/TasksAndConcepts/Plugins.html?highlight=ovirt

core/src/plugins/filed/BareosFdPluginOvirt.py
```diff
@@ -614,6 +640,75 @@ def restore_object_data(self, context, ROP):
...
+                "BareosFdPluginOvirt: Error reading config file %s: %s\n"
+                % (self.options["config_file"], err.strerror),
+            )
+            return False
+
+        return self.check_config(context)
+
+    def check_config(self, context):
+        """
+        Check the configuration and set or override options if necesary,
```
pstorz: Typo: necessary

PR #408, opened 2020-02-04
Release year for 19.2.4 and 19.2.5 is off by one (2019 instead of 2020)

dilruacs:


pstorz:
I have cherry-picked this change also to master, changes will be online later today.

Thank you

PR #410, opened 2020-02-06
Fix Bug #1175: bareos crashes with invalid character in ACL

arogge:
Bareos crashed when an invalid ACL syntax was detected. This patch adds a systemtest to check for this behaviour and a fix for the problem.

### DIFF
--- a/core/src/tests/test_acl_entry_syntax.cc
+++ b/core/src/tests/test_acl_entry_syntax.cc
@@ -28,14 +28,15 @@
 
 #include "lib/edit.h"
 
-POOLMEM* msg = GetPoolMemory(PM_FNAME);
 
 TEST(acl_entry_syntax_test, acl_entry_syntax_test)
 {
+  std::vector<char> msg;
+
   EXPECT_EQ(true, IsAclEntryValid("list", msg));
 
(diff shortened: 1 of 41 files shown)


systemtests/tests/config-syntax-crash/testrunner
```diff
@@ -0,0 +1,24 @@
+#!/bin/sh
+#
+# Run a simple backup
```
franku: A comment could help to understand that the config does not crash the director because of the invalid ACL Syntax.

franku: The unittest looks reasonable.

core/src/tests/test_acl_entry_syntax.cc
```diff
@@ -28,31 +28,32 @@
 
 #include "lib/edit.h"
 
-POOLMEM* msg = GetPoolMemory(PM_FNAME);
 
 TEST(acl_entry_syntax_test, acl_entry_syntax_test)
 {
+  std::vector<char> msg{};
```
franku:
```suggestion
  std::vector<char> msg;
```

PR #412, opened 2020-02-07
Fix Bug 1180: CRC checksum algorithm changed between 18.2.7 and 19.2.5, all volumes fail

arogge:
On FreeBSD the new CRC library misdetected endianess. This lead to the big endian variant of the algorighm being used on little endian machines.
While this is still a worthy checksumming algorithm it doesn't lead to the same result as the previous one did.

As this worked its way through automated testing, this PR updates the automated tests for crc32 to catch such an error.
This PR also fixes the underlying problem by adding detection for the endianess on FreeBSD.


franku:
You may want to move the variables to their scope, this would increase readability. Why malloc and not use a std::array?

PR #414, opened 2020-02-10
dev/franku/master/messages resource

franku:
Combined system-/unittest to check output of the configurable message destinations.


pstorz:
I would suggest to move "tests/messages.cc" to "tests/messages_resource_test.cc"
See the names of the other files in the tests directory.

PR #415, opened 2020-02-10
Update BareosFdPluginOvirt.py

hamazasalah:
Fix Ovirt Plugin "Backup a VM by using VM ID"


sduehr:
Thanks for your fix, I just verified that it works. It was a bug indeed.

But as Andreas mentioned, please change your commit message. This article may help you how to do that:
https://help.github.com/en/github/committing-changes-to-your-project/changing-a-commit-message

PR #419, opened 2020-02-11
Fix language RU

snuglinux:
Corrected an incorrect translation and supplemented the missing one for the Russian language.


snuglinux:
Please add my translation improvements

fbergkemper:
Thanks for the pull request. 

Please exclude changes to boostrap-table-locale-all.min.js from this pull request. The file is part of a bootstrap-table extension we make use of. Instead propose those changes to the upstream project, see https://github.com/wenzhixin/bootstrap-table .

Comments on https://github.com/bareos/bareos/pull/419/commits/112646bf5d89c703912ad07b2057fcda95431808#diff-9d2ad1fab3ec027c2efe34379ed0550e will follow.

Regards,
Frank

fbergkemper:
Thank you again for your effort.

With regard to the proposed adjustments to the translations in the PO file, it would be easier for both sides to use the integration via the external service http://poeditor.com to verify the translation.

You can join the translation team using the following link, https://poeditor.com/join/project/ELnLNbvQJb.

This is also described in https://github.com/bareos/bareos/blob/master/webui/doc/README-TRANSLATION.md.

If you need further help, please let us know.

I apologize that this unfortunately leads to double work.

Best regards
Frank

snuglinux:
Thank you, I will register at poeditor.com, I can also make a translation into Ukrainian, which is currently not available in bareos

fbergkemper:
> Thank you, I will register at poeditor.com, I can also make a translation into Ukrainian, which is currently not available in bareos

That would be great but would require a bit of preparation on our part to make it available in the online tool which I would like to discuss with our team first.

Please get in touch with us via the devel@googlegroups.com mailing list regarding the Ukrainian translation. It is a better place to discuss further than the comment section of this pull request, which I'm going to close now.

Thanks and regards
Frank

fbergkemper:
Sorry posted the wrong mailing list link in my previous comment.

Here is the correct one: bareos-devel@googlegroups.com

PR #430, opened 2020-02-22
scheduler: run jobs without client (copy/migrate)

franku:
- fixes a bug
- add another unittest that tests if a scheduled copy job runs, hence without client

### DIFF
--- a/core/src/dird/scheduler_private.cc
+++ b/core/src/dird/scheduler_private.cc
@@ -127,7 +127,6 @@ static void SetJcrFromRunResource(JobControlRecord* jcr, RunResource* run)
 JobControlRecord* SchedulerPrivate::TryCreateJobControlRecord(
     const SchedulerJobItem& next_job)
 {
-  if (next_job.job->client == nullptr) { return nullptr; }
   JobControlRecord* jcr = NewDirectorJcr();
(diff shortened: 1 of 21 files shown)


core/src/dird/scheduler_private.cc
```diff
@@ -127,7 +127,6 @@ static void SetJcrFromRunResource(JobControlRecord* jcr, RunResource* run)
 JobControlRecord* SchedulerPrivate::TryCreateJobControlRecord(
     const SchedulerJobItem& next_job)
 {
-  if (next_job.job->client == nullptr) { return nullptr; }
   JobControlRecord* jcr = NewDirectorJcr();
   SetJcrDefaults(jcr, next_job.job);
```
arogge: `SetJcrDefaults()` already has checks for `job->client` in place, so it should be safe to assume it may be `nullptr`.

arogge: `SetJcrFromRunResource()` does not touch `job->client` at all

arogge: The only additional requirement seems to be that `ExecuteJobCallback_()` needs to be able to handle a jcr with no client set, which is an assumption we can probably make.

franku:
Force pushed the Fixes comment.

PR #431, opened 2020-02-25
Fix issue 1192: Authorization key rejected by Storage daemon since upgrading director and storage daemons

arogge:
See https://bugs.bareos.org/view.php?id=1192


franku:
Very good you found that!

PR #436, opened 2020-02-26
Fix compile errors on Fedora 31

tuxmaster5000:
Without it, the file can't build under Fedora 31.


franku:
If you are adding this headerfile to a not yet existing, new compile unit your suggestion is reasonable. Would you mind to tell what's your purpose?

PR #441, opened 2020-03-02
Adapt debian build to work in obs and our docker build-environment

arogge:
This PR updates the control-files for the debian package build to allow building packages without open buildservice.
It also fixes a bug in the generation of dbconfig-common filenames.


arogge:
This breaks univention unless #443 is merged first.

PR #444, opened 2020-03-06
improve search and filtering on jobs bootstrap table (job module)

fbergkemper:
This is just a draft, refinements can follow.


webui/module/Job/view/job/job/index.phtml
```diff
@@ -167,19 +297,90 @@ $this->headTitle($title);
...
+
+      switch(row.jobstatus) {
+         case 'T':
+         case 'W':
+            switch(row.type) {
+               case 'B':
+                  return jobDetailsButton + '&nbsp;' + jobRerunButton + '&nbsp;' + jobRestoreButton;
+               case 'C':
+                  return jobDetailsButton + '&nbsp;' + jobRestoreButton;
+            }
```
franku:
```suggestion
            }
            break
```

franku: I would suggest a break here, if no fallthrough is intended otherwise.

fbergkemper: good point

PR #451, opened 2020-03-16
breg: fix crash when setting regexwhere

pstorz:
When manually setting the regular expression during restore like with:

OK to run? (yes/mod/no): mod
->11: File Relocation
->4: Enter a regexp
->Please enter a valid regexp (!from!to!): !what!with!
->6: Use this ?

The director crashed. This happens because the member regex_t preg of
the class BareosRegex was not zero initialized correctly.
This caused the freeing of random memory in void regfree(regex_t* preg).

Now all members of class BareosRegex are zero initialized and the
problem is fixed.

This problem also causes the filedaemon to crash when regexwhere is specified
in the restore job configuration. On the second restore run the filedaemon
crashes.

Fixes# 1211: bareos director and bareos fd crash when regexwhere is specified

### DIFF
--- a/core/src/lib/breg.h
+++ b/core/src/lib/breg.h
@@ -2,7 +2,7 @@
    BAREOS® - Backup Archiving REcovery Open Sourced
 
    Copyright (C) 2006-2006 Free Software Foundation Europe e.V.
-   Copyright (C) 2016-2016 Bareos GmbH & Co. KG
+   Copyright (C) 2016-2020 Bareos GmbH & Co. KG
 
    This program is Free Software; you can redistribute it and/or
    modify it under the terms of version three of the GNU Affero General Public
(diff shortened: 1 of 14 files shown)


core/src/lib/breg.h
```diff
@@ -68,7 +68,7 @@ class BareosRegex {
   /* private */
   POOLMEM* expr = nullptr;     /**< search epression */
   POOLMEM* subst = nullptr;    /**< substitution */
-  regex_t preg;                /**< regex_t result of regcomp() */
+  regex_t preg = {};           /**< regex_t result of regcomp() */
   regmatch_t regs[BREG_NREGS]; /**< contains match */
```
franku:
```suggestion
  regmatch_t regs[BREG_NREGS] {}; /**< contains match */
```

franku:
```suggestion
  regmatch_t regs[BREG_NREGS] = {}; /**< contains match */
```

franku: Both structs are declared as extern "C" and therefore no default constructor will be generated for them.

PR #455, opened 2020-03-24
Improve glusterfs backend

arogge:
* add logging options to glusterfs backend
* add a systemtest to test the glusterfs backend

### DIFF
--- a/core/src/stored/backends/gfapi_device.cc
+++ b/core/src/stored/backends/gfapi_device.cc
@@ -43,7 +43,9 @@ namespace storagedaemon {
 enum device_option_type
 {
   argument_none = 0,
-  argument_uri
+  argument_uri,
+  argument_logfile,
+  argument_loglevel
 };
 
 struct device_option {
@@ -53,6 +55,8 @@ struct device_option {
 };
 
 static device_option device_options[] = {{"uri=", argument_uri, 4},
+                                         {"logfile=", argument_logfile, 8},
+                                         {"loglevel=", argument_loglevel, 9},
                                          {NULL, argument_none}};
 
 /**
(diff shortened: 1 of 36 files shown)


core/src/stored/backends/gfapi_device.cc
```diff
@@ -338,20 +342,23 @@ int gfapi_device::d_open(const char* pathname, int flags, int mode)
...
             default:
+              Mmsg1(errmsg, _("Unable to parse device option: %s\n"), bp);
+              Emsg0(M_FATAL, 0, errmsg);
+              goto bail_out;
               break;
           }
+          done = true;
         }
       }
-
```
pstorz: Was this block intentionally removed?

arogge: If you look closely, you'll find it in the `switch()`'s default-case

core/src/stored/backends/gfapi_device.cc
```diff
@@ -338,20 +342,23 @@ int gfapi_device::d_open(const char* pathname, int flags, int mode)
           switch (device_options[i].type) {
             case argument_uri:
               gfapi_uri_ = bp + device_options[i].compare_size;
-              done = true;
+              break;
+            case argument_logfile:
+              gfapi_logfile_ = bp + device_options[i].compare_size;
+              break;
+            case argument_loglevel:
+              gfapi_loglevel_ =
```
franku: In case of error  gfapi_loglevel_ will be zero.

core/src/stored/backends/gfapi_device.h
```diff
@@ -40,6 +40,8 @@ class gfapi_device : public Device {
  private:
   char* gfapi_configstring_;
   char* gfapi_uri_;
+  char* gfapi_logfile_;
```
franku:
```suggestion
  char* gfapi_logfile_{};
```

franku: Should double check if the constructor is really called. Ohterwise this has no effect.

core/src/stored/backends/gfapi_device.cc
arogge: this is expected and not a problem.

PR #456, opened 2020-03-25
pre-fill restore form element "replace" according to selected restore job in the webui

fbergkemper:


core/src/lib/util.cc
```diff
@@ -584,6 +584,26 @@ const char* job_type_to_str(int type)
   return str;
 }
 
+const char* job_replace_to_str(int replace)
+{
+  const char* str = NULL;
```
franku:
```suggestion
  // remove the temporary variable and return the from each case level directly
```

core/src/lib/util.cc
```diff
@@ -584,6 +584,26 @@ const char* job_type_to_str(int type)
...
+      break;
+    case REPLACE_IFNEWER:
+      str = _("ifnewer");
+      break;
+    case REPLACE_IFOLDER:
+      str = _("ifolder");
+      break;
+    case REPLACE_NEVER:
+      str = _("never");
+      break;
```
franku:
```suggestion
      break;
    default:
      return "?? Unknown parameter";
    
```

core/src/lib/util.cc
```diff
@@ -584,6 +584,26 @@ const char* job_type_to_str(int type)
...
+  }
+  return str;
```
franku:
```suggestion
  // see comment above
```

core/src/lib/util.cc
fbergkemper: default case added

core/src/lib/util.cc
fbergkemper: For consistency, I have applied the style of the existing functions to the new one.

fbergkemper:
Thanks for the review.

PR #458, opened 2020-03-25
chunked_device: bugfix WaitUntilChunksWritten() (bareos-18.2)

joergsteffens:
Don't wait, if flushing failed.


core/src/stored/backends/chunked_device.cc
```diff
@@ -1266,6 +1266,7 @@ bool chunked_device::WaitUntilChunksWritten()
       if (!FlushChunk(false /* release */, false /* move_to_next_chunk */)) {
          dev_errno = EIO;
          retval = false;
+         return retval;
```
franku: You could eliminate the retval variable completely, this would be more clearly to read.

PR #459, opened 2020-03-25
droplet: only allow configuration maximum concurrent jobs of 0 or 1

joergsteffens:
There is a known bug when using Storage Backend Droplet with multiple simultaneous jobs (interleaving).
This workaround only allows settings of 0 or 1.
If the default is used (0), it will be set automatically to 1.
If any other value is specified, the bareos-sd will not start at all.


arogge:
Shouldn't the PR also update the Backend documentation and add information about this limitation?

PR #475, opened 2020-04-06
filed: round up to 512 bytes correctly for Windows/FreeBSD

franku:
In my opionion there is a roundup error when running backup on win/freebsd. However, there is no related bug, yet.

### DIFF
--- a/core/src/filed/backup.cc
+++ b/core/src/filed/backup.cc
@@ -1188,7 +1188,7 @@ static int send_data(JobControlRecord* jcr,
     /*
      * To read FreeBSD partitions, the read size must be a multiple of 512.
      */
-    bctx.rsize = (bctx.rsize / 512) * 512;
+    bctx.rsize = ((bctx.rsize + 512) / 512) * 512;
 #endif
   }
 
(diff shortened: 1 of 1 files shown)


core/src/filed/backup.cc
```diff
@@ -1188,15 +1188,15 @@ static int send_data(JobControlRecord* jcr,
     /*
      * To read FreeBSD partitions, the read size must be a multiple of 512.
      */
-    bctx.rsize = (bctx.rsize / 512) * 512;
+    bctx.rsize = ((bctx.rsize + 512) / 512) * 512;
```
arogge, line 1191: I guess `bctx.rsize` is the size of `bctx.rbuf`. Is that buffer allocated with enough space?

arogge: As far as I understand the code the previous version rounded down so rsize is a multiple of 512 and does not exceed its original size.

franku: I understand i.e.: 
(rsize=511 / 512) * 512 = 0  -> 511 Bytes too few
(rsize=513 / 512) * 512 = 512 -> 1 Byte too few

So I should check the allocation of bctx.rbuf, not round down to a smaller value.

franku: Now I understood from code-review, that bctx.rsize will be initialized by the maximum buffer length (that is multiple of 512) and rounded down to multiple of 512. The case with the result 0, therefore, won't happen.
bctx.rsize in turn is used to limit the copied amount of data into the buffer bctx.rbuf.

franku:
Will close as this does not emerge as a bug.

PR #479, opened 2020-04-09
general code-cleanup

arogge:
* migrate `host.h` and friends into `config.h`
* remove some unused definitions from `config.h` and their cmake counterparts
* allow usage of `FORTIFY_SOURCE` (which will generate a lot of additional warnings)
* introduce new macro `INTENDED_FALLTHROUGH`
* fix some old and new compiler warnings

When reviewing, looking at the individual commits is probably easiest.

TODO:
* remove even more definitions from `config.h`
* remove all unused checks from `BareosCheckFunctions.cmake`
* maybe move config.h into `CMAKE_BINARY_DIR`


core/CMakeLists.txt
```diff
@@ -385,8 +385,6 @@ set(LOCALEDIR \"${CMAKE_INSTALL_LOCALEDIR}\")
 
 set(HAVE_CRYPTO 1)
 
-include(BareosTypeSizes)
```
franku, line 395: This whole commit is one of the coolest I have seen on this code. ;-) Very good, now we have stdint.h.

core/CMakeLists.txt
```diff
@@ -745,12 +745,12 @@ message(" ")
...
 message(" ")
 
-message("   Host:                         ${host} -- ${DISTNAME} ${DISTVER} ")
+message("   Host:                         ${host} -- ${PLATFORM} ${DISTVER} ")
 message(
   "   Bareos version:               ${BAREOS} ${BAREOS_FULL_VERSION} (${DATE}) "
 )
 message(
-  "   Distribution:                 ${DISTNAME}"
+  "   Build platform:               ${PLATFORM}"
```
franku, line 753: "Build Host" should also be common.

core/src/lib/bsys.cc
```diff
@@ -217,23 +217,14 @@ char* bstrinlinecpy(char* dest, const char* src)
   if (src <= dest) { return NULL; }
 
   len = strlen(src);
-#if HAVE_BCOPY
```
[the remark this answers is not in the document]
arogge: Das `bstrlinecpy()` bestimmt die Länge des Strings, das macht `memmove()` natürlich nicht.

core/CMakeLists.txt
arogge: I thought this through. For me a "Build Host" is one specific system (i.e. "my laptop"), thus I ended up with "Build platform" which will end up containing "redhat" or "debian" or "solaris".

arogge:
Rebased to latest master and spell-checked.

PR #480, opened 2020-04-11
Fix ndmp link error on Fedora 31

tuxmaster5000:
Second fix for https://bugs.bareos.org/view.php?id=1170


arogge:
I thought it was something simple like that. Thank you for taking the time to look into it and thank you very much for the contribution!

arogge:
Also backported into bareos 19.2.

PR #481, opened 2020-04-13
Use system default ciphers like written here:

tuxmaster5000:
Don't use an self default cipher suite.
Instant use the system default one.
This will be default on RHEL and Fedora
https://docs.fedoraproject.org/en-US/packaging-guidelines/CryptoPolicies

### DIFF
--- a/core/src/lib/tls_openssl_private.cc
+++ b/core/src/lib/tls_openssl_private.cc
@@ -49,7 +49,7 @@ std::mutex TlsOpenSslPrivate::file_access_mutex_;
 /* No anonymous ciphers, no <128 bit ciphers, no export ciphers, no MD5 ciphers
  */
 const std::string TlsOpenSslPrivate::tls_default_ciphers_{
-    "ALL:!ADH:!LOW:!EXP:!MD5:@STRENGTH"};
+    "PROFILE=SYSTEM"};
 
 
 TlsOpenSslPrivate::TlsOpenSslPrivate()


tuxmaster5000:
Test on RHEL8 and Fedora

arogge:
Are you still interested in working on this? The PR doesn't build and is not ready to be merged, so if you don't want to continue the work, please close it.

tuxmaster5000:
I can't see, why it fails, because the test results are not public. So I can't do any more.
On my systems the patch to its job.

arogge:
Sorry, my bad. It fails with `Unknown CMake command "add_compile_definitions".` on RHEL/CentOS 8, all SUSE, Solaris, Debian and Ubuntu.
I'm not sure why it does that, but as I suggested using target_compile_definitions would probably be much better anyway. I'd also appreciate using the SPECfile to set the default.

tuxmaster5000:
That is very interesting, because I can build this on RHEL7+8.
What cmake version do you have in your environment?
Using the chroot from mock I see this versions:
RHEL7: cmake3 version 3.14.7
RHEL8: cmake version 3.11.4

arogge:
We're using the stock cmake on RHEL8, so it is 3.11.4 just like you're using.
On RHEL7 we use our own cmake 3.13.3 rpm (and your PR builds fine there).

tuxmaster5000:
When it is same version why it fails?
I simply build the the rpm package as is done the RedHat world using mock.

arogge:
If you need to reproduce the problem, I can give you a short overview how we build:
1. package the "core" directory into bareos-X.Y.Z.tar.bz2
1. take the specfile from core/platforms/packaging/bareos.spec
1. add open build service compatiblility macros
%rhel_version %{?rhel:%{rhel}00}
%centos_version %{?centos:%{centos}00}
%fedora_version %{?fedora}
1. rpmbuild -ba bareos.spec

There is no special magic here and I have really no idea how cmake can not know `add_compile_definitions`. However, I cannot take the time to look into this right now, so you're somewhat on your own. Sorry for that.

arogge:
btw. `add_compile_definitions` was added in CMake 3.12, so it should fail on your EL8 with cmake 3.11, too.

arogge:
Any news on this? I'd like to close the PR if you're not interested in working on it anymore.

tuxmaster5000:
It can be, that RHEL use a patched one.
My problem is, that without direct access to the test results, I can't adapt it the used cmake used by you. When the test results are public like on other projects, than I can modify the patch.

arogge:
We do not use a special cmake on CentOS 8 or RHEL 8. It is the stock RHEL cmake. It it too old and does not support the command you're using.
I just triple-checked. When I run `docker run --rm -it registry.redhat.io/ubi8/ubi dnf list cmake` it shows clearly that currently in RHEL8 there is only cmake-3.11.4-7.el8 (you will need a RHEL subscription on your docker-host to run this command).

I don't know what cmake version your mock buildroot for RHEL 8 uses, but if it works with your changes it is definitely not the vanilla one.

Nevertheless, as we're still building for Debian 8, everything is limited to cmake 3.5.0 anyway.

So to sum it up: do not `use add_compile_definitions` for the following reasons:
* it requires a too new version of cmake
* this kind of definition should be added per target and not globally (use `target_compile_definitions` instead)

Having said that, I would greatly appreciate if this was a settable cmake parameter that is passed by the SPEC and not determined by (rather ugly) autodetection in `CMakeLists.txt`.

PR #490, opened 2020-04-20
Fix building for newer openSUSE releases

bmwiedemann:
Because openSUSE Leap 15.2 and Tumbleweed dropped susefirewall.

also, Tumbleweed has newer glibc with split rpcgen, too

### DIFF
--- a/core/platforms/packaging/bareos.spec
+++ b/core/platforms/packaging/bareos.spec
@@ -63,7 +63,7 @@ BuildConflicts: libtirpc-devel
 %endif
 
 # fedora 28: rpc was removed from libc
-%if 0%{?fedora} >= 28 || 0%{?rhel} > 7
+%if 0%{?fedora} >= 28 || 0%{?rhel} > 7 || 0%{?suse_version} > 1500
 BuildRequires: rpcgen
 BuildRequires: libtirpc-devel
 %endif
@@ -78,9 +78,10 @@ BuildRequires: libtirpc-devel
 %define python_plugins 0
 %endif
 
-%if 0%{?suse_version} > 1010
+%if 0%{?suse_version} > 1010 && 0%{?suse_version} <= 1500
 %define install_suse_fw 1
 %define _fwdefdir   %{_sysconfdir}/sysconfig/SuSEfirewall2.d/services
+BuildRequires:  openSUSE-release
 %endif
 
 
@@ -941,7 +942,7 @@ for F in  \
     %{script_dir}/mtx-changer \
     %{_sysconfdir}/%{name}/mtx-changer.conf \
 %endif
-%if 0%{?install_suse_fw} == 0
+%if ! 0%{?install_suse_fw}
     %{_sysconfdir}/sysconfig/SuSEfirewall2.d/services/bareos-dir \
     %{_sysconfdir}/sysconfig/SuSEfirewall2.d/services/bareos-sd \
     %{_sysconfdir}/sysconfig/SuSEfirewall2.d/services/bareos-fd \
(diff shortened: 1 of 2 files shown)


arogge:
Thank you very much for the patch!
I will leave this open until our CI infrastructure can build openSUSE / SLE 15.2, so we can test it.

arogge:
I finally had a look at OpenSUSE Leap 15.2. It looks like we still have SUSEfirewall available. I wouldn't like to remove the service-files yet unless it hurts anyone.
For libtirpc-devel I'm wondering. The rpc components are still in glibc for 15.2 and we had issues with builds using both glibc-rpc and libtirpc, so would also refrain from changing that too early. Is the `0%{?suse_version} >= 1550` you used documented anywhere? It looks like "SLES / OpenSUSE 15.5", but are you sure (and have a source) that this change happens exactly there?

bmwiedemann:
https://build.opensuse.org/projects/openSUSE:Factory/prjconf has `%suse_version 1550`
but the [relevant documentation](https://en.opensuse.org/openSUSE:Build_Service_cross_distribution_howto) recommends `%if 0%{?suse_version} > 1500` - probably not much difference in practice.

Indeed, Leap 15.2 still has SuSEfirewall2 installable, but uses firewalld by default.

But if I enable it there, I get
```
error: File not found: /home/abuild/rpmbuild/BUILDROOT/bareos-19.2.7-0.x86_64/etc/sysconfig/SuSEfirewall2.d/services/bareos-dir
```

The build.log has no other mention of SuSEfirewall2

I guess, they are supposed to get installed via
`core/platforms/suse/CMakeLists.txt:set(FIREWALL_DIR ${sysconfdir}/sysconfig/SuSEfirewall2.d/services/)`

It starts to work, if I add
`BuildRequires:  openSUSE-release`
because that contains the `/etc/os-release` file used by cmake to detect that it is a SUSE OS.
However, default practice is to build packages without that or with `dummy-release` package or `distribution-release` Provides - but that does not do the trick with cmake.

Might be cleaner to rethink the cmake logic to check for existence of `/etc/sysconfig/SuSEfirewall2.d/services/`

arogge:
Thanks for taking the time to do this. We really appreciate it!

> https://build.opensuse.org/projects/openSUSE:Factory/prjconf has `%suse_version 1550`
> but the [relevant documentation](https://en.opensuse.org/openSUSE:Build_Service_cross_distribution_howto) recommends `%if 0%{?suse_version} > 1500` - probably not much difference in practice.

Point taken. I'd prefer the recommended way (that you now implemented).

> Indeed, Leap 15.2 still has SuSEfirewall2 installable, but uses firewalld by default.
> 
> But if I enable it there, I get
[...]
> Might be cleaner to rethink the cmake logic to check for existence of `/etc/sysconfig/SuSEfirewall2.d/services/`
While this is probably a lot cleaner I'm not sure it is actually worth it. We're building in docker-containers right now, so we always have /etc/os-release available. I'm not sure how to work around the problem, you're seeing and I think you will be  much better at proposing a fix than anyone of us is.
While you could check for SuSEfirewall2 directory in `platforms/suse/`, you have to make sure `$PLATFORM` is set correctly during the build. Otherwise the platform directory will not be picked up at all.

So if you really want to fix this, I believe the way to go is:
* make sure `$PLATFORM` is set to "suse" so `platforms/CMakeLists.txt` includes `platforms/suse/CMakeLists.txt`
* add an option in `platforms/suse/CMakeLists.txt` to enable/disable SuSEfirewall2 support and enable/disable it by adding the switch in the specfile

arogge:
As there hasn't been any progress on this on the past two months, I'll close the PR.
Feel free to reopen it if you want to continue working on it.

Thank you for your effort!

PR #492, opened 2020-04-21
add infrastructure for configuration warnings

arogge:
This PR will add a list of warnings to the configuration parser along with some utility functions. This list can then be filled during (and after) parsing the configuration.
It will also add a few ways to see these deprecation warnings.

### DIFF
--- a/core/src/dird/dird.cc
+++ b/core/src/dird/dird.cc
@@ -395,7 +395,16 @@ int main(int argc, char* argv[])
     goto bail_out;
   }
 
-  if (test_config) { TerminateDird(0); }
+  if (test_config) {
+    if (my_config->HasWarnings()) {
+      /* messaging not initialized, so Jmsg with  M_WARNING doesn't work */
+      fprintf(stderr, _("There are configuration warnings:\n"));
+      for (auto& warning : my_config->GetWarnings()) {
+        fprintf(stderr, " * %s\n", warning.c_str());
+      }
+    }
+    TerminateDird(0);
+  }
 
   if (!InitializeSqlPooling()) {
     Jmsg((JobControlRecord*)NULL, M_ERROR_TERM, 0,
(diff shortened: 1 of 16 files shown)


core/src/dird/dird.cc
```diff
@@ -395,7 +395,17 @@ int main(int argc, char* argv[])
     goto bail_out;
   }
 
-  if (test_config) { TerminateDird(0); }
+  if (test_config) {
+    if (my_config->HaveDeprecationWarnings()) {
+      // FIXME: cannot use Jmsg here as only M_FATAL and M_ERROR_TERM will
+      //        work because daemon_msgs is not initialized
+      fprintf(stderr, _("Deprecated configuration settings detected:\n"));
```
franku, line 400: When running as a daemon stderr could have been closed.

core/src/stored/status.cc
```diff
@@ -994,6 +1003,19 @@ static void sendit(PoolMem& msg, int len, StatusPacket* sp)
   }
 }
 
+static void sendit(const std::string& msg, StatusPacket* sp)
```
franku, line 1006: This general function is same for all three daemon. Could move it somewhere into libbareos.

core/src/dird/dird.cc
```diff
@@ -397,8 +397,6 @@ int main(int argc, char* argv[])
 
   if (test_config) {
     if (my_config->HasDeprecationWarnings()) {
-      // FIXME: cannot use Jmsg here as only M_FATAL and M_ERROR_TERM will
```
franku: A comment would be helpful anyway to understand why none of the Dmg or Jms can be used.

core/src/dird/dird.cc
arogge: `if (test_config)` gurarantees that we haven't daemonized, because `daemon_start()` is called in a `if (!test_config)`-block. And even if `daemon_start()` had already run stderr would be still open and point to `/dev/null` (see `SetupStdFileDescriptors()` in `core/src/lib/daemon.cc`.

core/src/stored/status.cc
arogge: Point taken. I should probably refactor the whole `sendit()`-family somewhere into the library.

core/src/dird/dird.cc
arogge: :+1:

PR #493, opened 2020-04-21
dird: make UA tree command auditing possible

fbergkemper:
1. Introduce a new boolean named audit_event in the struct cmdstruct
of the TreeContext to mark commands which require auditing.

2. Introduce a condition in the UserSelectFilesFromTree function to
call LogAuditEventCmdline if the given command wants an audit event.

### DIFF
--- a/core/src/dird/ua_audit.cc
+++ b/core/src/dird/ua_audit.cc
@@ -1,7 +1,7 @@
 /*
    BAREOS® - Backup Archiving REcovery Open Sourced
 
-   Copyright (C) 2014-2016 Bareos GmbH & Co. KG
+   Copyright (C) 2014-2020 Bareos GmbH & Co. KG
 
    This program is Free Software; you can redistribute it and/or
    modify it under the terms of version three of the GNU Affero General Public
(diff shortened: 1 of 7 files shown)


core/src/dird/ua_restore.cc
```diff
@@ -3,7 +3,7 @@
 
    Copyright (C) 2002-2011 Free Software Foundation Europe e.V.
    Copyright (C) 2011-2016 Planets Communications B.V.
-   Copyright (C) 2013-2019 Bareos GmbH & Co. KG
+   Copyright (C) 2013-2020 Bareos GmbH & Co. KG
```
franku, line 6: This should be placed in the next commit..

core/src/dird/ua_audit.cc
```diff
@@ -142,4 +142,20 @@ void UaContext::LogAuditEventCmdline()
   Emsg3(M_AUDIT, 0, _("Console [%s] from [%s] cmdline %s\n"), user_name, host,
         cmd);
 }
+
+void UaContext::LogAuditEventInfoMsg(const char* audit_msg, const char* item)
```
franku, line 146:
```suggestion
void UaContext::LogAuditEventInfoMsg(const char* fmt, ...)
```

franku: Would be nice to have an interface that has a format string as input parameter. 

See void UaContext::InfoMsg(const char* fmt, ...) as example.

core/src/dird/ua_run.cc
```diff
@@ -525,6 +525,15 @@ int DoRunCmd(UaContext* ua, const char* cmd)
 
     FreeJcr(jcr); /* release jcr */
 
+    /*
+     * For interactive runs we send a message to the audit log
+     */
+    if (jcr->impl->IgnoreLevelPoolOverides) {
```
franku, line 531:
```suggestion
    if (jcr->impl->IgnoreLevelPoolOverrides) {
```

franku: Although this typo is not yours, you may want to correct that.

core/src/dird/ua_audit.cc
fbergkemper: Suggestion implemented

core/src/dird/ua_run.cc
fbergkemper: corrected spelling error

core/src/dird/ua_restore.cc
fbergkemper: rebase accident corrected

PR #494, opened 2020-04-26
Added support for Manjaro Linux

apachler:
Also changed the lsb_release call for DISTVER to get the release number not the complete release description. I think this makes more sense for most distributions.

### DIFF
--- a/core/cmake/distname.sh
+++ b/core/cmake/distname.sh
@@ -54,7 +54,7 @@ else
       #
       if test "x$DISTNAME" != "x"
       then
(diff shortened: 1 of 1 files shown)


pstorz:
Hello,

thank you for your contribution.

However, chaniging the DISTVER string like this can have a lot of side effects.

As we are working on refactoring the whole OS information gathering, I would like to suggest we leve the DISTVER calculation in your PR like it is.

Thenk you very much for your contribution,

Philipp

arogge:
Created #513 as a merge of #494 and #495

PR #496, opened 2020-04-27
fix fadvise - issue 1227

gjelu:
As per request in https://bugs.bareos.org/view.php?id=1227

### DIFF
--- a/core/src/findlib/bfile.cc
+++ b/core/src/findlib/bfile.cc
@@ -1133,9 +1133,10 @@ int bopen(BareosWinFilePacket* bfd,
   bfd->win32DecompContext.liNextHeader = 0;
 
 #if defined(HAVE_POSIX_FADVISE) && defined(POSIX_FADV_WILLNEED)
-  if (bfd->fid != -1 && flags & O_RDONLY) {
+  /* If not RDWR or WRONLY must be Read Only */
+  if (bfd->fid != -1 && !(flags & (O_RDWR|O_WRONLY))) {
     int status = posix_fadvise(bfd->fid, 0, 0, POSIX_FADV_WILLNEED);
-    Dmsg2(400, "Did posix_fadvise on %s status=%d\n", fname, status);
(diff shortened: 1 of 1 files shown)


core/src/findlib/bfile.cc
```diff
@@ -1185,10 +1186,12 @@ int bclose(BareosWinFilePacket* bfd)
     bfd->cmd_plugin = false;
   } else {
 #if defined(HAVE_POSIX_FADVISE) && defined(POSIX_FADV_DONTNEED)
-    if (bfd->flags_ & O_RDONLY) {
+    /* If not RDWR or WRONLY must be Read Only */
+    if (!(bfd->m_flags & (O_RDWR|O_WRONLY))) {
       fdatasync(bfd->fid); /* sync the file */
```
franku, line 1191:
```suggestion
```

core/src/findlib/bfile.cc
```diff
@@ -1185,10 +1186,12 @@ int bclose(BareosWinFilePacket* bfd)
     bfd->cmd_plugin = false;
   } else {
 #if defined(HAVE_POSIX_FADVISE) && defined(POSIX_FADV_DONTNEED)
-    if (bfd->flags_ & O_RDONLY) {
+    /* If not RDWR or WRONLY must be Read Only */
+    if (!(bfd->m_flags & (O_RDWR|O_WRONLY))) {
```
arogge, line 1190:
```suggestion
    if (!(bfd->flags_ & (O_RDWR|O_WRONLY))) {
```

core/src/findlib/bfile.cc
arogge: We discussed this and we agree that calling `fdatasync()` is not a good idea in this case.

arogge:
Could you please update the commit message according to https://docs.bareos.org/DeveloperGuide/gitworkflow.html#commit-message-guideline so it at least contains `#1227: Fix fadvise bug`, so Mantis will pick up the change automatically?
Thank you!

PR #499, opened 2020-04-30
dird: fix status schedule(r) command behaviour

fbergkemper:
This commit fixes the behaviour of the status schedule(r) command.

- Take disabled resources into account when a job is given
- Take disabled resources into account when a client is given
- Take disabled resources into account when a schedule is given

### DIFF
--- a/core/src/dird/ua_status.cc
+++ b/core/src/dird/ua_status.cc
@@ -554,7 +554,7 @@ static void DoSchedulerStatus(UaContext* ua)
   int i;
   int max_date_len = 0;
   int days = DEFAULT_STATUS_SCHED_DAYS; /* Default days for preview */
-  bool schedulegiven = false;
+  bool schedule_given_by_cmdline_args = false;
   time_t time_to_check, now, start, stop;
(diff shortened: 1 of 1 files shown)


core/src/dird/ua_status.cc
```diff
@@ -625,7 +625,12 @@ static void DoSchedulerStatus(UaContext* ua)
       if (job->schedule &&
           bstrcmp(sched->resource_name_, job->schedule->resource_name_)) {
         if (cnt++ == 0) { ua->SendMsg("%s\n", sched->resource_name_); }
-        ua->SendMsg("                       %s\n", job->resource_name_);
+        if (job->enabled && (!job->client || job->client->enabled)) {
```
franku, line 628:
```suggestion
        if (!job->enabled || (job->client && !job->client->enabled)) {
```

franku: And swap if and else branch.

fbergkemper: suggested changes applied

PR #500, opened 2020-05-03
Fix login form problems

tikoflano:
Hides a the login form and shows a "loading" message while it does the Ajax call to get the versions. This avoids the problem that happens when you load the login form and click login too quickly (it logs in incorrectly) and also the visual effect of the dirctor and locale empty selects.

### DIFF
--- a/webui/module/Auth/view/auth/auth/login.phtml
+++ b/webui/module/Auth/view/auth/auth/login.phtml
@@ -37,6 +37,12 @@ $this->headTitle($title);
 <!-- Login form -->
 <div class="col-md-4">
 
+   <div class="progress">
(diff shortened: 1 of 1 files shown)


fbergkemper:
Hello Alvaro,

thanks for all the effort you put into this, but we would like to solve it in another way after we discussed your approach.

I found 2 major problems after digging into it. PR#505 is going to fix those, see https://github.com/bareos/bareos/pull/505 .

Especially the following commits in PR#505 should fix the problem you reported.

- https://github.com/bareos/bareos/pull/505/commits/2b897a64ed2a3983893f879ee064e986e6ae9d82
- https://github.com/bareos/bareos/pull/505/commits/e0c2a12054999a43908cef0450cdf767005a77e0

Can you maybe apply those 2 patches in your environment and give us some feedback if that fixes the issue for you as well?

Cheers,
Frank

tikoflano:
Yes, that works fine. Glad to help!

PR #505, opened 2020-05-06
improve frontend login dialog

fbergkemper:


webui/module/Auth/src/Auth/Controller/AuthController.php
```diff
@@ -253,7 +259,7 @@ public function getDirectorModel()
     *
     * @return object
     */
-   private function createNewLoginForm($form, $err_msg = null, $bsock = null)
+   private function createNewLoginForm($form, $multi_dird_env = null, $err_msg = null, $bsock = null)
```
franku, line 262:
```suggestion
   private function createNewLoginForm($form, $multi_dird_env, $err_msg = null, $bsock = null)
```

webui/module/Auth/view/auth/auth/login.phtml
```diff
@@ -58,7 +58,11 @@ $this->headTitle($title);
          echo '<div class="input-group">';
       }
       echo '<div class="input-group-addon"><span class="glyphicon glyphicon-log-in" aria-hidden="true"></span></div>';
-      echo $this->formSelect($form->get('director')->setAttribute('class','form-control selectpicker'));
+      if($multi_dird_env) {
```
franku, line 61: multi_dird_env: From which class is this variable?

fbergkemper: It is passed into the View from the AuthController, see line 217.

webui/module/Auth/src/Auth/Controller/AuthController.php
fbergkemper: It is passed to the View and checked there, I would like to leave it as it is.

PR #506, opened 2020-05-06
docs: add warning about job-intermixing

arogge:
The intermixing of jobs with the same client and fileset is considered counter-intuitive by some users.
This patch adds a warning to the documentation that clarifies the situation.
(User-Story 3403)


docs/manuals/source/Configuration/Director.rst
```diff
@@ -81,8 +81,16 @@ Only a single type (Backup, Restore, ...) can be specified for any job. If you w
...
 
-Multiple Storage daemons are not currently supported for Jobs, so if you do want to use multiple storage daemons, you will need to create a different Job and ensure that for each Job that the combination of Client and FileSet are unique. The Client and FileSet are what Bareos uses to restore a client, so if there are multiple Jobs with the same Client and FileSet or multiple Storage daemons that are used, the restore will not work. This problem can be resolved by defining multiple FileSet
-definitions (the names must be different, but the contents of the FileSets may be the same).
+Multiple Storage daemons are not currently supported for Jobs, if you do want to use multiple storage daemons, you will need to create a different Job and ensure that that the combination of Client and FileSet are unique.
+
+.. warning::
+
+  Bareos uses only :config:option:`dir/job/Client` and :config:option:`dir/job/Fileset` to determine which jobids belong together.
+  If job A and B have the same client and fileset defined, the resulting jobids will be intermixed:
```
franku, line 89:
```suggestion
  If job A and B have the same client and fileset defined, the resulting jobids will be intermixed as follows:
```

docs/manuals/source/Configuration/Director.rst
```diff
@@ -81,8 +81,16 @@ Only a single type (Backup, Restore, ...) can be specified for any job. If you w
...
+  When a job determines its predecessor to determine its required level and since-time, it will consider all jobs with the same client and fileset.
```
franku, line 90:
```suggestion
  A) When a job determines its predecessor to determine its required level and since-time, it will consider all jobs with the same client and fileset.
```

docs/manuals/source/Configuration/Director.rst
```diff
@@ -81,8 +81,16 @@ Only a single type (Backup, Restore, ...) can be specified for any job. If you w
...
+  Likewise when restoring a client you select the fileset and all jobs that used that fileset will be considered.
```
franku, line 91:
```suggestion
  B) When restoring a client you have to select the fileset and all jobs that used the regarding fileset will be considered.
```

docs/manuals/source/Configuration/Director.rst
```diff
@@ -81,8 +81,16 @@ Only a single type (Backup, Restore, ...) can be specified for any job. If you w
...
+  If you want separate backups, you have to duplicate your filesets with a different name and the same content.
```
franku, line 92:
```suggestion
  As a matter of fact, if you want separate backups, you have to duplicate your filesets with a different name and the same content.
```

docs/manuals/source/Configuration/Director.rst
arogge: I'm not sure this will be formatted in a way I like, but I'll try it.

docs/manuals/source/Configuration/Director.rst
```diff
@@ -81,8 +81,18 @@ Only a single type (Backup, Restore, ...) can be specified for any job. If you w
 
 Note, you define only a single Job to do the Full, Differential, and Incremental backups since the different backup levels are tied together by a unique Job name. Normally, you will have only one Job per Client, but if a client has a really huge number of files (more than several million), you might want to split it into to Jobs each with a different FileSet covering only part of the total files.
```
fbergkemper, line 82:
```suggestion
Note, you define only a single Job to do the Full, Differential, and Incremental backups since the different backup levels are tied together by a unique Job name. Normally, you will have only one Job per Client, but if a client has a really huge number of files (more than several million), you might want to split it into several Jobs each with a different FileSet covering only part of the total files.
```

docs/manuals/source/Configuration/Director.rst
```diff
@@ -81,8 +81,18 @@ Only a single type (Backup, Restore, ...) can be specified for any job. If you w
...
 
-Multiple Storage daemons are not currently supported for Jobs, so if you do want to use multiple storage daemons, you will need to create a different Job and ensure that for each Job that the combination of Client and FileSet are unique. The Client and FileSet are what Bareos uses to restore a client, so if there are multiple Jobs with the same Client and FileSet or multiple Storage daemons that are used, the restore will not work. This problem can be resolved by defining multiple FileSet
-definitions (the names must be different, but the contents of the FileSets may be the same).
+Multiple Storage daemons are not currently supported for Jobs, if you do want to use multiple storage daemons, you will need to create a different Job and ensure that that the combination of Client and FileSet are unique.
```
fbergkemper, line 84:
```suggestion
Multiple Storage daemons are not currently supported for Jobs, if you do want to use multiple storage daemons, you will need to create a different Job and ensure the combination of Client and FileSet is unique.
```

PR #507, opened 2020-05-07
docs: document Verify VolumeToCatalog limitation

arogge:
When running a Verify VolumeToCatalog job, the stored file content is not compared against the recorded checksum which allows data errors to remain undetected in specific cases.
This patch adds a description of that limitation.


docs/manuals/source/manually_added_config_directive_descriptions/dir-job-Level.rst.inc
```diff
@@ -112,7 +112,7 @@ Verify
 
       VolumeToCatalog jobs need a client to extract the metadata, but this client does not have to be the original client. We suggest to use the client on the backup server itself for maximum performance.
```
fbergkemper, line 113:
```suggestion
      VolumeToCatalog jobs require a client to extract the metadata, but this client does not have to be the original client. We suggest to use the client on the backup server itself for maximum performance.
```

PR #508, opened 2020-05-08
Added VirtualFull option to JobRun Form

tikoflano:
Added VirtualFull option for manual defined Jobs. In my case, sometimes you have a Full Job and some Incremental  ones and you just one to merge them into a new Virtual Full. From the UI it is really simple to do it if we add this option to the menu.

### DIFF
--- a/webui/module/Job/src/Job/Form/RunJobForm.php
+++ b/webui/module/Job/src/Job/Form/RunJobForm.php
@@ -415,6 +415,7 @@ class RunJobForm extends Form
       $selectData['Full'] = 'Full';
       $selectData['Differential'] = 'Differential';
       $selectData['Incremental'] = 'Incremental';
+      $selectData['VirtualFull'] = 'VirtualFull';
       return $selectData;
    }
 }


fbergkemper:
Thanks for the PR, Alvaro.

fbergkemper:
Looks good. Useful for many others, I think.

fbergkemper:
Ok, here is just an idea if you would like to extend this a bit in a future PR. It would be great to have an additional form element (next pool) displayed but just in case you switched to a VirtualFull. The displayed form element (input select) should have the default Next Pool directive of the selected pool preselected but it should be possible to override it too. I think you get the idea.

Going to merge this one now.

Cheers,
Frank

tikoflano:
Great! I'm working on the changes you proposes. They make sense to me.

fbergkemper:
Nice, looking forward to it.

By the way, if you like, join our bareos-devel mailing list which is always a good place to discuss Bareos development related stuff.

https://groups.google.com/forum/#!forum/bareos-devel/join

PR #509, opened 2020-05-08
Fixed JobDetails joblog messages order

tikoflano:
When you have multiple log messages with the same time the display order is weird. For example, lets say we have this joblogs in this order returned from bconsole:

- A 13:11
- B 13:11
- C 13:11
- D 13:12
- E 13:12

If we order them by desc time, we expect to have:

- E
- D
- C
- B
- A

But we are getting:

- D
- E
- A
- B
- C

By adding an Id to each message we can solve this problem.


fbergkemper:
Thanks for the PR, Alvaro.

PR #511, opened 2020-05-10
Show current bytes for running jobs in details page

tikoflano:
This will show the current bytes of the job when visiting the job details page for a running job


fbergkemper:
Thanks again for the PR, Alvaro.

Unfortunately, the calculation in your approach does not work having interleaved volume writing and concurrency in mind. If you use Bareos without Data Spooling, multiple concurrent backup jobs can be written to a Volume using interleaving.

You might want to have a look at https://github.com/bareos/bareos/blob/8a1e696653d58a9104343d6296b853eb5c708d52/webui/module/Job/src/Job/Model/JobModel.php#L299 to get an idea about how to retrieve information about running jobs.
The Dashboard Partial "Running Jobs" makes use of this, for example.

Best regards,
Frank

PR #510, opened 2020-05-10
Format joblogs to show direct link to queued jobs

tikoflano:
Add an special case for formatting the joblog message. It takes the messages in the form JobId=xxx and turns it to a link to that job. Usefull for those jobs that queue other jobs.

PR #519, opened 2020-05-15
Extra files to be ignored

tikoflano:
If I understood correctly the https://github.com/bareos/bareos/pull/510#issuecomment-628183672, this files can be ignored safely as they are generated on the build process.


tikoflano:
I realized that was commiting using the root user on my machine instead of my GitHub user. I rebased the commits but I'm not sure if this is the correct way of doing it.

Should I rebase the same way my other PR's ? Or should I leave them as they are?

fbergkemper:
> 
> 
> I realized that was commiting using the root user on my machine instead of my GitHub user. I rebased the commits but I'm not sure if this is the correct way of doing it.
> 
> Should I rebase the same way my other PR's ? Or should I leave them as they are?

In general please rebase, we like to know who did the commit.

Regarding this PR, this has been changed by Andreas meanwhile in a general cleanup, he did recently. Thus I'm going to close it, but thanks anyways.

PR #520, opened 2020-05-15
systemtests: setup_data: exit immediately, if it fails.

joergsteffens:
Before, it just returns an error code (1).
However, this code have never been checked,
making it hard to find the real problem of failing tests.

### DIFF
--- a/systemtests/scripts/functions
+++ b/systemtests/scripts/functions
@@ -115,7 +115,7 @@ setup_data()
   data_dir="${tmp}/data"
 
   [ -d "$data_dir" ] || mkdir "$data_dir"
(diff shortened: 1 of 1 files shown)


systemtests/scripts/functions
```diff
@@ -115,16 +115,16 @@ setup_data()
   data_dir="${tmp}/data"
 
   [ -d "$data_dir" ] || mkdir "$data_dir"
+  echo "${data_dir}" >"${tmp}/file-list"
```
franku, line 118:
```suggestion
  echo "$data_dir" >"$tmp/file-list"
```

PR #526, opened 2020-05-18
Show client description

tikoflano:
This PR allows to show the Description field in the definition of each Client as a new column on both clients and client views.

This can become handy in some cases, like the one it helps me now. I have each Client name defined as its IP, for simplicity. Now, some clients are special cases and I would like to add in the description some extra information that helps me identify it (could also be some info that coudl help the administrator to organize the clients in groups).


fbergkemper:
Thanks Alvaro for another good suggestion, but this needs an internal discussion first, if we would like to implement it this way. There is probably a better approach by extending the list clients command result output by another key/value pair (description) on api 2 which is currently not given. More feedback will follow soon.

fbergkemper:
We decided to implement API 2 (JSON) output on the show command to avoid such preg_match solutions in the frontend. Thus I would like to reject this PR. If you are interested in reimplementing the proposed feature in this PR when the change is available let me know and I'll keep you updated when it is available.

tikoflano:
Seems reasonable. Let me know when that is available on the API.

PR #530, opened 2020-05-20
Dev/maik/master/local fileset weird files

aussendorf:
Encodes filenames (string_escape) before sending to director, this way weird filenames will be restored correctly. 

Also renamed exisiting systemtest python-fd-plugin-local-fileset-test to python-fd-plugin-local-fileset-rop-test and created a new test including weird-files.

Also added class variable and new method to LocalFilest to ease file-handling for derived classes, e.g. SplitJobs plugin, which is derived from this clas.


franku:
Would change the first commit message to a smaller headline: 

**python-plugins: improve statp handling in LocalFileset plugin** 

- and make own test for localfilest with rop
- ...

PR #543, opened 2020-06-09
dev/franku/master/remove warnings

franku:


core/src/stored/ndmp_tape.cc
```diff
@@ -1364,12 +1366,12 @@ extern "C" void* ndmp_thread_server(void* arg)
...
-    pfds[nfds].fd = fd_ptr->fd;
-    pfds[nfds].events |= POLL_IN;
-    nfds++;
-  }
+  std::for_each(sockfds.begin(), sockfds.end(),
+                [&pfds, &nfds](const s_sockfd* fd_ptr) {
+                  pfds[nfds].fd = fd_ptr->fd;
+                  pfds[nfds].events |= POLL_IN;
+                  nfds++;
+                });
```
arogge, line 1374: What's the point of using `std::for_each` with a lambda instead of a ranged-for?

PR #546, opened 2020-06-25
Fix #1257: Prune volume command may crash the director

arogge:
This PR adds a test-case to the prune-volume systemtest that checks the broken behaviour. It also contains a fix for the issue.


arogge:
Nevermind, I was the one who didn't notice it during review.

PR #549, opened 2020-06-26
dev/franku/master/droplet s3 test

franku:
Uses a dedicated centos7 container.


franku:
Will continue with a different approach on another PR.

PR #554, opened 2020-07-10
allow backup/restore of cephfs mounts

arogge:
The filed did not pick up xattr on filesystems when st_dev == 0. This PR now also considers such filesystems. Additionally the virtual xattrs that cephfs provides were added to the blacklist for Linux.

### DIFF
--- a/core/src/findlib/xattr.h
+++ b/core/src/findlib/xattr.h
@@ -86,10 +86,11 @@ struct xattr_parse_data_t {
 /*
  * Internal tracking data.
  */
-struct xattr_data_t {
+struct XattrData {
   POOLMEM* last_fname;
-  uint32_t flags; /* See BXATTR_FLAG_* */
-  uint32_t current_dev;
+  uint32_t flags{0}; /* See BXATTR_FLAG_* */
+  uint32_t current_dev{0};
+  bool first_dev{true};
   union {
     struct xattr_build_data_t* build;
     struct xattr_parse_data_t* parse;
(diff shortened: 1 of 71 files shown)


arogge:
This patch works on linux, but doesn't compile on Solaris, FreeBSD and Windows. Will need (probably minor) work.

core/src/findlib/xattr.h
```diff
@@ -86,10 +86,11 @@ struct xattr_parse_data_t {
 /*
  * Internal tracking data.
  */
-struct xattr_data_t {
+struct XattrData {
   POOLMEM* last_fname;
-  uint32_t flags; /* See BXATTR_FLAG_* */
-  uint32_t current_dev;
+  uint32_t flags{0}; /* See BXATTR_FLAG_* */
```
franku, line 91: Where to find BXATTR_FLAG_ ? Could you make flags a enum class?

systemtests/tests/xattr/testrunner
```diff
@@ -0,0 +1,77 @@
...
+BackupDirectory="${tmp}/data"
+BackupFile="${tmp}/file-with-xattr"
+RestoreFile="${tmp}/bareos-restores/${BackupFile}"
+
+xattr_should="${tmp}/xattr_should.txt"
+xattr_is="${tmp}/xattr_is.txt"
+
+mkdir -p "${BackupDirectory}"
+touch "${BackupFile}"
+setfattr --name=user.bareos.attr1 --value=value1 "${BackupFile}"
```
franku, line 31:
```suggestion
setfattr --name=user.bareos.attr1 --value=value1 "${BackupFile}" || exit_error "Could not set attribute"
```

systemtests/tests/xattr/testrunner
```diff
@@ -0,0 +1,77 @@
+#!/bin/sh
+#
+# Run a simple backup
+#   then restore it.
+#
```
franku, line 5:
```suggestion
#
set -e
```

systemtests/tests/xattr/testrunner
franku: Could add something like this exit_error function that makes it easier to see an error during script execution or add set -e.

franku (approved):
Really good work, however I have two minor suggestions. Pls see the comments.

systemtests/tests/xattr/testrunner
arogge: The code doesn't quit in that case intentionally.
Point being is that if you run the systemtests on a filesystem w/o xattr-support, the setfattr command will fail and you will end up with a file with no xattrs. 
While it doesn't check the xattrs really work, it at least doesn't fail.

The right thing is probably to disable the test in that case. However, I don't see an easy way to do that properly.

franku: "At least it doesn't fail" would lead to a false positive test. Couldn't you use a cmake command to verify if the setfattr command succeeds in general?

core/src/findlib/xattr.h
arogge: I thought about it, but refrained from doing it. I was afraid of the side-effects in that `#ifdef` hell. However, I think it can be done without too much work, so I'll look into it.

systemtests/tests/xattr/testrunner
franku: imho it should be the same path as where the test will be running.

arogge: That would be best, but that path is only created after we decided whether or not the test should be enabled.

PR #555, opened 2020-07-14
build libdroplet as git submodule of bareos

franku:


core/src/stored/backends/droplet_device.cc
```diff
@@ -361,14 +361,26 @@ bool droplet_device::CheckRemote()
...
+  std::string hostaddr{h != nullptr ? h : "???"};
 
-  return true;
+  switch (status) {
+    case DPL_SUCCESS:
+      Dmsg1(100, "Host is accessible: %s\n", hostaddr.c_str());
+      return true;
+    case DPL_ENOENT:
+      Dmsg2(100,
+            "Host is accessible: %s (%s), but should be configured to accept "
```
pstorz, line 375: Is this the only reason that causes this return value? If not, then the error message should say something like "probably the host should be configured...."

PR #556, opened 2020-07-15
build from source: fix "make install" and add "make uninstall"

pstorz:
When building bareos from source, "make install" still had some problems which are solved now.
Also, "make uninstall" was added which removes the files that are listed in the cmake install "manifest" file.


core/src/console/CMakeLists.txt
```diff
@@ -45,10 +45,30 @@ endif()
...
+  TARGETS bconsole
+  DESTINATION "${sbindir}"
+  COMPONENT bconsole
+)
+
+install(
+  CODE "
+  message(STATUS \"DESTDIR is \$ENV{DESTDIR}\")
+  set(DESTDIR \$ENV{DESTDIR})
+  #if(EXISTS \$ENV{DESTDIR}${configtemplatedir} AND NOT EXISTS \$ENV{DESTDIR}${configtemplatedir}/bconsole.conf)
```
franku, line 63:
```suggestion

```

franku, line 63:
```suggestion
```

franku:
`make install` rises Bareos to a higher level of developer awareness. 
Good work!

PR #561, opened 2020-07-21
fix droplet timeout

franku:
For review with Stephan

### DIFF
--- a/core/src/stored/backends/droplet_device.cc
+++ b/core/src/stored/backends/droplet_device.cc
@@ -431,111 +431,127 @@ bool droplet_device::FlushRemoteChunk(chunk_io_request* request)
   /*
    * Set that we are uploading the chunk.
    */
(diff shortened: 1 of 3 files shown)


core/src/stored/backends/droplet_device.cc
```diff
@@ -431,111 +431,127 @@ bool droplet_device::FlushRemoteChunk(chunk_io_request* request)
...
-    default:
-      Mmsg2(errmsg, _("Failed to flush %s using dpl_fput(): ERR=%s.\n"),
-            chunk_name.c_str(), dpl_status_str(status));
-      dev_errno = DropletErrnoToSystemErrno(status);
-      goto bail_out;
-  }
+    switch (status) {
+      case DPL_SUCCESS:
+        success = true;
+        break;
```
franku, line 535:
```suggestion
        goto bail_out;
        break;
```

PR #564, opened 2020-07-30
Add deprecation notifications for catalogs using mysql/sqlite3

arogge:
This PR adds configuration warning for every catalog that uses DB driver mysql oder sqlite3.
To implement this a simple validation call-back was added to the director's configparser and every resource needing validation can just implement that.


arogge:
I just realized there are two cleanup commits in the branch. Hope that doesn't hurt.

PR #565, opened 2020-08-05
Fix hearbeat thread shutdown race condition

franku:
This is a quick fix for the heartbeat shutdown where it could happen that the thread is already down but the stopping thread is still waiting for it to start. However this works the complete heartbeat code should be refactored again. 

The fix uses a new variable that is only set if the thread has been started once.  

Otherwise I left the code as much as before as I could.


franku:
Jenkins build #5 succeeded completely on commit 8980ca5f8.

PR #569, opened 2020-08-11
fix build problems on new distributions

pstorz:
Fixes problems on Fedora 32 , SLE15sp1, openSUSE_Leap_15.1:
* bareos.spec: dependency on libtermcap-devel which does not exist anymore,
* ceph: detection of ceph_statx structure

### DIFF
--- a/core/cmake/BareosCheckIncludes.cmake
+++ b/core/cmake/BareosCheckIncludes.cmake
@@ -1,6 +1,6 @@
 #   BAREOS® - Backup Archiving REcovery Open Sourced
 #
-#   Copyright (C) 2017-2019 Bareos GmbH & Co. KG
+#   Copyright (C) 2017-2020 Bareos GmbH & Co. KG
 #
 #   This program is Free Software; you can redistribute it and/or
 #   modify it under the terms of version three of the GNU Affero General Public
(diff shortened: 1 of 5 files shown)


core/cmake/BareosCheckIncludes.cmake
```diff
@@ -60,7 +60,14 @@ check_include_files(sys/ea.h HAVE_SYS_EA_H)
 check_include_files(sys/proplist.h HAVE_SYS_PROPLIST_H)
 check_include_files(sys/xattr.h HAVE_SYS_XATTR_H)
 
-check_include_files("sys/stat.h;cephfs/ceph_statx.h" HAVE_CEPHFS_CEPH_STATX_H)
+
+include(CheckSymbolExists)
+include(CMakePushCheckState)
+cmake_push_check_state()
+set(CMAKE_REQUIRED_LIBRARIES cephfs)
+check_symbol_exists(ceph_statx "sys/stat.h;cephfs/libcephfs.h" HAVE_CEPHFS_CEPH_STATX_H)
```
arogge, line 68: The name `HAVE_CEPHFS_STATX_H` is now misleading. Maybe we could rename it `HAVE_CEPHFS_STATX`?

PR #571, opened 2020-08-12
MSVDI: use a configurable timeout when requesting the config

franku:
User Story 4151

### DIFF
--- a/core/src/win32/plugins/filed/mssqlvdi-fd.cc
+++ b/core/src/win32/plugins/filed/mssqlvdi-fd.cc
@@ -73,6 +73,7 @@ static const int debuglevel = 150;
   "  stopbeforemark=<log sequence number specification>:\n" \
   "  stopatmark=<log sequence number specification>:\n"     \
   "  stopat=<timestamp>\n"                                  \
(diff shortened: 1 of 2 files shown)


core/src/win32/plugins/filed/mssqlvdi-fd.cc
```diff
@@ -1373,18 +1406,54 @@ static inline bool SetupVdiDevice(bpContext* ctx, struct io_pkt* io)
...
+        break;
+    }
+
+    std::array<char, 300> error_msg;
+    const char* fmt =
+        "mssqlvdi-fd: IClientVirtualDeviceSet2::GetConfiguration "
+        "%s: %0#x (%s)\n";
+
+    if (success) {
+      sprintf(error_msg.data(), fmt, "successful",
```
arogge, line 1445: would the compiler warn if the destination buffer is too short? The possible values are all known at compile-time, so there should be a warning.
I'd consider using `snprintf()` if the compiler doesn't warn, as it makes the code more robust to changes.

franku:
Build succeeded with #11.

PR #572, opened 2020-08-18
add python3 support for fd, sd and dir plugins

pstorz:


franku (approved):
Huge amount of work! Great!

PR #573, opened 2020-08-19
enable testing of deprecated databases (MySQL and SQLite)

pstorz:
The ability to create deprecated catalog databases was removed before. Unfortunately, we still need to be able to create those databases for our own tests.

This PR enables the creation of deprecated databases for only when running in our test environment. Otherwise, the creation will be denied like before.

### DIFF
--- a/core/src/cats/create_bareos_database.in
+++ b/core/src/cats/create_bareos_database.in
@@ -81,12 +81,25 @@ fi
 
 case ${db_type} in
    sqlite3)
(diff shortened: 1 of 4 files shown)


arogge (approved):
On one hand I'm not happy with BAREOS_TEST_RUNNING, but then again it makes absolutely sure that you should not set this value when installing.
Good idea overall. Probably best / simplest solution to work around our issue.

PR #574, opened 2020-08-20
Extend bconsole show command to support JSON output format (API 2)

joergsteffens:


core/src/console/console_conf.cc
```diff
@@ -44,7 +44,7 @@ static bool SaveResource(int type, ResourceItem* items, int pass);
 static void FreeResource(BareosResource* sres, int type);
 static void DumpResource(int type,
                          BareosResource* reshdr,
-                         void sendit(void* sock, const char* fmt, ...),
+                         bool sendit(void* sock, const char* fmt, ...),
```
franku, line 47: PrintMessage is the only function called by this handler and this returns true, always.

core/src/dird/dird.cc
```diff
@@ -81,7 +81,7 @@ static JobControlRecord* PrepareJobToRun(const char* job_name);
 
 /* Exported subroutines */
 extern bool ParseDirConfig(const char* configfile, int exit_code);
-extern void PrintMessage(void* sock, const char* fmt, ...);
+extern bool PrintMessage(void* sock, const char* fmt, ...);
```
franku, line 84: Always returns true.

core/src/lib/parse_conf.cc
```diff
@@ -73,13 +73,15 @@
...
-void PrintMessage(void* sock, const char* fmt, ...)
+bool PrintMessage(void* sock, const char* fmt, ...)
 {
   va_list arg_ptr;
 
   va_start(arg_ptr, fmt);
   vfprintf(stdout, fmt, arg_ptr);
   va_end(arg_ptr);
+
+  return true;
```
franku, line 84: This is always true. As a matter of fact does not change code behaviour anywhere. 

You could just easily revert all changes from bool to void and remove the part of the false code branch above because it is dead code. (Please show me an other part in case I was wrong.)

core/src/lib/output_formatter.cc
```diff
@@ -878,7 +1056,7 @@ void OutputFormatter::JsonFinalizeResult(bool result)
     /*
      * send json string, on failure, send json error message
      */
-    if (!send_func(send_ctx, string)) {
+    if (!send_func(send_ctx, "%s", string)) {
```
franku, line 1059: This will never be false.

core/src/lib/output_formatter_resource.cc
```diff
@@ -0,0 +1,342 @@
...
+  }
+  lineformat.bsprintf(format.c_str(), key, value);
+  send_->ArrayItem(item, lineformat.c_str(), false);
+}
+
+
+void OutputFormatterResource::KeyMultipleStringsOnePerLine(
+    const char* key,
+    alist* list,
+    const char* (*GetValue)(void* item),
```
franku, line 257:
```suggestion
    std::function<const char* (void* item)> GetValue
```

core/src/lib/output_formatter_resource.cc
```diff
@@ -0,0 +1,342 @@
...
+      KeyMultipleStringsOnePerLineAddItem(key, GetValue(item), as_comment,
+                                          quoted_strings);
+    }
+    send_->ArrayEnd(key);
+  }
+}
+
+
+void OutputFormatterResource::KeyMultipleStringsOnePerLine(const char* key,
+                                                           alist& list,
```
franku, line 284:
```suggestion
                                                           const alist& list,
```

core/src/lib/output_formatter_resource.h
```diff
@@ -0,0 +1,148 @@
...
+ */
+
+#ifndef BAREOS_LIB_OUTPUT_FORMATTER_RESOURCE_H_
+#define BAREOS_LIB_OUTPUT_FORMATTER_RESOURCE_H_
+
+#include "lib/alist.h"
+#define NEED_JANSSON_NAMESPACE 1
+#include "lib/output_formatter.h"
+
+class PoolMem;
```
franku, line 33: Could you eliminate PoolMem somehow?

core/src/lib/output_formatter_resource.h
```diff
@@ -0,0 +1,148 @@
...
+                                   bool quoted_strings = true);
+
+  void KeyMultipleStringsInOneLine(const char* key,
+                                   alist* list,
+                                   const char* (*GetValue)(void* item),
+                                   bool as_comment = false,
+                                   bool quoted_strings = true);
+
+  void KeyMultipleStringsOnePerLine(const char* key,
+                                    alist& list,
```
franku, line 120: Why alist as pointer here and alist* otherwise? 
Could you try to eliminate alist at all?

core/src/lib/output_formatter_resource.h
```diff
@@ -0,0 +1,148 @@
...
+                                    bool quoted_strings = true);
+
+  void KeyMultipleStringsOnePerLine(const char* key,
+                                    alist* list,
+                                    const char* (*GetValue)(void* item),
+                                    bool as_comment = false,
+                                    bool quoted_strings = true);
+
+  void KeyMultipleStringsOnePerLine(const char* key,
+                                    const std::vector<std::string>&,
```
franku, line 136: This is std::vector.. good!

core/src/lib/output_formatter_resource.h
```diff
@@ -0,0 +1,148 @@
...
+                                    bool as_comment = false,
+                                    bool quoted_strings = true);
+
+  void ArrayStart(const char* key,
+                  bool as_comment = false,
+                  std::string baseformat = "%s {\n");
```
franku, line 142: baseformat ist never used as a different parameter than the default. You could use a constexpr member variable in the class and eliminate the unused parameter.

core/src/lib/output_formatter_resource.h
```diff
@@ -0,0 +1,148 @@
...
+  void ArrayEnd(const char* key,
+                bool as_comment = false,
+                std::string baseformat = "}\n");
```
franku, line 145: Same applies to baseformat of ArrayEnd. Could Eliminate this function parameter completely.

core/src/lib/util.cc
```diff
@@ -85,6 +85,15 @@ void EscapeString(PoolMem& snew, const char* old, int len)
   *n = 0;
 }
 
+
+std::string EscapeString(const char* old)
+{
+  PoolMem snew;
+  EscapeString(snew, old, strlen(old));
+  return std::string(snew.c_str());
```
franku, line 93: This returns a copy of the local string and thats ok. 

However, I'd prefer to avoid PoolMem and use a std::vector<char> instead. In turn this needs a slight refactoring of the EscapeString(PoolMem,..) function but would eliminate another occurrence of PoolMem.

core/src/lib/util.cc
```diff
@@ -1137,3 +1146,19 @@ std::string getenv_std_string(std::string env_var)
...
+  PoolMem* pm = (PoolMem*)pm_string;
+
+  va_list arg_ptr;
+  va_start(arg_ptr, fmt);
+  additionalstring.Bvsprintf(fmt, arg_ptr);
+  va_end(arg_ptr);
+
+  pm->strcat(additionalstring);
+
+  return true;
```
franku, line 1163: Another one that returns always true.

core/src/lib/util.cc
```diff
@@ -1137,3 +1146,19 @@ std::string getenv_std_string(std::string env_var)
   const char* v = (std::getenv(env_var.c_str()));
   return v ? std::string(v) : std::string();
 }
+
+
+bool pm_append(void* pm_string, const char* fmt, ...)
```
franku, line 1151: I know, this is part of old legace code. But this ugly void* casts should be eliminated in the future. Otherwise it is always as difficult to read.

core/src/lib/util.cc
```diff
@@ -1137,3 +1146,19 @@ std::string getenv_std_string(std::string env_var)
...
+{
+  PoolMem additionalstring;
+  PoolMem* pm = (PoolMem*)pm_string;
```
franku, line 1154:
```suggestion
  PoolMem* pm = static_cast<PoolMem*>(pm_string);
```

core/src/stored/autochanger_resource.cc
```diff
@@ -48,7 +48,10 @@ AutochangerResource& AutochangerResource::operator=(
   return *this;
 }
 
-bool AutochangerResource::PrintConfigToBuffer(PoolMem& buf)
+bool AutochangerResource::PrintConfig(OutputFormatterResource& send,
+                                      const ConfigurationParser& /* unused */,
```
franku, line 52:
```suggestion
                                      const ConfigurationParser&,
```

core/src/stored/autochanger_resource.h
```diff
@@ -35,7 +35,11 @@ class AutochangerResource : public BareosResource {
   AutochangerResource();
   virtual ~AutochangerResource() = default;
   AutochangerResource& operator=(const AutochangerResource& rhs);
-  bool PrintConfigToBuffer(PoolMem& buf);
+  bool PrintConfig(OutputFormatterResource& send,
+                   const ConfigurationParser& /* unused */,
```
franku, line 39:
```suggestion
                   const ConfigurationParser&,
```

core/src/stored/device_resource.cc
```diff
@@ -230,26 +230,29 @@ DeviceResource& DeviceResource::operator=(const DeviceResource& rhs)
   return *this;
 }
 
-bool DeviceResource::PrintConfig(PoolMem& buf,
+
+bool DeviceResource::PrintConfig(OutputFormatterResource& send,
                                  const ConfigurationParser& /* unused */,
```
franku, line 235:
```suggestion
                                 const ConfigurationParser&,
```

franku (commented):
Started with review mostly on semantic things with some suggestions and comments. 
In my opinion the major work for switching void sendit into bool sendit could habe been avoided because all function-callbacks that are given to this argument always return true.

core/src/lib/output_formatter.cc
joergsteffens: It can fail, if there is a network problem.

core/src/lib/output_formatter_resource.h
joergsteffens: Unfortenatly, it is sometimes "`%s {\n`" (RunScripts, FileSets) and sometimes "`%s = {\n`" (equal sign, CFG_TYPE_ADDRESSES).

core/src/lib/output_formatter_resource.h
joergsteffens: It is set to `""` in some cases.

core/src/lib/util.cc
joergsteffens: Well, as the code is right now, this function must be defined in the same parameter as PrintMessage(void* sock, const char* fmt, ...), ...

systemtests/tests/python-bareos/python-bareos-unittest.py
```diff
@@ -1332,6 +1340,52 @@ def test_admin_runscript_client(self):
...
+
+        self.configure_add(
+            director,
+            resourcesname,
+            newclient,
+            u"client={} password={} address=127.0.0.1".format(newclient, newpassword),
+        )
+
+        result = director.call("show all")
+        result = director.call("show all verbose")
```
franku, line 1383: += ??

systemtests/tests/config-dump/testrunner
```diff
@@ -0,0 +1,164 @@
+#!/bin/bash
+#
```
franku, line 2:
```suggestion
#
-set e
-set -u

```

systemtests/tests/config-dump/testrunner
```diff
@@ -0,0 +1,164 @@
...
+  exit 1
+fi
+
+# Don't compare directly,
+# as there are known differences.
+#if ! diff "${conf}/bareos-dir-full.conf" "$tmp/bareos-dir-show-full1.conf"; then
+#  set_error "Configuration dumped (full1) differs. This should not happen."
+#  exit 1
+#fi
+if ! diff "${conf}/bareos-dir-full.conf" "$tmp/bareos-dir-show-verbose-full1.conf" > "$tmp/bareos-dir-full_bareos-dir-show-verbose-full1.diff"; then
```
franku, line 143:
```suggestion
if ! diff -q "${conf}/bareos-dir-full.conf" "$tmp/bareos-dir-show-verbose-full1.conf" > "$tmp/bareos-dir-full_bareos-dir-show-verbose-full1.diff"; then
```

franku (commented):
Some comments more.

systemtests/tests/config-dump/testrunner
joergsteffens: The output is redirected to a file.

systemtests/tests/python-bareos/python-bareos-unittest.py
joergsteffens: Covered with last commits.

systemtests/tests/config-dump/testrunner
joergsteffens: Done.

PR #575, opened 2020-08-20
Use configurable OpenSSL protocol parameter

franku:
- add SSL_CONF_cmd for Protocol
- add config parameter to tls common
- move creation of contexts to TlsOpenSslPrivate
- cleanup constructor of TlsResource
- add changed json files for documentation

### DIFF
--- a/core/src/lib/tls_openssl.cc
+++ b/core/src/lib/tls_openssl.cc
@@ -2,7 +2,7 @@
    BAREOS® - Backup Archiving REcovery Open Sourced
 
    Copyright (C) 2005-2010 Free Software Foundation Europe e.V.
-   Copyright (C) 2014-2018 Bareos GmbH & Co. KG
+   Copyright (C) 2014-2020 Bareos GmbH & Co. KG
 
    This program is Free Software; you can redistribute it and/or
    modify it under the terms of version three of the GNU Affero General Public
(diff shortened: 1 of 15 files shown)


core/src/lib/tls_openssl.h
```diff
@@ -50,6 +50,7 @@ class TlsOpenSsl : public Tls {
 
   std::string TlsCipherGetName() const override;
   void SetCipherList(const std::string& cipherlist) override;
+  void SetProtocol(std::string protocol) override;
```
arogge, line 53: every other signature in this domain is `const std::string&`, is there any reason not to keep it consistent?

core/src/lib/tls_openssl.cc
```diff
@@ -47,31 +47,9 @@
...
-  if (!d_->openssl_ctx_) {
-    OpensslPostErrors(M_FATAL, _("Error initializing SSL context"));
-    return;
-  }
-
-  SSL_CTX_set_options(d_->openssl_ctx_, SSL_OP_ALL);
-
-  SSL_CTX_set_options(d_->openssl_ctx_, SSL_OP_NO_SSLv2 | SSL_OP_NO_SSLv3);
-}
+TlsOpenSsl::TlsOpenSsl() : d_(new TlsOpenSslPrivate) {}
```
arogge, line 50: :+1: for using unique_ptr pimpl

core/src/lib/tls_openssl_private.cc
```diff
@@ -95,6 +125,29 @@ bool TlsOpenSslPrivate::init()
     return false;
   }
 
+  if (!protocol_.empty()) {
+    SSL_CONF_CTX_set_flags(openssl_conf_ctx_,
+                           SSL_CONF_FLAG_FILE | SSL_CONF_FLAG_SHOW_ERRORS |
+                               SSL_CONF_FLAG_CLIENT | SSL_CONF_FLAG_SERVER);
+
+    bool err =
+        SSL_CONF_cmd(openssl_conf_ctx_, "Protocol", protocol_.c_str()) != 2;
```
arogge, line 134: I still believe we're missing `SSL_CONF_finish()`. I'll try to find out why it didn't work for you and keep you updated.

core/src/lib/tls_openssl_private.cc
```diff
@@ -556,3 +609,9 @@ void TlsOpenSsl::SetCipherList(const std::string& cipherlist)
   Dmsg1(100, "Set cipherlist:\t<%s>\n", cipherlist.c_str());
   d_->cipherlist_ = cipherlist;
 }
+
+void TlsOpenSsl::SetProtocol(std::string protocol)
+{
+  Dmsg1(100, "Set protocol:\t<%s>\n", protocol.c_str());
```
arogge, line 615: do we really need that debugging message here? The value isn't "set" here, it is just pushed into the data structure.

arogge (commented):
Looks great. I'd like to look into that linking issue with `SSL_CONF_finish()`.

core/src/lib/tls_openssl_private.cc
```diff
@@ -55,18 +56,47 @@ const std::string TlsOpenSslPrivate::tls_default_ciphers_{
...
     , tcp_file_descriptor_(0)
     , pem_callback_(nullptr)
     , pem_userdata_(nullptr)
     , verify_peer_(false)
 {
   Dmsg0(100, "Construct TlsOpenSslPrivate\n");
+
+  /* the SSL_CTX object is the factory that creates
+   * openssl objects, so initialize this first */
+
```
franku, line 69:
```suggestion
#if (OPENSSL_VERSION_NUMBER < 0x10002000L)
  #error "OPENSSL VERSION < 1.0.2 not supported"
#endif

```

franku:
Waits only for RHEL7 repo test.

arogge:
RHEL7 test succeeded.

PR #576, opened 2020-08-21
Fix no "Available versions" in restore dialog for Filenames with spaces

MSP-Stev:
https://bugs.bareos.org/view.php?id=1265

### DIFF
--- a/webui/module/Restore/src/Restore/Model/RestoreModel.php
+++ b/webui/module/Restore/src/Restore/Model/RestoreModel.php
@@ -155,7 +155,7 @@ class RestoreModel
     */
    public function getFileVersions(&$bsock=null, $clientname=null, $pathid=null, $filename=null) {
       if(isset($bsock)) {
-         $cmd = '.bvfs_versions jobid=0 client='.$clientname.' pathid='.$pathid.' fname="'.$filename.'"';
+         $cmd = '.bvfs_versions jobid=0 client='.$clientname.' pathid='.$pathid.' fname="'.addslashes($filename).'"';
          $result = $bsock->send_command($cmd, 2);
          $versions = \Zend\Json\Json::decode($result, \Zend\Json\Json::TYPE_ARRAY);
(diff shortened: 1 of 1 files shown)


arogge:
Would you mind testing this with a file that has a double-quotes in the name? It looks like that might be broken, too.

webui/module/Restore/src/Restore/Model/RestoreModel.php
```diff
@@ -155,7 +155,7 @@ public function getFiles(&$bsock=null, $jobid=null, $pathid=null) {
     */
    public function getFileVersions(&$bsock=null, $clientname=null, $pathid=null, $filename=null) {
       if(isset($bsock)) {
-         $cmd = '.bvfs_versions jobid=0 client='.$clientname.' pathid='.$pathid.' fname='.$filename;
+         $cmd = '.bvfs_versions jobid=0 client='.$clientname.' pathid='.$pathid.' fname="'.$filename.'"';
```
fbergkemper, line 158:
```suggestion
         $cmd = '.bvfs_versions jobid=0 client='.$clientname.' pathid='.$pathid.' fname="'.addslashes($filename).'"';
```
We should add backslashes before characters that need to be escaped in the fname string, too.

fbergkemper (changes requested):
Thanks for the contribution. Can you please apply the suggested change?

fbergkemper:
Suggestion applied by an additional commit.

fbergkemper:
MSP-Stev, if you would like to get mentioned in the Authors file with your real name, please let us know by a comment here.

MSP-Stev:
Sorry for the late response. i will test your suggestion on our system. my name is Stev Dubau :)

fbergkemper:
No problem, Stev. Going to add you to the Authors file, thanks again for the contribution.

MSP-Stev:
you are welcome.

PR #578, opened 2020-08-25
Allow multiline quoted strings

franku:
Allow strings this: 

{
...
Plugin = "python"                                                              
      ":module_path=/python-modules"
      ":module_name=bareos-fd-local-fileset"                                       
      ":filename=/tmp/file-list"
...
}

### DIFF
--- a/core/src/lib/lex.cc
+++ b/core/src/lib/lex.cc
@@ -3,7 +3,7 @@
 
    Copyright (C) 2000-2012 Free Software Foundation Europe e.V.
    Copyright (C) 2011-2012 Planets Communications B.V.
-   Copyright (C) 2013-2016 Bareos GmbH & Co. KG
+   Copyright (C) 2013-2020 Bareos GmbH & Co. KG
 
    This program is Free Software; you can redistribute it and/or
    modify it under the terms of version three of the GNU Affero General Public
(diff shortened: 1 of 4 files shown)


core/src/lib/lex.cc
```diff
@@ -549,6 +552,53 @@ static uint64_t scan_pint64(LEX* lf, char* str)
   return val;
 }
 
+class M {
```
arogge, line 555: You might be able to find a better name for that class.

core/src/lib/lex.cc
```diff
@@ -549,6 +552,53 @@ static uint64_t scan_pint64(LEX* lf, char* str)
...
+{
+  int i = lf->col_no;
+  while (lf->line[i]) {
+    if (lf->line[i] == '"') {
+      return true;
+    } else if (lf->line[i] != ' ') {
+      return false;
+    }
+    ++i;
+  };
```
arogge, line 598: see above

core/src/lib/lex.cc
```diff
@@ -549,6 +552,53 @@ static uint64_t scan_pint64(LEX* lf, char* str)
...
+  if (bfgets(m.pm, lf->fd) != NULL) {
+    int i = 0;
+    while (m.pm[i]) {
+      if (m.pm[i] == '"') {
+        return true;
+      } else if (m.pm[i] != ' ') {
+        return false;
+      }
+      ++i;
+    };
```
arogge, line 583: I'd compare to '\0' to clarify that this is an end-of-string check.
```suggestion
    for (int i = 0; m.pm[i] != '\0'; ++i) {
      if (m.pm[i] == ' ') { continue; } /* skip spaces */
      return (m.pm[i] == '"'); /* is the first non-space a double quote? */
    }```

arogge (commented):
Really nice work. Who would have thought our lexer could learn that?
Great application of RAII btw.

franku:
Build succeeded with #8

PR #579, opened 2020-08-26
Changes to Bareos 17.2 to build with the current Continous Integration Suite

arogge:
This PR applies some changes to bareos-17.2 so we can build it with the current build-images again.


pstorz:
Looks good.

PR #580, opened 2020-08-27
 Fix hearbeat thread shutdown race condition

franku:
Backport from PR#565


franku:
Build succeeded with #2

PR #581, opened 2020-08-28
Display archived jobs on "list backups"

arogge:
This PR changes `list backups` and `llist backups` to display archive jobs, too.
Also a systemtest that checks the behaviour of `list backups`/`llist backups` is added.

### DIFF
--- a/core/src/dird/ua_output.cc
+++ b/core/src/dird/ua_output.cc
@@ -1224,7 +1224,7 @@ static bool ParseListBackupsCmd(UaContext* ua,
     return false;
   }
 
-  selection.bsprintf("AND Job.Type='B' AND Client.Name='%s' ",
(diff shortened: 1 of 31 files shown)


core/src/dird/ua_output.cc
```diff
@@ -1224,7 +1224,7 @@ static bool ParseListBackupsCmd(UaContext* ua,
     return false;
   }
 
-  selection.bsprintf("AND Job.Type='B' AND Client.Name='%s' ",
+  selection.bsprintf("AND Job.Type IN('B','a') AND Client.Name='%s' ",
```
fbergkemper, line 1227:
```suggestion
  selection.bsprintf("AND Job.Type IN('B','A') AND Client.Name='%s' ",
```
According to include/job_types.h JT_ARCHIVE is 'A' not 'a' for Archive Jobs.

fbergkemper (approved):

PR #582, opened 2020-08-29
travis: fix the travis checks

joergsteffens:
The Travis checks did fail, because the Bareos project directory structure have changed.
The Bareos WebUI is now build and packaged in the same step as the Bareos core.

Also disables the tests for MySQL and Sqlite, as these are deprecated.

Debian build now honors the flag
DEB_BUILD_OPTIONS="nocheck"
to skip tests.


arogge (approved):
Looks great.

PR #583, opened 2020-09-01
Add configuration warnings to SD

arogge:
This PR adds configuration warnings to the SD when
* `Maximum Block Size` is configured on a device that is not a tape
* The SD is invoked with `-p` to ignore i/o errors

### DIFF
--- a/core/src/stored/stored.cc
+++ b/core/src/stored/stored.cc
@@ -253,6 +253,12 @@ int main(int argc, char* argv[])
   my_config = InitSdConfig(configfile, M_ERROR_TERM);
   ParseSdConfig(configfile, M_ERROR_TERM);
 
+  if (forge_on) {
+    my_config->AddWarning(
+        "Running with '-p' is for testing and emergency recovery purposes "
+        "only");
+  }
+
   if (export_config) {
     my_config->DumpResources(PrintMessage, NULL);
     goto bail_out;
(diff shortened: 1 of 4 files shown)


core/src/stored/stored.cc
```diff
@@ -253,6 +253,10 @@ int main(int argc, char* argv[])
   my_config = InitSdConfig(configfile, M_ERROR_TERM);
   ParseSdConfig(configfile, M_ERROR_TERM);
 
+  if (forge_on) {
+    my_config->AddWarning("Running with '-p' is for testing purposes only");
```
pstorz, line 257: How about "for testing and emergency recovery purposes only"

core/src/stored/stored_conf.cc
```diff
@@ -736,6 +736,11 @@ static bool SaveResource(int type, ResourceItem* items, int pass)
   int i;
   int error = 0;
 
+  BareosResource* allocated_resource = *resources[rindex].allocated_resource_;
+  if (pass == 2 && !allocated_resource->Validate()) {
+    return false;
```
pstorz, line 741: Does this return value mean that the storage daemon does not start?

pstorz (changes requested):

core/src/stored/stored_conf.cc
arogge: Yes. If Validate() returns false, the the SD will not start. This is analogous to the behaviour of the Director.
The idea is that we can run in-depth checks of the configuration that way (which was why that function added to the director in the first place).

core/src/stored/stored.cc
arogge: agreed.

PR #584, opened 2020-09-02
Fixes to Debian build configuration

arogge:
This PR fixes a few issues in the Debian build configuration (and some other minor improvements)


pstorz (approved):
Looks good. Very good point to store the cmake settings in the CMAKE_BINARY_DIR.

PR #585, opened 2020-09-03
core: console command show in JSON mode names resource types in plural form

joergsteffens:
This way, the result of list and show are more similar.
Disadvantage: if only a specific resource if requested, the structure still names the resource in plural.

    *show clients
    {
      "jsonrpc": "2.0",
      "id": null,
      "result": {
        "clients": {
          "bareos1-fd": {
            "name": "bareos1-fd",
            "address": "bareos1.example.com"
          },
          "bareos2-fd": {
            "name": "bareos2-fd",
            "address": "bareos2.example.com"
          }
        }
      }
    }

    *show client=bareos2-fd
    {
      "jsonrpc": "2.0",
      "id": null,
      "result": {
         "clients": {
          "bareos2-fd": {
            "name": "bareos2-fd",
            "address": "bareos2.example.com"
          }
        }
      }
    }

### DIFF
--- a/core/src/console/console_conf.cc
+++ b/core/src/console/console_conf.cc
@@ -83,11 +83,11 @@ static ResourceItem dir_items[] = {
 };
 
 static ResourceTable resources[] = {
-  { "Console", cons_items, R_CONSOLE, sizeof(ConsoleResource),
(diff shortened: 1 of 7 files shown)


arogge (approved):
Looks good to me.

PR #586, opened 2020-09-03
systemtests: detect PYTHON correctly before it is used

pstorz:


systemtests/CMakeLists.txt
```diff
@@ -623,6 +623,8 @@ else()
   set(in_container FALSE)
 endif()
 
+find_program(PYTHON NAMES python python2 python3)
```
arogge, line 626:
```suggestion
find_program(PYTHON NAMES python python3 python2)
```

PR #591, opened 2020-09-10
webui: Fix login issue

fbergkemper:
Remove a misplaced return null in AuthController.

Fixes #1039: Can not login in webui

### DIFF
--- a/webui/module/Application/src/Application/Controller/Plugin/SessionTimeoutPlugin.php
+++ b/webui/module/Application/src/Application/Controller/Plugin/SessionTimeoutPlugin.php
@@ -5,7 +5,7 @@
  * bareos-webui - Bareos Web-Frontend
  *
  * @link      https://github.com/bareos/bareos for the canonical source repository
- * @copyright Copyright (c) 2013-2019 Bareos GmbH & Co. KG (http://www.bareos.org/)
+ * @copyright Copyright (c) 2013-2020 Bareos GmbH & Co. KG (http://www.bareos.org/)
  * @license   GNU Affero General Public License (http://www.gnu.org/licenses/)
  *
  * This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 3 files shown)


webui/module/Auth/src/Auth/Controller/AuthController.php
```diff
@@ -130,11 +130,11 @@ public function loginAction()
       if(isset($bareos_updates) && $bareos_updates == false) {
          // updates could not be retrieved by ajax call
          $_SESSION['bareos']['product-updates-status'] = false;
-         return null;
```
arogge, line 133: not returning here will make the rest of the code run. Line 176 uses `$updates` as an array without checking it before.
This branch should therefore initialize it with somewhat like `$updates = []`.

fbergkemper: true, made a few adjustments

webui/module/Auth/src/Auth/Controller/AuthController.php
```diff
@@ -273,4 +222,58 @@ private function createNewLoginForm($form, $multi_dird_env = null, $err_msg = nu
...
+
+   /**
+    * DIRD update check
+    */
+   private function checkUpdateStatusDIRD() {
+      try {
+         $dird_version = $this->getDirectorModel()->getDirectorVersion($this->bsock);
+      }
+      catch(Exception $e) {
+         echo $e->getMessage();
```
arogge, line 234: This `catch`-block should set `$dird_version` to a sane value as it is used later, or it should cancel the execution of the function. 
In the latter case it might be beneficial to remove the `try`-block, as there is usually no point in catching an exception you cannot handle.

fbergkemper: Adjusted.

webui/module/Application/src/Application/Controller/Plugin/SessionTimeoutPlugin.php
```diff
@@ -26,42 +26,43 @@
...
+
+      if($this->session->offsetGet('authenticated')) {
          if($this->timeout()) {
-            return true;
-         }
-         else {
             return false;
+         } else {
+            return true;
          }
```
arogge, line 62: this could be replaced with a more readable (and shorter) `return !$this->timeout()`

arogge (approved):
Looks good.

PR #594, opened 2020-09-11
webui: Clean the URL after taking the GET parameters

fbergkemper:
Avoid duplicate actions after submission of an action followed
by a page refresh.


arogge (approved):
Works great! Totally different approach than I expected though.

PR #596, opened 2020-09-11
libcloud-plugin: fix filenames with non-ascii characters

sduehr:
Also modified the systemtest to include such filenames.
Additional changes to make the plugin compatible with both
Python 2 and 3, and some minor changes to successfully
run the systemtest.

For enabling the plugin build and systemtest, the following is
required:

- The Python 3 version of the s3cmd tool must be installed, may be
  necessary to use
  pip3 install s3cmd
  instead of OS provided s3cmd package
- Apache Libcloud must be installed, use
  pip2 install apache-libcloud
  or pip3 install apache-libcloud
- minio must be installed, use
  curl -o /usr/local/bin/minio https://dl.min.io/server/minio/release/linux-amd64/minio
  chmod 755 /usr/local/bin/minio

### DIFF
--- a/core/src/plugins/filed/python/libcloud/BareosFdPluginLibcloud.py
+++ b/core/src/plugins/filed/python/libcloud/BareosFdPluginLibcloud.py
@@ -23,7 +23,11 @@
 import BareosFdPluginBaseclass
 import bareosfd
 from bareosfd import *
-import ConfigParser as configparser
+
+try:
+    import ConfigParser as configparser
+except ImportError:
+    import configparser
 import datetime
 import dateutil.parser
 from bareos_libcloud_api.bucket_explorer import TASK_TYPE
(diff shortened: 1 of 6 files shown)


core/src/plugins/filed/python/libcloud/BareosFdPluginLibcloud.py
```diff
@@ -357,7 +363,10 @@ def create_file(self, restorepkt):
             100, "create_file() entry point in Python called with %s\n" % (restorepkt)
         )
         FNAME = FilenameConverter.BackupToBucket(restorepkt.ofname)
-        dirname = os.path.dirname(FNAME)
+        if version_info.major < 3:
+            dirname = os.path.dirname(FNAME)
+        else:
+            dirname = os.path.dirname(FNAME).encode("utf-8")
```
bklang, line 369: This logic appears inverted with the changes on line 322-325 above. Is that intentional?

sduehr: Good point, but yes, it is intentional because lines 366-369 are for restore, while 322-325 are for backup.

arogge (approved):
Looks good.

PR #598, opened 2020-09-15
python plugins: do not build for solaris

pstorz:
The current way to detect the python build parameter leads to

CC: error: unrecognized command line option '-KPIC'; did you mean '-fPIC'?

Probably that would be solved when using the sunpro compiler, but that
caused other problems.


arogge (approved):
Looks good.

PR #600, opened 2020-09-16
allow terabytes, tebibytes, petabytes,pebibytes, exabytes and exbibytes modifiers when specifying sizes

pstorz:
The configuratoin parser now additionally allows the following modifiers:

   **t**   1,099,511,627,776 (tebibytes)

  **tb**  1,000,000,000,000 (terabytes)

   **p**   1,125,899,906,842,624 (pebibytes)

   **pb**   1,000,000,000,000,000 (petabytes)

   **e**   1,152,921,504,606,846,976 (exbibytes)

   **eb**   1,000,000,000,000,000,000 (exabytes)

### DIFF
--- a/core/src/lib/edit.cc
+++ b/core/src/lib/edit.cc
@@ -3,7 +3,7 @@
 
    Copyright (C) 2002-2011 Free Software Foundation Europe e.V.
    Copyright (C) 2011-2012 Planets Communications B.V.
-   Copyright (C) 2013-2019 Bareos GmbH & Co. KG
+   Copyright (C) 2013-2020 Bareos GmbH & Co. KG
 
    This program is Free Software; you can redistribute it and/or
    modify it under the terms of version three of the GNU Affero General Public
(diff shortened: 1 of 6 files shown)


core/src/lib/edit.cc
```diff
@@ -402,6 +415,44 @@ static bool strunit_to_uint64(char* str, uint64_t* value, const char** mod)
...
+  const uint64_t multiplier[] = {1152921504606846976,  // EiB Exbibyte
+                                 1125899906842624,     // PiB Pebibyte
+                                 1099511627776,        // TiB Tebibyte
+                                 1073741824,           // GiB Gibibyte
+                                 1048576,              // MiB Mebibyte
+                                 1024,                 // KiB Kibibyte
+                                 1};
+
+  if (value == 0) {
+    result += "0";
```
arogge, line 437: can't we just `return "0"` here?

core/src/lib/res.cc
```diff
@@ -1503,39 +1503,7 @@ void IndentConfigItem(PoolMem& cfg_str,
 
 std::string PrintNumberSiPrefixFormat(ResourceItem* item, uint64_t value_in)
 {
-  PoolMem temp;
```
arogge, line 1506: this function will probably yield an unused parameter warning on compile.
You could just use `ResourceItem*` (without a parameter name) to identify the value as unused.
Probably it is even better to remove this function completely and replace with `SizeAsSiPrefixFormat()` where applicable.

core/src/tests/test_edit.cc
```diff
@@ -0,0 +1,161 @@
+/*
+   BAREOS® - Backup Archiving REcovery Open Sourced
+
+   Copyright (C) 2018-2020 Bareos GmbH & Co. KG
```
arogge, line 4:
```suggestion
   Copyright (C) 2020-2020 Bareos GmbH & Co. KG
```
it's a new file, isn't it?

core/src/tests/test_edit.cc
```diff
@@ -0,0 +1,161 @@
...
+      "1 e 1 p 1 t 1 g 1 m 1 k 1");
+}
+
+// kibibyte
+TEST(edit, convert_siunits_to_numbers)
+{
+  {
+    char str[] = "1 k";
+    uint64_t retvalue = 0;
+    size_to_uint64(str, &retvalue);
```
arogge, line 54: changing `size_to_uint64(char*, uint64_t*)` into `size_to_uint64(const char*, uint64_t*)` should be rather trivial and would allow to pass constant strings.

arogge (approved):
I have just some minor remarks. The whole `edit.cc` would probably benefit a lot if it was redesigned completely.

arogge:
I just looked at the code, but didn't test it.

PR #601, opened 2020-09-16
Improve LDAP Plugin for Filedaemon (Backport PR #595 to bareos-19.2)

arogge:


pstorz (approved):
Good work! I only added one commit so that the PYTHON detection is done before the first usage.

PR #602, opened 2020-09-16
Fixes newly introduced problems with the show command

joergsteffens:
This PR fixes a problem introduced by 486f5976a9f2ee9b944c398846c46f653f48bfb8.
The show command have to use the plural form of the resource name in JSON mode, but the original form when in plain text mode.


arogge (approved):
Looks good and all tests pass on my machine.

PR #603, opened 2020-09-17
debian: no mysql/sqlite3 dbconfig-common install

arogge:
This patch removes the dbconfig-common installation scripts for MySQL
and SQlite3.
Upgrades will still work.


pstorz (approved):
Looks OK.

PR #605, opened 2020-09-24
Run minio tests in parallel

franku:
- droplet and libcloud tests use a unique port number for each separate minio instance
- disable droplet test without minio


pstorz (approved):
Good work! Using ctest --repeat-until-fail 20 iterations runnning in parallel were successful.

PR #606, opened 2020-09-24
docs: mark directive TlsProtocol as introduced with bareos-20

joergsteffens:
Also fixes two other doc typos.


franku (approved):
Thanks for catching this up.

PR #611, opened 2020-09-28
plugin: use signed conversion for uint64_t

franku:
-1 is used as "invalid"-number for the value of the size variable, which is uint64_t. On 32 Bit systems -1 will not be extended to 64 bit if the conversion from Python is unsigned. 

This fix solves the problem of the missing sign extension during Python initialization of a C-Structure. Drawback is that sizes can only have up to 63 Bit positive numbers.


pstorz (approved):
Good work! Test also passes in my setup.

PR #612, opened 2020-09-28
tests: add 32bit state file to test

franku:
fixes recent job results test


pstorz (approved):
Good work. Problem solved, test now passes in my setup.

PR #616, opened 2020-09-30
Fix memory corruptions in SD

arogge:
When a device in the SD is configured with custom Maximum Block Size, autolabeling may cause a memory corruption. When writing to a volume with a larger blocksize than the label blocksize and then autolabels a new volume, `WriteNewVolumeLabel()` would call `EmptyBlock()` with an invalid pointer leading to heap corruption.
While pinpointing this problem, valgrind also flagged the use of uninitialized `struct mtop` in generic_tape_device. While this was probably not a real issue, the struct mtop is now always initialized, so valgrind does not complain anymore.


franku (approved):
Very good you found that!

PR #617, opened 2020-10-01
devtools: introduce devtools directory

arogge:
* add a README.md
* move prepare-release.sh to devtools dir

### DIFF
--- a/devtools/new-changelog-release.sh
+++ b/devtools/new-changelog-release.sh
@@ -1,4 +1,23 @@
 #!/bin/bash
+#   BAREOS® - Backup Archiving REcovery Open Sourced
+#
+#   Copyright (C) 2020-2020 Bareos GmbH & Co. KG
+#
+#   This program is Free Software; you can redistribute it and/or
+#   modify it under the terms of version three of the GNU Affero General Public
+#   License as published by the Free Software Foundation and included
+#   in the file LICENSE.
+#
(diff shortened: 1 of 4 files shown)


devtools/new-changelog-release.sh
```diff
@@ -0,0 +1,85 @@
...
+
+# this script edits the current release in the release notes or adds a new
+# unreleased section
+
+prog="$(basename "$0")"
+topdir="$(realpath "$(dirname "$0")/..")"
+pushd "$topdir" >/dev/null
+
+if [ $# -ne 2 ] && [ "${1:-}" != unreleased ]; then
+  echo "usage: $prog {<version> <date>|unreleased}" >&2
```
pstorz, line 32: What date format is required here? DD-MM-YYYY ? A hint would be helpful.

pstorz (commented):
Other than the one comment this looks good to me.

devtools/new-changelog-release.sh
arogge: The provided string is appended after the release-number in `CHANGELOG.md`, so `## [Unreleased]` will become `## [<version>] - <date>`. 
While this is meant to contain a date, you can pass any string you like.
As the script is meant to be called by `prepare-release.sh` and that script knows what to pass, I didn't put much effort in parameter sanitation here.

PR #620, opened 2020-10-02
droplet: rename library to libbareosdroplet.so

arogge:
This patch renames the contained libdroplet.so to libbareosdroplet.so,
so packages do not accidentially provide a 3rd party usable libdroplet.

### DIFF
--- a/core/platforms/packaging/bareos.spec
+++ b/core/platforms/packaging/bareos.spec
@@ -1314,7 +1314,7 @@ mkdir -p %{?buildroot}/%{_libdir}/bareos/plugins/vmware_plugin
 %defattr(-, root, root)
 %{backend_dir}/libbareossd-chunked*.so
 %{backend_dir}/libbareossd-droplet*.so
-%{library_dir}/libdroplet*.so
+%{library_dir}/libbareosdroplet.so*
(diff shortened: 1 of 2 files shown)


pstorz (approved):
Looks good. 
Systemtest also passes on my system.
Packaging for RPM also looks good.

FlxPeters:
I have an question related to this PR. When installing the storage daemon on Ubuntu 18.04 using the package repo from http://download.bareos.org/bareos/ version 19.02 i get the following error message when starting the bareos-sd: 
```
bareos-sd -f -s 
bareos-sd ERROR TERMINATION
Unable to load any shared library for libbareossd-droplet.so
```
As mentioned in the docs i tried to install `bareos-storage-droplet` via apt, but there is no package available with this name. 
So what is the recommended way to install libdroplet on Ubuntu 18.04?

arogge:
Hi @FlxPeters 
we're adding support für Droplet on Debian/Ubuntu in Bareos 20.
The nightly builds current don't package it yet, but it is on the TODO and will happen really soon.
If you keep an eye on the pull requests, you'll see another PR adding the packaging for Debian/Ubuntu, soon. As soon as that is merged, the nightly build will allow you to install bareos-storage-droplet on Debian and Ubuntu.

Hope that helps!

PR #621, opened 2020-10-02
Fix LSN reading for incremental backups of MySQL 8.0 servers

frank-f:
This will fix the following error message, that occurs when trying to start an incremental backup of a MySQL 8.0 server

```
Fatal error: python-fd: Error reading LSN: "
" not an integer
```


aussendorf (approved):
Manually tested on MariaDB 5.5, works as expected.

peterloeffler:
Any chance to get this fix into the bareos-filedaemon-percona-xtrabackup-python-plugin package in the yum repo https://download.bareos.org/bareos/release/20/RHEL_8/x86_64/ anytime soon?

arogge:
The fix is available in 20.0.2 and 20.0.3. However, these will not be on download.bareos.org, as these packages are only available in the subscription repositories on download.bareos.com.
Nevertheless, you can - of course - take the sources and build it yourself.

peterloeffler:
Thanks for the fast reply!

PR #622, opened 2020-10-05
Remove unused and unmaintained sample code

arogge:
- filed: remove test-deltaseq plugin
- stored: remove object_store_device
- tools: remove smtp-orig.cc
- plugins: remove unmaintained templates and example


pstorz (commented):
Good work. This cleanup helps very much.

PR #623, opened 2020-10-06
pkglist: remove libdroplet references

arogge:
since we now include libdroplet in the bareos-storage-droplet package,
there is no need for separate libdroplet packages anymore.


arogge:
This PR must be merged together with the changes from CD branch PR-623, which will in turn require the changes we applied here for successful builds..

pstorz (commented):
Looks OK. Good work!

PR #624, opened 2020-10-07
bareos 19.2: Also build for Ubuntu 20

pstorz:


arogge (approved):
this matches the other Ubuntu's.
Looks good. Finally Bareos 19.2 for Ubuntu 20

PR #625, opened 2020-10-07
tests: run systemtests during build and exclude broken tests

pstorz:
- Tests that are not reliable were marked with the label "broken"
      and are excluded when running tests during build.
    
    - Removed check for container environment to exclude tests.
    
    - Adjusted test timeout to suffice when build system is busy.
    
    - Buildsystem now requires all tests to pass

### DIFF
--- a/core/platforms/packaging/bareos.spec
+++ b/core/platforms/packaging/bareos.spec
@@ -1071,7 +1071,7 @@ pushd %{CMAKE_BUILDDIR}
 make clean
 
 # run the tests and fail build if test fails
-REGRESS_DEBUG=1 ctest --label-exclude broken --no-compress-output -j 10 --repeat until-pass:2 -D Experimental || echo "ctest result:$?"
+REGRESS_DEBUG=1 ctest --label-exclude broken --no-compress-output --parallel 10 --repeat until-pass:2 --dashboard Continous
 
 %install
 ##if 0#{?suse_version}
--- a/core/platforms/packaging/bareos.spec
+++ b/core/platforms/packaging/bareos.spec
@@ -1069,16 +1069,9 @@ cmake  .. \
 # run unit tests
 pushd %{CMAKE_BUILDDIR}
 make clean
-REGRESS_DEBUG=1 ctest -j 10 -V -D Continuous || result=$?
-result=$?
-if [ $result -eq 1 ]; then
-  echo "ctest result $result is expected and OK"
-  exit 0
-else
-  echo "ctest result $result is not 1, ERROR"
-fi
 
-
+# run the tests and fail build if test fails
+REGRESS_DEBUG=1 ctest --label-exclude broken --no-compress-output -j 10 --repeat until-pass:2 -D Experimental || echo "ctest result:$?"
 
 %install
 ##if 0#{?suse_version}
--- a/core/platforms/packaging/bareos.spec
+++ b/core/platforms/packaging/bareos.spec
@@ -1071,7 +1071,7 @@ pushd %{CMAKE_BUILDDIR}
 make clean
 
 # run the tests and fail build if test fails
-REGRESS_DEBUG=1 ctest --label-exclude broken --no-compress-output --parallel 10 --repeat until-pass:2 --dashboard Continuous || echo "ctest result:$?"
+REGRESS_DEBUG=1 ctest --label-exclude broken --parallel 10 --repeat until-pass:2 --dashboard Continuous || echo "ctest result:$?"
 
 %install
 ##if 0#{?suse_version}
--- a/core/platforms/packaging/bareos.spec
+++ b/core/platforms/packaging/bareos.spec
@@ -1071,7 +1071,7 @@ pushd %{CMAKE_BUILDDIR}
 make clean
 
 # run the tests and fail build if test fails
-REGRESS_DEBUG=1 ctest --label-exclude broken --no-compress-output --parallel 10 --repeat until-pass:2 --dashboard Continuous
+REGRESS_DEBUG=1 ctest --label-exclude broken --no-compress-output --parallel 10 --repeat until-pass:2 --dashboard Continuous || echo "ctest result:$?"
 
 %install
 ##if 0#{?suse_version}
--- a/core/platforms/packaging/bareos.spec
+++ b/core/platforms/packaging/bareos.spec
@@ -1071,7 +1071,7 @@ pushd %{CMAKE_BUILDDIR}
 make clean
 
 # run the tests and fail build if test fails
-REGRESS_DEBUG=1 ctest --label-exclude broken --no-compress-output --parallel 10 --repeat until-pass:2 --dashboard Continous
+REGRESS_DEBUG=1 ctest --label-exclude broken --no-compress-output --parallel 10 --repeat until-pass:2 --dashboard Continuous
 
 %install
 ##if 0#{?suse_version}
--- a/core/src/lib/bsys.h
+++ b/core/src/lib/bsys.h
@@ -65,4 +65,12 @@ bool PathAppend(PoolMem& path, PoolMem& extra);
 bool PathCreate(const char* path, mode_t mode = 0750);
 bool PathCreate(PoolMem& path, mode_t mode = 0750);
 
+struct StateFileHeader {
+  char id[14];
+  int32_t version;
+  uint64_t last_jobs_addr;
+  uint64_t end_of_recent_job_results_list;
+  uint64_t reserved[19];
+};
+
 #endif  // BAREOS_LIB_BSYS_H_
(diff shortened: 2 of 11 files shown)


core/platforms/packaging/bareos.spec
```diff
@@ -1065,16 +1065,9 @@ cmake  .. \
...
-if [ $result -eq 1 ]; then
-  echo "ctest result $result is expected and OK"
-  exit 0
-else
-  echo "ctest result $result is not 1, ERROR"
-fi
-
 
+# run the tests and fail build if test fails
+REGRESS_DEBUG=1 ctest --label-exclude broken --no-compress-output -j 10 --repeat until-pass:2 -D Experimental || echo "ctest result:$?"
```
arogge, line 1070: * this doesn't actually fail, when ctest fails, or does it?
* can we use `--parallel` instead of `-j` and `--dashboard` instead of `-D`
* is "Experimental" the right Dashboard? It was "Continous" previously
```suggestion
# run the tests and fail build if test fails
REGRESS_DEBUG=1 ctest --label-exclude broken --no-compress-output --parallel 10 --repeat until-pass:2 --dashboard Continous
```

core/src/tests/CMakeLists.txt
```diff
@@ -174,17 +175,13 @@ if(NOT client-only)
   bareos_add_test(
     test_db_list_ctx
     LINK_LIBRARIES
-      #dird_objects
+      # dird_objects
```
arogge, line 178:
```suggestion
```

core/src/tests/CMakeLists.txt
```diff
@@ -174,17 +175,13 @@ if(NOT client-only)
...
       bareos
-      #bareosfind
-      bareoscats
-      bareossql
-      ${GTEST_LIBRARIES}
-      ${GTEST_MAIN_LIBRARIES}
+      # bareosfind
```
arogge, line 180:
```suggestion
```

core/src/tests/recent_job_results_list.cc
```diff
@@ -32,7 +32,40 @@
...
+    EXPECT_EQ(offsetof(StateFileHeader, reserved), 40);
+  }
+
+#if HAVE_WIN32
+  detected += ", Windows";
+#endif
+  detected += " detected";
+
+  std::cout << " <<< " << detected << " >>> " << std::endl;
+}
```
arogge, line 68: In my opinion, this doesn't make sense. Most of this can (and should) be checked at compile-time. No need for a runtime check.
The premise of `Is32BitWordsize()` is just wrong. The default struct packing is implementation specific. While it may rely on the width of the processor registers, this is nothing you can rely on.

The different location of `last_jobs_addr` on Windows has nothing to do with pointer width or anything. It is plainly based on the different [default alignment for Windows](https://docs.microsoft.com/en-us/cpp/c-language/storage-and-alignment-of-structures?view=vs-2019).
On most 64-bit systems the int32_t will be aligned as 64-bits, on most 32-bit systems it will be aligned as 32-bits.
For Windows the rule is different and completely ignores the architecture: Each member of a struct is aligned as the largest member of the struct. So a struct containing only int16_t will be aligned as 16-bits, even on a 64-bit machine.
In this case, alignment will be 64 bits on windows, no matter what.

```suggestion
static inline bool Is32BitAligned()
{
  return offsetof(StateFileHeader, last_jobs_addr) == 20;
};
```

core/src/tests/recent_job_results_list.cc
```diff
@@ -41,7 +74,7 @@ TEST(recent_job_results_list, read_job_results_from_file)
 
   char orig_path[]{TEST_ORIGINAL_FILE_DIR};
 
-  const char *fname = Is32BitWordsize() ? "bareos-dir-32bit" : "bareos-dir";
+  const char* fname = Is32BitWordsize() ? "bareos-dir-32bit" : "bareos-dir";
```
arogge, line 77:
```suggestion
  const char* fname = Is32BitAligned() ? "bareos-dir-32bit" : "bareos-dir";
```

core/src/tests/recent_job_results_list.cc
```diff
@@ -62,7 +95,7 @@ TEST(recent_job_results_list, write_job_results_to_file)
   OSDependentInit();
   RecentJobResultsList::Cleanup();
 
-  const char *fname = Is32BitWordsize() ? "bareos-dir-32bit" : "bareos-dir";
+  const char* fname = Is32BitWordsize() ? "bareos-dir-32bit" : "bareos-dir";
```
arogge, line 98:
```suggestion
  const char* fname = Is32BitAligned() ? "bareos-dir-32bit" : "bareos-dir";
```

core/src/tests/recent_job_results_list.cc
```diff
@@ -123,9 +156,8 @@ TEST(recent_job_results_list, read_job_results_from_file_truncated_jobs)
 
   char orig_path[]{TEST_ORIGINAL_FILE_DIR};
 
-  const char *fname = Is32BitWordsize()
-                      ? "bareos-dir-truncated-jobs-32bit"
-                      : "bareos-dir-truncated-jobs";
+  const char* fname = Is32BitWordsize() ? "bareos-dir-truncated-jobs-32bit"
```
arogge, line 159:
```suggestion
  const char* fname = Is32BitAligned() ? "bareos-dir-truncated-jobs-32bit"
```

core/src/lib/bsys.h
```diff
@@ -65,4 +65,12 @@ bool PathAppend(PoolMem& path, PoolMem& extra);
...
 bool PathCreate(PoolMem& path, mode_t mode = 0750);
 
+struct StateFileHeader {
+  char id[14];
+  int32_t version;
+  uint64_t last_jobs_addr;
+  uint64_t end_of_recent_job_results_list;
+  uint64_t reserved[19];
+};
+
```
arogge, line 75: As this struct is used file-mapped (and we already ran into issues with it) we should check that the members are in locations we expect. With `static_assert` this is a simple compile-time check.

```suggestion

static_assert(offsetof(StateFileHeader, version) == 16);
static_assert(offsetof(StateFileHeader, last_jobs_addr) == 20 ||
                      offsetof(StateFileHeader, last_jobs_addr) == 24);
static_assert(offsetof(StateFileHeader, end_of_recent_job_results_list) == 28 ||
                      offsetof(StateFileHeader, end_of_recent_job_results_list) == 32);
static_assert(offsetof(StateFileHeader, reserved) == 36 ||
                      offsetof(StateFileHeader, reserved) == 40);
```

debian/rules
```diff
@@ -133,7 +133,7 @@ override_dh_auto_test:
 # see https://www.debian.org/doc/debian-policy/ch-source.html#s-debianrules-options
 # No tabs allowed before ifeq.
 ifneq (nocheck,$(filter nocheck,$(DEB_BUILD_OPTIONS)))
-	cd obj-x86_64-linux-gnu && make clean && REGRESS_DEBUG=1 ctest -j 10 -V -D Continuous || echo "ctest failed"
+	cd obj-$(DEB_BUILD_GNU_TYPE) && make clean && REGRESS_DEBUG=1 ctest --label-exclude broken --no-compress-output -j 10 --repeat until-pass:2 -D Experimental || echo "ctest result:$?"; cp -av Testing ../../../BUILD_RESULTS/*
```
arogge, line 136: see previous comment for SPEC

arogge (changes requested):

core/platforms/packaging/bareos.spec
pstorz: It does not fail, that is correct. The Failing is implemented in the quality gate that analyses the output of ctest and then sets the build to failed. The ctest result unfortunately is not helpful to check if tests fail or were skipped.

Using **--parallel** and **--dashboard**  instead of the short switches make sense.

However, we might discuss about if we still need the cdash dashboard at all when we do the analysis inside of jenkins.

core/platforms/packaging/bareos.spec
```diff
@@ -1065,16 +1065,9 @@ cmake  .. \
...
-if [ $result -eq 1 ]; then
-  echo "ctest result $result is expected and OK"
-  exit 0
-else
-  echo "ctest result $result is not 1, ERROR"
-fi
-
 
+# run the tests and fail build if test fails
+REGRESS_DEBUG=1 ctest --label-exclude broken --no-compress-output --parallel 10 --repeat until-pass:2 --dashboard Continuous || echo "ctest result:$?"
```
arogge, line 1070:
```suggestion
REGRESS_DEBUG=1 ctest --label-exclude broken --parallel 10 --repeat until-pass:2 --dashboard Continuous || echo "ctest result:$?"
```

arogge (commented):
i suggested to remove `--no-compess-output` as our Jenkins Pipeline can now handle decompression

arogge (approved):

PR #627, opened 2020-10-08
libcloud-plugin: replace treat_download_errors_as_warnings

franku:
- replace treat_download_errors_as_warnings by fail_on_download_error
- set fail_on_download_error to False by default because this is
  the default behaviour of bareos backups
- as a matter of fact by default download errors will be flagged
  as warnings and jobs will be marked "with warnings"

### DIFF
--- a/core/src/plugins/filed/python/libcloud/BareosFdPluginLibcloud.py
+++ b/core/src/plugins/filed/python/libcloud/BareosFdPluginLibcloud.py
@@ -298,8 +298,6 @@ class BareosFdPluginLibcloud(BareosFdPluginBaseclass.BareosFdPluginBaseclass):
                 if self.options["fail_on_download_error"]:
                     self.active = False
                     error = True
-                else:
-                    pass
             elif worker_result == ABORT:
                 self.active = False
                 error = True
--- a/core/src/plugins/filed/python/libcloud/BareosFdPluginLibcloud.py
+++ b/core/src/plugins/filed/python/libcloud/BareosFdPluginLibcloud.py
@@ -93,7 +93,7 @@ class BareosFdPluginLibcloud(BareosFdPluginBaseclass.BareosFdPluginBaseclass):
 
         super(BareosFdPluginLibcloud, self).__init__(plugindef)
         super(BareosFdPluginLibcloud, self).parse_plugin_definition(plugindef)
-        self.options["treat_download_errors_as_warnings"] = False
+        self.options["fail_on_download_error"] = False
         self.__parse_options()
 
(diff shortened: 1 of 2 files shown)


core/src/plugins/filed/python/libcloud/BareosFdPluginLibcloud.py
```diff
@@ -350,20 +350,20 @@ def start_backup_file(self, savepkt):
...
                         % (self.current_backup_task["name"]),
                     )
-                    return bRC_Skip
+                    return bRC_Error
                 else:
                     jobmessage(
-                        M_ERROR,
-                        "File %s does not exist anymore"
+                        M_WARNING,
+                        "Skipped file %s because it does not exist anymore"
```
arogge, line 363: when `bopen()` in `SaveFile()` fails for whatever reason, it emits a jobmessage of type `M_NOTSAVED`:
```
Jmsg(jcr, M_NOTSAVED, 0, _("     Cannot open \"%s\": ERR=%s.\n"),                                                                                                                                                                            
           ff_pkt->fname, be.bstrerror());       
```
So maybe we should reconsider the message type here.

docs/manuals/source/TasksAndConcepts/Plugins.rst
```diff
@@ -1636,10 +1636,11 @@ Optional Plugin Options:
 
 This option in the config file is optional:
 
-treat_download_errors_as_warnings
-   This parameter can be set to True to keep a job running if for some reason a file cannot
+fail_on_download_error
+   This parameter is by default False to keep a job running if for some reason a file cannot
    be downloaded from a bucket because it is either deleted or moved to another space during
```
arogge, line 1641:
```suggestion
   When this option is enabled, any error during a file download will fail the backup job.
   By default a warning will be issued and the next file will be backed up.
```

docs/manuals/source/TasksAndConcepts/Plugins.rst
```diff
@@ -1636,10 +1636,11 @@ Optional Plugin Options:
...
-   download. The default for this value is False.
+   download. In this case a Job will be marked with warnings as this is the default of Bareos.
+   It is recommended not to change this option.
```
arogge, line 1643:
```suggestion
```

arogge (commented):
We should take another look at the message-type for an object that cannot be loaded/opened.

core/src/plugins/filed/python/libcloud/BareosFdPluginLibcloud.py
franku: This is the tested behaviour in the Python Plugin code. Code code is probably different.

PR #628, opened 2020-10-09
configparser fixes

joergsteffens:
This PR fixes some issues with the configparser, most notable handling of lists with multiple values per line.

It also does some clarifications and testing on the Fileset File directive.


core/src/dird/inc_conf.cc
```diff
@@ -845,7 +845,7 @@ static void StoreFname(LEX* lc,
                     _("Backslash found. Use forward slashes or quote the "
                       "string.: %s\n"),
                     lc->str);
-          /* NOT REACHED */
```
franku, line 848: I thought a break would be sufficient here.

core/src/dird/inc_conf.cc
```diff
@@ -860,7 +860,7 @@ static void StoreFname(LEX* lc,
                     lc->str);
           return;
         }
-        break;
```
franku: This should be no fallthrough in case if (strchr(lc->str, '\\')) { from line 846 ist false.

core/src/dird/inc_conf.cc
```diff
@@ -913,7 +914,7 @@ static void StorePluginName(LEX* lc,
...
```
franku: No fallthrough!

franku (changes requested):
:+1: 

Please see the recommendations for the fallthrough. In my opinion this could be wrong.

Thanks for your huge work!

franku (approved):

PR #629, opened 2020-10-12
core: add break statement after scan_err

franku:
Fix some defects found with Coverity Scan.


franku:
Will be committed in #628

PR #632, opened 2020-10-19
Source sanitization based on "check-sources"

arogge:
This PR changes the sources so that our new tool "check-sources" will not complain anymore.
As a first step, changed to copyright headers are implemented.

The PR shall also implement changes to fix problems other check-sources plugins detect.
We will add Commits step by step, review and squash them.

The included ignorelist for git-blame will contain previous fixes and most changes on this branch. It must be adapted on every rebase and squash as it contains commit IDs.


core/src/filed/sd_cmds.h
```diff
@@ -1,8 +1,8 @@
 /*
    BAREOS® - Backup Archiving REcovery Open Sourced
 
-   Copyright (C) 2013-2016 Bareos GmbH & Co. KG
-   Copyright (C) 2018-2018 Bareos GmbH & Co. KG
+   Copyright (C) 2013-2018 Bareos GmbH & Co. KG
+   Copyright (C) 2013-2018 Bareos GmbH & Co. KG
```
arogge, line 5: duplicated line

joergsteffens (commented):
The Bareos AGPL header states:

 This program is Free Software; you can redistribute it and/or
 modify it under the terms of version three of the GNU Affero General Public
 License as published by the Free Software Foundation and included
 in the file LICENSE.

This might be confusing, as the Bareos GIT repository contains a number of LICENSE files (and even a larger number of LICENSE.txt files), each valid for the subdirectory in which it resides. I hope, it is obvious, which LICENSE file is meant.
I suppose that at least the file python-bareos/LICENSE.txt should be renamed to  python-bareos/LICENSE.

arogge:
I implemented the changes from the reviews.
I also removed capsulethunk.h, because it seems to be not needed (builds will show if I'm right) and removed the core-dependency in ndmp that was added in Bareos 19.2

franku (approved):
Looks good!

PR #635, opened 2020-10-21
FreeBSD: build and package fd python plugin for python3

pstorz:


core/platforms/freebsd/bareos-freebsd/bareos.com-common/BareosCommonMakefile
```diff
@@ -75,7 +75,7 @@ CPPFLAGS+=      -I${LOCALBASE}/include
 
 LIB_DEPENDS+=	liblzo2.so:archivers/lzo2 \
 		          libjansson.so:devel/jansson
-LIB_DEPENDS+= libpython2.7.so:lang/python27
+LIB_DEPENDS+= libpython3.7m.so:lang/python37
```
arogge, line 78: how can we build python 2.7 if we don't depend on it?

arogge (approved):
Besides the LIB_DEPENDS, this looks good.

core/platforms/freebsd/bareos-freebsd/bareos.com-common/BareosCommonMakefile
pstorz: We decided to not support python2.7 on FreeBSD anymore as it will be removed by end of the year.
See:
https://www.freshports.org/lang/python27/: 

DEPRECATED: EOLed upstream
EXPIRATION DATE: 2020-12-31

PR #636, opened 2020-10-21
gtest: CFG_TYPE_STR_VECTOR_OF_DIRS: skip check for default value of backend_directories on windows

joergsteffens:
The test_CFG_TYPE_STR_VECTOR_OF_DIRS test
reads a configuration file containing a list of BackendDirectories,
with the first entry being the default value (PATH_BAREOS_BACKENDDIR).
However, WIN32 compilation overwrites the default value,
so that this can not succeed in WIN32 and is therefore excluded.


pstorz (approved):
Looks good

PR #638, opened 2020-10-23
fd local fileset python plugin: add a forgotten import statement

franku:


pstorz (changes requested):
PR tiltle is not specific. I would like to suggest:
**BareosFdPluginLocalFileset: import new BareosFdPluginLocalFilesBaseclass**

Other than that, all local-fileset tests pass on my machine. Good work.

pstorz (approved):
Very good.

PR #640, opened 2020-10-27
lib/tools: remove unused code to reduce license-diversity a bit

arogge:
* remove md5 and sha1 implementations
   OpenSSL is used instead, so this is not necessary anymore
* remove tool timelimit
   This was only used in regression testing as an optional component
* remove FreeBSD tapetest.c
   This was not built at all and btape offers a better test


pstorz (approved):
Looks good to me.

The only thing that is a small drawback is that now we use the magic number 16 instead of "MD5HashSize"

joergsteffens (approved):
Looks good.

PR #645, opened 2020-10-30
webui: adapt changes from master on webui-selenium-test.py

fbergkemper:


pstorz (approved):
I guess this is the full current file from master, right? It makes sense to use it.

fbergkemper:
> 
> 
> I guess this is the full current file from master, right? It makes sense to use it.

Yes. it is.

PR #650, opened 2020-10-30
packaging: set php-fpm as a required dependency on Fedora >= 33 (backport 18.2)

fbergkemper:
There is no mod_php in default configuration. mod_php is deprecated as
FPM is now used by default with httpd in event mode. mod_php is only
used when explicitly enabled or httpd switch to prefork mode.


pstorz (approved):
Looks good to me.

PR #651, opened 2020-11-01
python-bareos: fixes support for sslpsk with Python >= 3

joergsteffens:
Current sslpsk version require identity and passwort as bytes. Strings are no longer supported.
The python-bareos os now adapted to this.
    
Unfortenatly, depending on the Python and OpenSSL versions
sslpsk (1.0.0) does not work with all TLS protocol versions.
Normally, the best matching protocol version is selected automatically,
but this fails in many environments using sslpsk.
As a workaround it is now possible to specify the TLS protocol version (v1, v1.1, v1.2)
python-bareos should use.
Note: newer newer protocols (TLS v1.3) can not be specified in this may.

See also: https://github.com/drbild/sslpsk/pull/22

PR #652, opened 2020-11-01
python-bareos: fixes support for sslpsk with Python >= 3 (bareos-19.2)

joergsteffens:
Current sslpsk version require identity and passwort as bytes. Strings are no longer supported.
The python-bareos os now adapted to this.

Unfortenatly, depending on the Python and OpenSSL versions
sslpsk (1.0.0) does not work with all TLS protocol versions.
Normally, the best matching protocol version is selected automatically,
but this fails in many environments using sslpsk.
As a workaround it is now possible to specify the TLS protocol version (v1, v1.1, v1.2)
python-bareos should use.
Note: newer newer protocols (TLS v1.3) can not be specified in this may.

See also: drbild/sslpsk#22

Downstream PR for https://github.com/bareos/bareos/pull/651

### DIFF
--- a/python-bareos/bareos/bsock/__init__.py
+++ b/python-bareos/bareos/bsock/__init__.py
@@ -3,9 +3,12 @@
 from bareos.exceptions import *
 from bareos.util.password import Password
 from bareos.bsock.connectiontype import ConnectionType
+from bareos.bsock.constants import Constants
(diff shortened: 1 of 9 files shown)


pstorz (approved):
Looks good and works in my Environment

PR #658, opened 2020-11-06
Remove default tcp port numbers from systemtests environment 

franku:
Each daemon has its own tcp port number for each systemtest. But some remainder from previously ported systemtests were confusing because static assignments were used. These value assignments are now removed from the systemtests cmake script.


pstorz (approved):
Looks good to me.

PR #665, opened 2020-11-11
adapt oVirt plugin for python3

pstorz:


sduehr:
As of https://bugzilla.redhat.com/show_bug.cgi?id=1862722 the use of "ImageTransfer.signed_ticket"
is not necessary since oVirt 4.2. Removed it and verified that it work without also when using oVirt SDK 4.2 and 4.3.
Pushed the changed code.

Remaining Problem:
When using non-ascii characters in disk alias name, with Python 2 the backup fails with
```
  File "/home/stephand/github/bareos/core/src/plugins/filed/python/pyfiles/BareosFdWrapper.py", line 42, in start_backup_file
    return bareos_fd_plugin_object.start_backup_file(savepkt)
  File "/home/stephand/github/bareos/core/src/plugins/filed/python/ovirt/BareosFdPluginOvirt.py", line 182, in start_backup_file
    self.backup_obj["file"]["fh"] = io.BytesIO(vmfile["data"].encode('utf-8'))
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 1568: ordinal not in range(128)

```

and with Python 3 the restore fails with
```

01-Dec 14:58 py3plug-fd-ovirt-fd JobId 2: Fatal error: bareosfd: Traceback (most recent call last):
  File "/home/stephand/github/bareos/core/src/plugins/filed/python/pyfiles/BareosFdWrapper.py", line 66, in create_file
    return bareos_fd_plugin_object.create_file(restorepkt)
  File "/home/stephand/github/bareos/core/src/plugins/filed/python/ovirt/BareosFdPluginOvirt.py", line 302, in create_file
    if not os.path.exists(dirname):
  File "/usr/lib64/python3.6/genericpath.py", line 19, in exists
    os.stat(path)
UnicodeEncodeError: 'ascii' codec can't encode character '\xf6' in position 107: ordinal not in range(128)
```

sduehr (changes requested):
I've recently pushed my suggested changes and fixes, they should be reviewed by somebody else, please.

core/src/plugins/filed/python/ovirt/BareosFdPluginOvirt.py
```diff
@@ -107,11 +118,20 @@ def check_options(self, mandatory_options=None):
...
+        but it's used to report if the Python oVirt SDK is not installed.
         If it is required to know if it is a backup or a restore, it
         may make more sense to invoke the options checking from
         start_backup_job() and start_restore_job()
         """
+
+        if SDK_IMPORT_ERROR:
+            bareosfd.JobMessage(
+                bareosfd.M_FATAL,
+                "Please install the Python SDK for oVirt Engine API.\n",
```
arogge, line 131: That message doesn't tell me what package to get from pip. Maybe we should mention `pyvmomi` here?

core/src/plugins/filed/python/ovirt/BareosFdPluginOvirt.py
```diff
@@ -1126,7 +1181,7 @@ def start_download(self, snapshot, disk):
         )
 
         bareosfd.JobMessage(
-            M_INFO,
+            bareosfd.M_INFO,
             "   Transfer disk snapshot of %s bytes\n" % (str(self.bytes_to_transf)),
```
pstorz, line 1185:
```suggestion
            "Transfer disk snapshot of %s bytes\n" % (str(self.bytes_to_transf)),
```

core/src/plugins/filed/python/ovirt/BareosFdPluginOvirt.py
```diff
@@ -1644,16 +1703,17 @@ def start_upload(self, disk):
         self.proxy_connection.endheaders()
 
         bareosfd.JobMessage(
-            M_INFO,
+            bareosfd.M_INFO,
             "   Upload disk of %s bytes\n" % (str(self.bytes_to_transf)),
```
pstorz, line 1707:
```suggestion
            "Upload disk of %s bytes\n" % (str(self.bytes_to_transf)),
```

pstorz (commented):
Are these indentations intentionally there?

core/src/plugins/filed/python/ovirt/BareosFdPluginOvirt.py
pstorz: It is intentionally indented

PR #666, opened 2020-11-11
Some improvements for systemtests

franku:
- achieve a unique postgresql user when running systemtest scripts and bareos
- document how to run the postgresql-server for systemtests
- add some more debugging output when connecting to the catalog

### DIFF
--- a/docs/manuals/source/DeveloperGuide/BuildAndTestBareos/systemtests.rst
+++ b/docs/manuals/source/DeveloperGuide/BuildAndTestBareos/systemtests.rst
@@ -71,6 +71,25 @@ sloppiness = file_macro
   hash_dir = false
   sloppiness = file_macro
 
+Prepare the PostgreSQL database for systemtests
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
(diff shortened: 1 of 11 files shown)


docs/manuals/source/DeveloperGuide/BuildAndTestBareos/systemtests.rst
```diff
@@ -71,6 +71,31 @@ sloppiness = file_macro
...
+  psql --command='CREATE DATABASE regress OWNER regress TEMPLATE template0;'
+
+  # insert authentication methods for database user 'regress'
+  # BEFORE the general rules into file /var/lib/pgsql/data/pg_hba.conf
+  #
+  # local   all             regress                                 trust
+  # host    all             regress         127.0.0.1/32            trust
+  # host    all             regress         ::1/128                 trust
+
 Systemtests
```
pstorz, line 99: I would remove everything that is OS specific and only describe how to configure the database.

So "as use postgres do: createuser regress -s, psql, etc"
Then the title can be "How to configure postgresql for systemtests"

This is true for every OS that can run postgresql.

pstorz (changes requested):

pstorz (approved):
Looks good to me.

PR #667, opened 2020-11-11
VMware plugin adapted to modernized plugin interface

sduehr:
Also adapted bareos_vadp_dumper.py for compatibility with Python 2 and 3.
Additionally includes a fix for https://bugs.bareos.org/view.php?id=1057
by adding an optionial plugin option that allows disabling guest quescing on snapshot creation.


pstorz (approved):
Looks good to me.

core/src/vmware/vmware_cbt_tool/vmware_cbt_tool.py
```diff
@@ -136,14 +144,14 @@ def main():
...
     # see https://github.com/vmware/pyvmomi/issues/212
-    py_ver = sys.version_info[0:3]
-    if py_ver[0] == 2 and py_ver[1] == 7 and py_ver[2] >= 5:
-        import ssl
-
-        try:
-            ssl._create_default_https_context = ssl._create_unverified_context
-        except AttributeError:
-            pass
+    #    py_ver = sys.version_info[0:3]
```
pstorz, line 147: Do we want to remove these comments?

sduehr: Indeed, the comment lines 143-154 can be removed. This is now handled by using SmartConnectNoSSL from pyvmomi to skip SSL verification after issuing a warning when SSL verification fails.

https://github.com/bareos/bareos/blob/5548b1544433178d1b86023cf9ec61b18dc0df51/core/src/vmware/vmware_cbt_tool/vmware_cbt_tool.py#L143-L154

docs/manuals/source/TasksAndConcepts/Plugins.rst
```diff
@@ -811,7 +811,7 @@ And the config file as follows:
 
    Do not use quotes in the above config file, it is processed by the Python ConfigParser module and the quotes would not be stripped from the string.
 
-
+Since :sinceVersion:`20.2.0: VMware Plugin: quiesce` To allow backing up VMs which do not support quiesced snapshots, it is now possible to use the plugin option **quiesce**. By default quiescing when not explicitly using this option, quiescing is enabled to create backups that are as consistent as possible. When setting **quiesce=no** it is more likely to backup an inconsistent state. In this case, the backup job log will contain an appropriate warning and the the job termination will be **Backup OK -- with warnings**.
```
arogge, line 814: wrong :sinceVersion:
```suggestion
Since :sinceVersion:`20: VMware Plugin: quiesce` To allow backing up VMs which do not support quiesced snapshots, it is now possible to use the plugin option **quiesce**. By default quiescing when not explicitly using this option, quiescing is enabled to create backups that are as consistent as possible. When setting **quiesce=no** it is more likely to backup an inconsistent state. In this case, the backup job log will contain an appropriate warning and the the job termination will be **Backup OK -- with warnings**.
```

PR #669, opened 2020-11-13
webui: change updateBvfsCache trigger in restore controller

gnieark:
Hello,
Actually updateBvfsCache is done when the main restoration page is open. As no jobid is given, it's a full bvfs_update....  and it's long before the page can bu used.

By reversing the condition. It's done only when a jobid is selected. As updateBvfsCache() method check the current jobid, and target it, It's very faster!

### DIFF
--- a/webui/module/Restore/src/Restore/Controller/RestoreController.php
+++ b/webui/module/Restore/src/Restore/Controller/RestoreController.php
@@ -550,10 +550,9 @@ class RestoreController extends AbstractActionController
     $this->setRestoreParams();
     $this->getJobIds();
 
-    if($this->restore_params['id'] == null || $this->restore_params['id'] == "#") {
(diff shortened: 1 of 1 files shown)


fbergkemper:
Good catch and change.

But please reword the commit message to something like ...

webui: change updateBvfsCache trigger condition in restore controller
    
    By reversing the condition updateBvfsCache triggers only when a jobid
    is selected, this avoids unnecessary cache updates.

It would also be kind if you please add yourself to the AUTHORS file.

gnieark:
@fbergkemper 
Thanks for the feed-back
Commit Title and description are changed now.

For adding me on the Author File, no thanks, not for a simple line changed ;) , I'll do that next time i'll contribute.

fbergkemper (approved):

PR #670, opened 2020-11-17
core: Fix wrong droplet.profile path in droplet example configuration

fbergkemper:
This commit fixes a wrong path setting to the droplet.profile in the
device options directive of the example configuration.

### DIFF
--- a/core/src/stored/backends/droplet_device.d/bareos-sd.d/device/S3_ObjectStorage.conf.example
+++ b/core/src/stored/backends/droplet_device.d/bareos-sd.d/device/S3_ObjectStorage.conf.example
@@ -18,10 +18,10 @@ Device {
   #
 
   # testing:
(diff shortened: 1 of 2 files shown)


rkarlsba:
hi

Seems ubuntu 20.04 doesn't have Bareos and the versjon of Bacula-fd that comes with u20.04 won't work with bareos. There's no bareos for u20.04 on the bareos download site either.

Is this the way it's meant to be?

Vennlig hilsen

roy
-- 
Roy Sigurd Karlsbakk
(+47) 98013356
http://blogg.karlsbakk.net/
GPG Public key: http://karlsbakk.net/roysigurdkarlsbakk.pubkey.txt
--
Hið góða skaltu í stein höggva, hið illa í snjó rita.

franku (approved):
This is a reasonable change.

PR #672, opened 2020-11-18
deprecate rados storage daemon backend

franku:


arogge:
Looks good. I'll have to try it out though.

core/src/stored/device_resource.cc
```diff
@@ -294,6 +294,10 @@ bool DeviceResource::Validate()
     my_config->AddWarning(
         "Setting 'Maximum Block Size' on a non-tape device is unsupported");
   }
+  if (dev_type == DeviceType::B_RADOS_DEV) {
+    my_config->AddWarning(
+        "The Rados Storage Backend Device is deprecated since Bareos 20");
```
arogge, line 299: I don' think we should duplicate the version information here, people will have to look at the docs anyway and I don't see how having the versison number that started deprecating the feature is beneficial to the user.
I would also not tell the user that we will remove it in Bareos 21, because chances are that we won't (e.g. if a customer requires more time to move to S3)
```suggestion
        "The Rados Storage Backend Device is deprecated");
```

arogge (approved):
I would argue that we shouldn't reference version numbers in deprecation messages, but that's nothing that hinders a merge.
For testing I improved the existing deprecation systemtest to also check for this specific deprecation. That commit is - of course - not yet reviewed.

franku:
The added systemtest looks reasonable for me. Will wait for the build chain to succeed.

PR #674, opened 2020-11-20
Droplet connection issues

franku:


Adiost:
So now everyone would have to make a `bareos-test` directory to use the droplet backend?

I just patched 19.2.7 to use `marker.bareos` and put an empty file with that name into every bucket I use, which is the same thing essentially, but I feel like it's an awful, temporary solution.

arogge:
> So now everyone would have to make a `bareos-test` directory to use the droplet backend?
No. You didn't read the code right.
It checks for a directory "bareos-test" and if that exists or does not exist it is good to go.
If any other error occurs, it is not good to go.

arogge:
> My bad. Still, this solution is just as roundabout as the old one, which suddenly stopped working.

Feel free to come up with a better solution. Right now we're still trying to understand the cause of the problem, because in our test environment everything works fine.

PR #675, opened 2020-11-23
debian: set version number in "Provides:"

pstorz:
Also provide virtual packages in python3 versions of the packages.


arogge (approved):
Looks good and works in my test scenario.

PR #676, opened 2020-11-25
Fix memory access violations in mem_pool.cc and ua_acl.cc

pstorz:


core/src/dird/ua_acl.cc
```diff
@@ -55,7 +55,7 @@ static inline bool is_regex(const char* regexp)
   bool retval = false;
 
   p = regexp;
-  while (p) {
+  while (p && *p) {
```
arogge, line 58: I really don't like this. It is neither clearly understandable what was was broken and how this change fixes it.
I would rather refactor the whole function (it is only a few lines)

arogge (commented):
I approve the change to `mem_pool.cc`, but I replaced `is_regex()` with a version based on `std::string`.

core/src/dird/ua_acl.cc
```diff
@@ -49,34 +51,9 @@ bool UaContext::AclAccessOk(int acl, const char* item, bool audit_event)
  * A regexp uses the following chars:
  * ., (, ), [, ], |, ^, $, +, ?, *
  */
-static inline bool is_regex(const char* regexp)
+static bool is_regex(std::string regexp)
```
franku, line 54:
```suggestion
static bool is_regex(std::string probe_for_regexp)
```

arogge: i renamed the parameter

core/src/dird/ua_acl.cc
```diff
@@ -49,34 +51,9 @@ bool UaContext::AclAccessOk(int acl, const char* item, bool audit_event)
  * A regexp uses the following chars:
  * ., (, ), [, ], |, ^, $, +, ?, *
  */
-static inline bool is_regex(const char* regexp)
+static bool is_regex(std::string probe_for_regexp)
```
pstorz, line 54:
```suggestion
static bool is_regex(std::string string_to_check)
```

core/src/dird/ua_acl.cc
```diff
@@ -49,34 +51,9 @@ bool UaContext::AclAccessOk(int acl, const char* item, bool audit_event)
...
-        retval = true;
-        goto bail_out;
-      default:
-        break;
-    }
-  }
-
-bail_out:
-  return retval;
+  return std::string::npos != probe_for_regexp.find_first_of(".()[]|^$+?*");
```
pstorz, line 56:
```suggestion
  return std::string::npos != string_to_check.find_first_of(".()[]|^$+?*");
```

pstorz (commented):
I have checked for the variable name and "probe_for_regexp" sounds like a function name for the function and not like the string that will be checked if it is a regex or not.
So my suggestion is "string_to_check"

core/src/lib/mem_pool.cc
```diff
@@ -284,16 +284,12 @@ void GarbageCollectMemoryPool()
 void CloseMemoryPool()
 {
   struct abufhead *buf, *next;
-  int count = 0;
-  uint64_t bytes = 0;
 
   P(mutex);
   for (int i = 1; i <= PM_MAX; i++) {
     buf = pool_ctl[i].free_buf;
```
arogge, line 290:
```suggestion
    abufhead* buf = pool_ctl[i].free_buf;
```

core/src/lib/mem_pool.cc
```diff
@@ -284,16 +284,12 @@ void GarbageCollectMemoryPool()
...
     while (buf) {
       next = buf->next;
```
arogge, line 292:
```suggestion
      abufhead* next = buf->next;
```

core/src/lib/mem_pool.cc
```diff
@@ -284,16 +284,12 @@ void GarbageCollectMemoryPool()
 void CloseMemoryPool()
 {
   struct abufhead *buf, *next;
```
arogge, line 286:
```suggestion
```
I think we can reduce the scope of these a lot.

core/src/lib/mem_pool.cc
```diff
@@ -284,16 +284,12 @@ void GarbageCollectMemoryPool()
...
-  uint64_t bytes = 0;
 
   P(mutex);
   for (int i = 1; i <= PM_MAX; i++) {
     buf = pool_ctl[i].free_buf;
     while (buf) {
       next = buf->next;
-      count++;
-      bytes += SizeofPoolMemory((char*)buf);
       free((char*)buf);
```
arogge, line 293:
```suggestion
      free(buf);
```
I don't see how casting to `char*` makes any difference here. Casting `abufhead*` to `void*` should have the same effect as casting to `char*` and then to `void*`, so the explicit cast could be avoided.

arogge (commented):
I have additional suggestions when it comes to `CloseMemoryPool()`. We can probably simplify even more.

arogge:
I just godbolted the additional changes to CloseMemoryPool() and as I suspected they don't change the generated assembly at all.

rainerjung:
We observed a crash in SizeofPoolMemory() called from CloseMemoryPool() yesterday (in 19.2.8) exactly because of "cp -= HEAD_SIZE" in SizeofPoolMemory() resulted in an invalid address. So I support this (obvious) fix for CloseMemoryPool().

In our case the crash happened during two concurrent runs of updating the bvfs cache. I have no idea, whether that is related to triggering this bug and I noticed, that concurrent runs will no longer happen due to locking in the forthcoming version. Just wanted to mention it.

arogge:
> We observed a crash in SizeofPoolMemory() called from CloseMemoryPool() yesterday (in 19.2.8) exactly because of "cp -= HEAD_SIZE" in SizeofPoolMemory() resulted in an invalid address. So I support this (obvious) fix for CloseMemoryPool().
> 
> In our case the crash happened during two concurrent runs of updating the bvfs cache. I have no idea, whether that is related to triggering this bug and I noticed, that concurrent runs will no longer happen due to locking in the forthcoming version. Just wanted to mention it.
Even though the bug is obvious (and I cannot believe we didn't notice this earlier), the crash happens only in special circumstances. The memory is "only" read, so if the address is mapped for the process, it will not result in a crash.

I think (and that is definitely not knowing) that when we `malloc()` a lot of this "pool memory" and then garbage collect it, we free up a whole page that is then unmapped by the OS.
When there is another "pool memory" at the very start of the next page and that object is garbage collected, the wrong offset (24 bytes before the start of the page) will access the unmapped memory eventually leading to a segmentation fault.

core/src/lib/mem_pool.cc
```diff
@@ -194,23 +192,19 @@ int32_t SizeofPoolMemory(POOLMEM* obuf)
...
-  void* buf;
-  int pool;
-
   ASSERT(obuf);
-  P(mutex);
-  cp -= HEAD_SIZE;
-  buf = realloc(cp, size + HEAD_SIZE);
+  char* cp = (char*)obuf - HEAD_SIZE;
+
+  void* buf = realloc(cp, size + HEAD_SIZE);
```
arogge, line 198: if `cp` is not used otherwise, we could move `(char*)obuf - HEAD_SIZE` into the `realloc()` call.
```suggestion
  void* buf = realloc((char*)obuf - HEAD_SIZE, size + HEAD_SIZE);
```

core/src/lib/mem_pool.cc
```diff
@@ -194,23 +192,19 @@ int32_t SizeofPoolMemory(POOLMEM* obuf)
...
   if (buf == NULL) {
-    V(mutex);
     MemPoolErrorMessage(__FILE__, __LINE__,
                         _("Out of memory requesting %d bytes\n"), size);
     return NULL;
   }
 
   ((struct abufhead*)buf)->ablen = size;
-  pool = ((struct abufhead*)buf)->pool;
+  int pool = ((struct abufhead*)buf)->pool;
```
arogge, line 206: I like to `const` values, that don't change anymore, so maybe `pool` could be const?
```suggestion
  const int pool = ((struct abufhead*)buf)->pool;
```

core/src/lib/mem_pool.cc
```diff
@@ -228,28 +222,30 @@ POOLMEM* CheckPoolMemorySize(POOLMEM* obuf, int32_t size)
...
-  int pool;
-
   ASSERT(obuf);
-  P(mutex);
-  buf = (struct abufhead*)((char*)obuf - HEAD_SIZE);
-  pool = buf->pool;
-  pool_ctl[pool].in_use--;
+  struct abufhead* buf = (struct abufhead*)((char*)obuf - HEAD_SIZE);
+
+  int pool = buf->pool;
```
arogge, line 228:
```suggestion
  const int pool = buf->pool;
```

arogge (changes requested):
* entry in ChangeLog is missing
* bug should be linked to the release-ticket for the next release
* commit messages can be improved
  * "fix integer" -> "fixed integer"
  * "lock mutex for pool 0 too" -> "lock mutex for pool 0, too"

I also have some minor comments in the code.

arogge (approved):

PR #677, opened 2020-11-26
Update default config fileset for windows

VeselaHouba:
Exclude `hyberfile.sys` and `swapfiles.sys` from backups (used on latest windows instead of `pagefile.sys`)

### DIFF
--- a/core/src/defaultconfigs/bareos-dir.d/fileset/Windows All Drives.conf
+++ b/core/src/defaultconfigs/bareos-dir.d/fileset/Windows All Drives.conf
@@ -6,7 +6,9 @@ FileSet {
       Signature = MD5
       Drive Type = fixed
       IgnoreCase = yes
+      WildFile = "[A-Z]:/hiberfil.sys"
       WildFile = "[A-Z]:/pagefile.sys"
+      WildFile = "[A-Z]:/swapfile.sys"
       WildDir = "[A-Z]:/RECYCLER"
       WildDir = "[A-Z]:/$RECYCLE.BIN"
       WildDir = "[A-Z]:/System Volume Information"


core/src/defaultconfigs/bareos-dir.d/fileset/Windows All Drives.conf
```diff
@@ -6,7 +6,9 @@ FileSet {
       Signature = MD5
       Drive Type = fixed
       IgnoreCase = yes
+      WildFile = "[A-Z]:/hyberfile.sys"
```
arogge, line 9: don't you man hiberfil.sys?
```suggestion
      WildFile = "[A-Z]:/hiberfil.sys"
```

VeselaHouba: Ah yes, thanks for checking. Stupid naming.

arogge:
Thank you very much for taking the time to provide a patch for this!

Would you take a look at https://docs.bareos.org/DeveloperGuide/gitworkflow.html#commit-message-guideline and squash the two commits into one while updating the commit message to follow our guideline?

VeselaHouba:
@arogge Did my best :)
The guide is bit funny

> You should write the summary in imperative mood: “Fix bug” and not “Fixed bug” or “Fixes bug.”

and then example two lines below:

> Fixes #12345: The exact title of the bug you fix.

arogge:
Thank you!
Yes - that actually sucks. The summary (i.e. the first line) should be in imperative mood, but we usually write the attention sequence for the bug tracker as "Fixes". One of the many things we should probably improve :)

pstorz (approved):

PR #678, opened 2020-11-26
Poolmem suggestions

franku:
Some suggestions how to work with mempool mutex to talk about tomorrow.

### DIFF
--- a/core/src/lib/mem_pool.cc
+++ b/core/src/lib/mem_pool.cc
@@ -173,10 +173,12 @@ POOLMEM* GetMemory(int32_t size)
   buf->ablen = size;
   buf->pool = 0;
   buf->next = NULL;
+  P(mutex);
   pool_ctl[0].in_use++;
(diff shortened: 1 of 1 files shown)


core/src/lib/mem_pool.cc
```diff
@@ -228,28 +225,30 @@ POOLMEM* CheckPoolMemorySize(POOLMEM* obuf, int32_t size)
...
+    return;
+  }
+  /* otherwise link it to the free pool chain */
+  P(mutex);
+  struct abufhead* next;
+  for (next = pool_ctl[pool].free_buf; next; next = next->next) {
+    if (next == buf) {
+      V(mutex);
+      return; // skip same buffer
     }
```
arogge, line 247: I would really like to keep the `assert()` as the condition checked for shows a damaged linked-list, which is worth crashing the software in my opinion.
If we moved to `std::mutex` and `std::lock_guard`, we could limit it to:
```suggestion
    assert(next != buf);
```

PR #679, opened 2020-11-30
Fix memory access violations in mem_pool.cc and ua_acl.cc (backport from master)

franku:
Fixes #1281: Director crashes on memory access violation
             in CloseMemoryPool() and is_regex()

mem_pool: Remove calculation of the variables "count"
and "bytes" which caused the crash while the results are
not used at all. Also reduces variable scope and removes
a pointless cast.

ua_acl: fix is_regex() function, replace error-prone and
buggy function with simpler c++-based string-handling.

### DIFF
--- a/core/src/dird/ua_acl.cc
+++ b/core/src/dird/ua_acl.cc
@@ -33,6 +33,8 @@
 #include "dird/dird_globals.h"
 #include "lib/edit.h"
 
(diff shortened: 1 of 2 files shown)


arogge:
I think the PR title is misleading now.

arogge (approved):

PR #680, opened 2020-11-30
Fix memory access violations in mem_pool.cc and ua_acl.cc (backport from master)

franku:
Fixes #1281: Director crashes on memory access violation
in CloseMemoryPool() and is_regex()


pstorz (approved):
Looks good to me

PR #684, opened 2020-12-02
webui: sync webui-selenium-test.py with bareos-19.2

pstorz:


fbergkemper (approved):
That change definitely makes sense, since we don't just have one Input Type Select element in the login dialog, but two instead. The recent backport was from master and it was overseen that the login form construction is handled differently on that branch.

PR #685, opened 2020-12-04
LibCloud Populate statp on new file

bsperduto:
In current implementation statp is only populated with default stats instead of object's actual stats. This causes an accurate incremental backup to backup every file as the modified time does not match the time stored in the catalog.


sduehr (approved):
Tested the changed, confirming that accurate mode works properly with this fix.
Thanks a lot for your contribution.

PR #686, opened 2020-12-04
Storage daemon DeviceResource: Set autoselect using a ui command

franku:
This PR implements the ui command:
`setdevice storage=<storage-name> device=<device-name> autoselect=<bool>`

It will set the value of DeviceResource::autoselect to false or true, thus dis- or enabling an autochanger from automatically choosing a device during reservation.


core/src/stored/dir_cmd.cc
```diff
@@ -455,6 +459,50 @@ static bool SetdebugCmd(JobControlRecord* jcr)
...
+
+  auto res = GetDeviceResource(device_name.data());
+  if (res == nullptr) {
+    std::string err{"Device not found: "};
+    err += device_name.data();
+    dir->fsend(BADcmd, "setdevice", err.c_str());
+    return false;
+  }
+
+  if (autoselect_value.data() == std::string("on")) {
```
pstorz, line 491: Maybe we should also accept "1" for true and "0" for false as this is what is being printed in the status output.
Accepting yes and no probably also would be userfriendly.

arogge: as a user I'd assume that I can use every syntax that is allowed for a boolean value in the configuration file

core/src/lib/util.h
```diff
@@ -75,5 +77,33 @@ void StringToLowerCase(std::string& s);
...
 bool pm_append(void* pm_string, const char* fmt, ...);
 
+class BoolString {
+ public:
+  BoolString(const std::string& s) : str_value(s)
+  {
+    if (true_values.find(str_value) == true_values.end()
+        && false_values.find(str_value) == false_values.end()) {
+      std::string err = {"Wrong parameter: "};
+      throw std::out_of_range(err + str_value);
```
pstorz, line 87: As we tried to stick to the google style guide, I would like to refer to:
https://google.github.io/styleguide/cppguide.html#Exceptions

pstorz (commented):
I think  it makes sense to talk about the introduction of exceptions.

core/src/lib/util.h
```diff
@@ -75,5 +77,33 @@ void StringToLowerCase(std::string& s);
 void StringToLowerCase(std::string& out, const std::string& in);
 bool pm_append(void* pm_string, const char* fmt, ...);
 
+class BoolString {
```
pstorz, line 80: There is no test that tests the functionality of the BoolString  Class.

core/src/lib/bool_string.cc
```diff
@@ -0,0 +1,26 @@
...
+   along with this program; if not, write to the Free Software
+   Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+   02110-1301, USA.
+*/
+
+#include "lib/bool_string.h"
+#include <set>
+
+const std::set<std::string> BoolString::true_values{"on", "true", "yes", "1"};
+const std::set<std::string> BoolString::false_values{"off", "false", "no", "0"};
```
arogge, line 26: The initialization order of global variables over multiple translation units is unspecified.
So if someone uses `BoolString` to initialize a global variable, it may use `true_values` and `false_values` before they are initialized.

core/src/lib/bool_string.h
```diff
@@ -0,0 +1,57 @@
...
+   WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+   Affero General Public License for more details.
+
+   You should have received a copy of the GNU Affero General Public License
+   along with this program; if not, write to the Free Software
+   Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+   02110-1301, USA.
+*/
+
```
arogge, line 21: Wouldn't the following function do essentially the same?
```
bool str2bool(const std::string& s) {
  static std::map<std::string, bool> m{{"yes", true}, {"no", false}, ...};
  return m.at(s);
}
```

core/src/stored/dir_cmd.cc
```diff
@@ -455,6 +462,45 @@ static bool SetdebugCmd(JobControlRecord* jcr)
...
+  }
+
+  try {
+    BoolString s{autoselect_value.data()};
+    res->autoselect = s.get<bool>();
+    if (res->dev) { res->dev->autoselect = res->autoselect; }
+    return dir->fsend(OKsetdevice, device_name.data(), s.get().c_str());
+  } catch (const std::out_of_range& e) {
+    return dir->fsend(BADcmd, "setdevice", e.what());
+  }
```
arogge, line 501: I wouldn't allow all values for true/false in the protocol.
I'd convert it when the commandline is parsed in the director and then use a common convention when talking to the SD.

arogge (commented):
check-sources said: Plugin 'fix copyright year' would modify 'core/src/dird/ua_server.h'
I'll run the tests now.

core/src/lib/bool_string.h
```diff
@@ -0,0 +1,57 @@
...
+#ifndef BAREOS_LIB_BOOL_STRING_H_
+#define BAREOS_LIB_BOOL_STRING_H_
+
+#include <set>
+#include <stdexcept>
+#include <string>
+
+class BoolString {
+ public:
+  BoolString(const std::string& s) : str_value(s)
```
arogge, line 31: single-argument constructor should probably be marked explicit

core/src/dird/ua_cmds.cc
```diff
@@ -1590,6 +1594,111 @@ static bool SetdebugCmd(UaContext* ua, const char* cmd)
...
+
+  if (!ConnectToStorageDaemon(ua->jcr, 1, 15, false)) {
+    ua->ErrorMsg(_("Failed to connect to Storage daemon.\n"));
+    return false;
+  }
+
+  Dmsg0(120, _("Connected to storage daemon\n"));
+  ua->jcr->store_bsock->fsend("setdevice device=%s autoselect=%s",
+                              arguments.at("device").c_str(),
+                              arguments.at("autoselect").c_str());
```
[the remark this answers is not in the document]
franku: const ArgumentsList& allows only at()

core/src/lib/bool_string.h
franku: Now used member class initialization.

PR #689, opened 2020-12-11
tests: fix typos in cram_md5

pstorz:


core/src/tests/cram_md5.cc
```diff
@@ -163,7 +166,7 @@ TEST(cram_md5, different_stored_qualified_name)
   EXPECT_EQ(cs.client_cram.result, CramMd5Handshake::HandshakeResult::SUCCESS)
       << "  CramMd5Handshake::HandshakeResult::"
       << cram_result_to_string.at(cs.client_cram.result) << std::endl
-      << "  Wich is probably a side client_cram of a challange attack."
+      << "  Which is probably a side client_cram of a challenge attack."
```
arogge, line 169: what's a "side client_cram"?
And isn't it a relay attack that's being mitigated?

franku, line 169:
```suggestion
      << "  Which is probably a side effect of a challenge attack."
```

franku (approved):

PR #691, opened 2020-12-13
Droplet as subtree - full history

franku:


arogge:
We decided to use the squashed variant. The original history is still available in the upstream repository.

PR #696, opened 2020-12-21
Remove unneeded X file rights

tuxmaster5000:
During the rpm build process error messages about executable files files without shebang occurs.


joergsteffens:
Your change is correct and definitively an improvement. While testing it, I found some more files with unneeded executable permission and added them. This also have influence on #695.
I'll run it through our test suite.

joergsteffens (approved):

joergsteffens:
Successfully passes the test suite.

PR #698, opened 2020-12-24
Install python3 bin file to the correct diretory.

tuxmaster5000:
The python binary files are installed to the arch independence place instant of the arch depend.
See:
https://cmake.org/cmake/help/latest/module/FindPython3.html
Python3_SITEARCH will be the correct macro for the .so stuff.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@ and since Bareos version 20 this project adheres to [Semantic Versioning](https:
 - dird: console: add the ability to rerun multiple commas separated jobids [PR #1170]
 
 ### Fixed
+- python plugins: store architecture specific modules in sitearch (instead of sitelib) [PR #698]
 - debian: fix package dependencies for webui and Ceph [PR #1183]
 - Python plugins: fix handling of additional pluginoptions parameter [PR #1177]
 - NDMP_BAREOS: support autoxflate plugin [PR #1013]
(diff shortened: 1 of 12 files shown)


procule:
A way to specify the proper Python3 version with a cmake variable would be nice (I hate cmake...)
The cmake FindPython3 file will usually try to find the latest version installed which is not always the version you want.
The way your cmake configuration files are made prevents to specify the Python environment to use since it only relies on the awful FindPython3 file. 

A simple way would be something like that:

```
--- ../core/cmake/BareosFindAllLibraries.cmake	2021-11-25 22:43:22.601594339 -0500
+++ BareosFindAllLibraries.cmake	2021-11-25 22:42:51.329218300 -0500
@@ -38,7 +38,12 @@
 
 else()
   find_package(Python2 COMPONENTS Interpreter Development)
-  find_package(Python3 COMPONENTS Interpreter Development)
+
+  if (${USE_PYTHON3_VERSION})
+    find_package(Python3 ${USE_PYTHON3_VERSION} EXACT COMPONENTS Interpreter Development)
+  else()
+    find_package(Python3 COMPONENTS Interpreter Development)
+  endif()
 
   set(Python3_VERSION_MAJOR
       ${Python3_VERSION_MAJOR}
```

joergsteffens:
Rebased the branch on master, fix RPM building and extend also for Python2.
Automatic testing went through and also manual testing both with python2 and python3 on CentOS 7 have been successful.

joergsteffens (approved):

PR #699, opened 2020-12-25
Add options to use the vddk from antoher path

tuxmaster5000:
Add the cmake options vddk_incl and vddk_lib to make the path to the vmware vddk more flexible.


franku:
Please use cmake list instead of manual string manipulations (`git am` the attached file for a suggestion);

[0001-use-a-cmake-list-instead-of-string.patch.log](https://github.com/bareos/bareos/files/5938703/0001-use-a-cmake-list-instead-of-string.patch.log)

franku:
Closed due to no activity.

PR #700, opened 2020-12-26
On modern systemd the network is only 'ip' ready at network-online.target

tuxmaster5000:
Using only network,target, can result in ip unready network state.


joergsteffens (approved):
Looks good. The `network-online.target` is available since 2013, Systemd V201.

joergsteffens:
Adding some cleanup and rebased on master.

PR #701, opened 2021-01-04
systemtests: check if cmake-configured hostname is same

franku:
Pre-check, when running ctest, that the configured hostname matches the current hostname. Otherwise tests will fail right after start because bconsole cannot connect to director.


joergsteffens:
For me, this change would not work, as I had to run `cmake -Dhostname=127.0.0.1` to circumvent problems with IPv6 name resolution.

franku:
Tried to circumvent the "-Dhostname  cmake option" with an extra flag.

joergsteffens (approved):
Looks good. I've tested it with different parameters and it worked successfully. Only thing to observe is that after specifying the hostname `cmake -Dhostname=...`, and before reconfiguring with `-Dhostname`, the cache file CMakeCache.txt must be deleted. Otherwise it still uses the old value. I've added a hint to the output of the check.
As my changes are only cosmetic, feel free to discard, squash or use them as you like.

PR #702, opened 2021-01-06
libcloud plugin invalid filedescriptor

franku:
This PR should fix the invalid file descriptor issue in the libcloud plugin. Remark, there could be information about some other issue that will be lost when written to /dev/null. 

Therefore we should decide if /dev/null is right or if we want to redirect to a regular file. The reason I ask is, the invalid filedescriptor error occurs randomly and only if there was a previous write to stdout just before the new process starts. This text could containt more information about any other issue we do not know yet.

### DIFF
--- a/core/src/plugins/filed/python/libcloud/BareosLibcloudApi.py
+++ b/core/src/plugins/filed/python/libcloud/BareosLibcloudApi.py
@@ -1,6 +1,6 @@
 #   BAREOS - Backup Archiving REcovery Open Sourced
 #
-#   Copyright (C) 2020-2020 Bareos GmbH & Co. KG
+#   Copyright (C) 2020-2021 Bareos GmbH & Co. KG
 #
 #   This program is Free Software; you can redistribute it and/or
 #   modify it under the terms of version three of the GNU Affero General Public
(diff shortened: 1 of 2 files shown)


core/src/plugins/filed/python/libcloud/BareosLibcloudApi.py
```diff
@@ -88,6 +97,12 @@ def __init__(self, options, last_run, tmp_dir_path):
             for i in range(self.number_of_worker)
         ]
 
+        try:
+            BareosLibcloudApi.redirect_stdout_stderr()
+        except Exception as e:
+            jobmessage(M_ERROR, "Could not redirect stdout: %s" % (str(e)))
```
arogge, line 103:
```suggestion
            jobmessage(M_ERROR, "Could not redirect stdout and/or stderr: %s" % (str(e)))
```

core/src/plugins/filed/python/libcloud/BareosLibcloudApi.py
```diff
@@ -50,6 +50,15 @@ def probe_driver(options):
...
         return "failed"
 
+    @staticmethod
+    def redirect_stdout_stderr():
+        if sys.stdout != None:
+            sys.stdout.close()
+        if sys.stderr != None:
+            sys.stderr.close()
+        sys.stdout = open("/dev/null", "w")
+        sys.stderr = open("/dev/null", "w")
```
arogge, line 60: you could have reused the filehandle you opened for sys.stdout and use it for sys.stderr, too.
However, two filedescriptors won't hurt either.

arogge (approved):
Looks good to me.

PR #705, opened 2021-01-08
docs: make bareos-20 the default (backport bareos-20)

joergsteffens:
While this change is not required for https://docs.bareos.org, it is required for the test documentation in https://download.bareos.com/bareos/testing/CD/bareos-20/BareosMainReference/


joergsteffens:
@arogge alternative approaches would be to change the way the documentation is created.Maybe you have additional thoughts about this.

PR #707, opened 2021-01-08
User date time formats

alaaeddineelamri:
add support for shorter date formats, where shorter dates are compensated with lowest value possible to make a full date
also fix a bug where Bareos accepted bogus datetime formats.

### DIFF
--- a/core/src/dird/ua_restore.cc
+++ b/core/src/dird/ua_restore.cc
@@ -3,7 +3,7 @@
 
    Copyright (C) 2002-2011 Free Software Foundation Europe e.V.
    Copyright (C) 2011-2016 Planets Communications B.V.
-   Copyright (C) 2013-2020 Bareos GmbH & Co. KG
+   Copyright (C) 2013-2021 Bareos GmbH & Co. KG
 
    This program is Free Software; you can redistribute it and/or
    modify it under the terms of version three of the GNU Affero General Public
@@ -42,6 +42,7 @@
 #include "dird/ua_select.h"
 #include "dird/ua_tree.h"
 #include "dird/ua_run.h"
+#include "dird/ua_restore.h"
 #include "dird/bsr.h"
 #include "lib/breg.h"
 #include "lib/edit.h"
@@ -560,14 +561,19 @@ static int UserSelectJobidsOrFiles(UaContext* ua, RestoreContext* rx)
         have_date = true;
         break;
       case 2: /* before */
+      {
         if (have_date || !HasValue(ua, i)) { return 0; }
-        if (StrToUtime(ua->argv[i]) == 0) {
-          ua->ErrorMsg(_("Improper date format: %s\n"), ua->argv[i]);
+
+        std::string cpdate = CompensateShortDate(ua->argv[i]);
+
+        if (StrToUtime(cpdate.c_str()) == 0) {
+          ua->ErrorMsg(_("Improper date format: %s\n"), cpdate.c_str());
           return 0;
         }
-        bstrncpy(date, ua->argv[i], sizeof(date));
+        bstrncpy(date, cpdate.c_str(), sizeof(date));
         have_date = true;
         break;
+      }
       case 3: /* file */
       case 4: /* dir */
         if (!HasValue(ua, i)) { return 0; }
@@ -839,13 +845,63 @@ static bool get_date(UaContext* ua, char* date, int date_len)
   ua->SendMsg(
       _("The restored files will the most current backup\n"
         "BEFORE the date you specify below.\n\n"));
+  std::string cmpdate;
   for (;;) {
     if (!GetCmd(ua, _("Enter date as YYYY-MM-DD HH:MM:SS :"))) { return false; }
-    if (StrToUtime(ua->cmd) != 0) { break; }
+    cmpdate = CompensateShortDate(ua->cmd);
+
+    if (StrToUtime(cmpdate.c_str()) != 0) { break; }
     ua->ErrorMsg(_("Improper date format.\n"));
   }
-  bstrncpy(date, ua->cmd, date_len);
+  bstrncpy(date, cmpdate.c_str(), date_len);
   return true;
+}
+
+/**
+ * Compensate missing date-time elements in shorter formats
+ * Returns a full compensated date when argument has a correct short format
+ * Returns the argument unchanged, if there is a format problem
+ */
+std::string CompensateShortDate(const char* cmd)
+{
+  tm datetime{};
+  char trailinggarbage[16]{""};
+
+  if ((sscanf(cmd, "%u-%u-%u %u:%u:%u%15s", &datetime.tm_year, &datetime.tm_mon,
+              &datetime.tm_mday, &datetime.tm_hour, &datetime.tm_min,
+              &datetime.tm_sec, trailinggarbage)
+           == 7
+       || sscanf(cmd, "%u-%u-%u %u:%u%15s", &datetime.tm_year, &datetime.tm_mon,
+                 &datetime.tm_mday, &datetime.tm_hour, &datetime.tm_min,
+                 trailinggarbage)
+              == 6
+       || sscanf(cmd, "%u-%u-%u %u%15s", &datetime.tm_year, &datetime.tm_mon,
+                 &datetime.tm_mday, &datetime.tm_hour, trailinggarbage)
+              == 5
+       || sscanf(cmd, "%u-%u-%u%15s", &datetime.tm_year, &datetime.tm_mon,
+                 &datetime.tm_mday, trailinggarbage)
+              == 4
+       || sscanf(cmd, "%u-%u%15s", &datetime.tm_year, &datetime.tm_mon,
+                 trailinggarbage)
+              == 3
+       || sscanf(cmd, "%u%s", &datetime.tm_year, trailinggarbage) == 2)
+      && (trailinggarbage[0] == '\0')) {
+    if (datetime.tm_mon == 0) { datetime.tm_mon = 1; }
+    if (datetime.tm_mday == 0) { datetime.tm_mday = 1; }
+
+    if (datetime.tm_year >= 1900) {
+      datetime.tm_year -= 1900;
+    } else {
+      return cmd;
+    }
+
+    if (datetime.tm_mon > 0) { datetime.tm_mon--; }
+
+    char buffer[MAX_TIME_LENGTH];
+    std::strftime(buffer, MAX_TIME_LENGTH, "%Y-%m-%d %H:%M:%S", &datetime);
+    return buffer;
+  }
+  return cmd;
 }
 
 /**
--- a/core/src/dird/ua_restore.h
+++ b/core/src/dird/ua_restore.h
@@ -1,7 +1,7 @@
 /*
    BAREOS® - Backup Archiving REcovery Open Sourced
 
-   Copyright (C) 2018-2018 Bareos GmbH & Co. KG
+   Copyright (C) 2018-2021 Bareos GmbH & Co. KG
 
    This program is Free Software; you can redistribute it and/or
    modify it under the terms of version three of the GNU Affero General Public
@@ -23,6 +23,8 @@
 #define BAREOS_DIRD_UA_RESTORE_H_
 
 namespace directordaemon {
+
+std::string CompensateShortDate(const char* cmd);
 
 void FindStorageResource(UaContext* ua,
                          RestoreContext& rx,
--- a/core/src/lib/btime.cc
+++ b/core/src/lib/btime.cc
@@ -2,7 +2,7 @@
    BAREOS® - Backup Archiving REcovery Open Sourced
 
    Copyright (C) 2000-2011 Free Software Foundation Europe e.V.
-   Copyright (C) 2015-2019 Bareos GmbH & Co. KG
+   Copyright (C) 2015-2021 Bareos GmbH & Co. KG
 
    This program is Free Software; you can redistribute it and/or
    modify it under the terms of version three of the GNU Affero General Public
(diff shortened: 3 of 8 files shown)


core/src/dird/ua_restore.h
```diff
@@ -22,8 +22,12 @@
 #ifndef BAREOS_DIRD_UA_RESTORE_H_
 #define BAREOS_DIRD_UA_RESTORE_H_
 
+#include <sstream>
```
franku, line 25:
```suggestion
```

core/src/lib/btime.cc
```diff
@@ -139,23 +139,72 @@ utime_t StrToUtime(const char* str)
...
   time = mktime(&tm);
   if (time == -1) { time = 0; }
   return (utime_t)time;
 }
 
+/*
+ * Check if a given date is valid
+ */
+
+bool DateIsValid(tm tm)
```
franku, line 168:
```suggestion
static bool DateIsValid(tm tm)
```

franku: same as above

core/src/lib/btime.h
```diff
@@ -39,7 +39,7 @@ char* bstrftime_ny(char* dt, int maxlen, utime_t tim);
 char* bstrftime_nc(char* dt, int maxlen, utime_t tim);
 char* bstrftime_wd(char* dt, int maxlen, utime_t tim);
 utime_t StrToUtime(const char* str);
-
+bool DateIsValid(tm tm);
```
franku, line 42:
```suggestion
```

core/src/lib/btime.cc
franku: As long as it is not used outside of btime.cc this function should be static.

core/src/lib/btime.cc
```diff
@@ -139,23 +139,72 @@ utime_t StrToUtime(const char* str)
...
+{
+  if (tm.tm_mon > 12 || tm.tm_hour > 23 || tm.tm_min > 59 || tm.tm_sec > 59) {
+    return 0;
```
franku, line 171:
```suggestion
    return false;
```

franku: Should use a bool value below, too.

core/src/lib/btime.h
franku: .. see below

core/src/dird/ua_restore.cc
```diff
@@ -839,15 +848,63 @@ static bool get_date(UaContext* ua, char* date, int date_len)
...
 
+/**
+ * Compensate missing date-time elements in shorter formats
+ */
+
+std::string CompensateShortDate(const char* cmd)
+{
+  std::stringstream ss;
+  ss << cmd;
+  std::string time, stime;
```
franku, line 872:
```suggestion
  std::string time;
```

core/src/dird/ua_restore.cc
```diff
@@ -839,15 +848,63 @@ static bool get_date(UaContext* ua, char* date, int date_len)
...
+  getline(ss, date, ' ');
+  getline(ss, time);
+
+  ss = std::stringstream();
+  ss << date;
+  while (std::getline(ss, sdate, '-')) { date_length++; }
+
+  ss = std::stringstream();
+  ss << time;
+  while (std::getline(ss, stime, ':')) { date_length++; }
```
franku, line 885:
```suggestion
  std::string stime;
  while (std::getline(ss, stime, ':')) { date_length++; }
```

core/src/dird/ua_restore.cc
franku: Pls declare variables each line separately and just as close as where they are used.

core/src/dird/ua_restore.cc
```diff
@@ -839,15 +848,63 @@ static bool get_date(UaContext* ua, char* date, int date_len)
...
+/**
+ * Compensate missing date-time elements in shorter formats
+ */
+
+std::string CompensateShortDate(const char* cmd)
+{
+  std::stringstream ss;
+  ss << cmd;
+  std::string time, stime;
+  std::string date, sdate;
```
franku, line 873:
```suggestion
  std::string date;
```

core/src/dird/ua_restore.cc
```diff
@@ -839,15 +848,63 @@ static bool get_date(UaContext* ua, char* date, int date_len)
...
+  int date_length = 0;
+
+  getline(ss, date, ' ');
+  getline(ss, time);
+
+  ss = std::stringstream();
+  ss << date;
+  while (std::getline(ss, sdate, '-')) { date_length++; }
```
franku, line 881:
```suggestion
  std::string sdate;
  while (std::getline(ss, sdate, '-')) { date_length++; }
```

franku (commented):
Added some remarks.

core/src/dird/ua_restore.cc
```diff
@@ -839,15 +845,67 @@ static bool get_date(UaContext* ua, char* date, int date_len)
...
+ * Returns a full compensated date when argument has a correct short format
+ * Returns the argument unchanged, if there is a format problem
+ */
+
+std::string CompensateShortDate(const char* cmd)
+{
+  struct tm tm;
+  char* garbage = nullptr;
+
+  memset(&tm, -1, sizeof(tm));
```
arogge, line 871: This exploits the fact that `-1` is represented as `~0` and thus you will get sane values when filling a struct of integers with bytes.
While this should work on any platform, two's complement is not guaranteed by the C++ standard, which in turn means this may be undefined behaviour and the compiler is allowed to break your implementation anytime.

However, you could simply initialize with zero and even simplify the below code.

core/src/dird/ua_restore.cc
```diff
@@ -839,15 +845,67 @@ static bool get_date(UaContext* ua, char* date, int date_len)
...
+                 &garbage)
+              == 4
+       || sscanf(cmd, "%u-%u%s", &tm.tm_year, &tm.tm_mon, &garbage) == 3
+       || sscanf(cmd, "%u%s", &tm.tm_year, &garbage) == 2)
+      && (garbage == nullptr)) {
+    if (tm.tm_mon == -1) { tm.tm_mon = 1; }
+    if (tm.tm_mday == -1 || tm.tm_mday == 0) { tm.tm_mday = 1; }
+    if (tm.tm_hour == -1) { tm.tm_hour = 0; }
+    if (tm.tm_min == -1) { tm.tm_min = 0; }
+    if (tm.tm_sec == -1) { tm.tm_sec = 0; }
```
arogge, line 891: if you zero-initialized `tm`, then this would simplify to:
```suggestion
    if (tm.tm_mon == 0) { tm.tm_mon = 1; }
    if (tm.tm_mday == 0) { tm.tm_mday = 1; }
```

core/src/dird/ua_restore.cc
```diff
@@ -839,15 +845,67 @@ static bool get_date(UaContext* ua, char* date, int date_len)
...
+    } else {
+      return cmd;
+    }
+
+    if (tm.tm_mon > 0) { tm.tm_mon--; }
+
+    char buffer[MAX_TIME_LENGTH];
+    std::strftime(buffer, MAX_TIME_LENGTH, "%Y-%m-%d %H:%M:%S", &tm);
+    std::string date(buffer);
+    return date;
```
arogge, line 904: if you don't bind the string to a variable (i.e. make it an lvalue), it will be an rvalue which may help the compiler to produce better code.
It should also be possible to just return `buffer` and let the compiler handle the conversion.
```suggestion
    return buffer;
```

arogge (commented):
Thank you very much! This is amazing work.
Great implementation, nicely tested and well documented.
You should definitely add your name to the AUTHORS file!

I have some remarks and suggestions. Most of them are not too important, feel free to discuss.

core/src/lib/btime.cc
```diff
@@ -129,26 +177,35 @@ utime_t StrToUtime(const char* str)
...
    */
   if (!str || *str == 0) { return 0; }
 
-  if (sscanf(str, "%d-%d-%d %d:%d:%d", &tm.tm_year, &tm.tm_mon, &tm.tm_mday,
-             &tm.tm_hour, &tm.tm_min, &tm.tm_sec)
-      != 6) {
+  if ((sscanf(str, "%d-%d-%d %d:%d:%d%s", &tm.tm_year, &tm.tm_mon, &tm.tm_mday,
+              &tm.tm_hour, &tm.tm_min, &tm.tm_sec, &garbage)
+       != 7)
+      || garbage != nullptr) {
```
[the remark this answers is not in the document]
alaaeddineelamri: yes exactly. I remarked that if the date is followed with any kind of garbage, it creates an error later on in the process that calls `StrToUtime()`

For example, in the restore command, `StrToUtime(ua->cmd)` is used to check if  the date inside `ua->cmd` is valid. If so, the whole `ua->cmd` is copied into another variable which in turn will be passed for the next step.

In this case, let's take `ua->cmd` is `2020-01-22 12:00:00 cpp`. Due to the nature of `sscanf`, the first part of `ua->cmd` containing the correct date will be scanned and deemed a correct date even though there is a trailing string at the end. In the next step, since `ua->cmd` is deemed correct, all of it is copied to a variable including the trailing string ` cpp`. And that creates errors in other places that are waiting for an actual full date. 

And so now, the trailing string is caught at the level of `StrToUtime()` instead of later stages.

core/src/dird/ua_restore.cc
```diff
@@ -839,15 +845,65 @@ static bool get_date(UaContext* ua, char* date, int date_len)
...
+ */
+std::string CompensateShortDate(const char* cmd)
+{
+  tm datetime{};
+  char* trailinggarbage = nullptr;
+
+  if ((sscanf(cmd, "%u-%u-%u %u:%u:%u%s", &datetime.tm_year, &datetime.tm_mon,
+              &datetime.tm_mday, &datetime.tm_hour, &datetime.tm_min,
+              &datetime.tm_sec, &trailinggarbage)
+           == 7
```
arogge, line 873: the use of `trailinggarbage` will not work correctly. I'll provide some sample-code to show the issue.

core/src/dird/ua_restore.cc
```diff
@@ -839,15 +845,65 @@ static bool get_date(UaContext* ua, char* date, int date_len)
...
+                 &datetime.tm_mday, &datetime.tm_hour, &trailinggarbage)
+              == 5
+       || sscanf(cmd, "%u-%u-%u%15s", &datetime.tm_year, &datetime.tm_mon,
+                 &datetime.tm_mday, &trailinggarbage)
+              == 4
+       || sscanf(cmd, "%u-%u%15s", &datetime.tm_year, &datetime.tm_mon,
+                 &trailinggarbage)
+              == 3
+       || sscanf(cmd, "%u%s", &datetime.tm_year, &trailinggarbage) == 2)
+      && (trailinggarbage[0] == '\0')) {
```
arogge, line 888: For strings, you need to pass the array to `sscanf()` and not the address of the array.
```suggestion
  if ((sscanf(cmd, "%u-%u-%u %u:%u:%u%15s", &datetime.tm_year, &datetime.tm_mon,
              &datetime.tm_mday, &datetime.tm_hour, &datetime.tm_min,
              &datetime.tm_sec, trailinggarbage)
           == 7
       || sscanf(cmd, "%u-%u-%u %u:%u%15s", &datetime.tm_year, &datetime.tm_mon,
                 &datetime.tm_mday, &datetime.tm_hour, &datetime.tm_min,
                 trailinggarbage)
              == 6
       || sscanf(cmd, "%u-%u-%u %u%15s", &datetime.tm_year, &datetime.tm_mon,
                 &datetime.tm_mday, &datetime.tm_hour, trailinggarbage)
              == 5
       || sscanf(cmd, "%u-%u-%u%15s", &datetime.tm_year, &datetime.tm_mon,
                 &datetime.tm_mday, trailinggarbage)
              == 4
       || sscanf(cmd, "%u-%u%15s", &datetime.tm_year, &datetime.tm_mon,
                 trailinggarbage)
              == 3
       || sscanf(cmd, "%u%s", &datetime.tm_year, trailinggarbage) == 2)
      && (trailinggarbage[0] == '\0')) {
```

core/src/lib/btime.cc
```diff
@@ -121,37 +121,96 @@ char* bstrutime(char* dt, int maxlen, utime_t utime)
...
    * Check for bad argument
    */
   if (!str || *str == 0) { return 0; }
 
-  if (sscanf(str, "%d-%d-%d %d:%d:%d", &tm.tm_year, &tm.tm_mon, &tm.tm_mday,
-             &tm.tm_hour, &tm.tm_min, &tm.tm_sec)
-      != 6) {
+  if ((sscanf(str, "%u-%u-%u %u:%u:%u%15s", &datetime.tm_year, &datetime.tm_mon,
+              &datetime.tm_mday, &datetime.tm_hour, &datetime.tm_min,
+              &datetime.tm_sec, &trailinggarbage)
```
arogge, line 191: same here: pass the array-pointer instead of the array-pointer's address.
```suggestion
              &datetime.tm_sec, trailinggarbage)
```

arogge (changes requested):
I found a small problem that's left, otherwise it is really fine now.

core/src/dird/ua_restore.cc
```diff
@@ -839,15 +845,65 @@ static bool get_date(UaContext* ua, char* date, int date_len)
...
   for (;;) {
     if (!GetCmd(ua, _("Enter date as YYYY-MM-DD HH:MM:SS :"))) { return false; }
-    if (StrToUtime(ua->cmd) != 0) { break; }
+    cmpdate = CompensateShortDate(ua->cmd);
+
+    if (StrToUtime(cmpdate.c_str()) != 0) { break; }
     ua->ErrorMsg(_("Improper date format.\n"));
   }
-  bstrncpy(date, ua->cmd, date_len);
+  bstrncpy(date, cmpdate.c_str(), date_len);
```
franku, line 856: What happens with date after copying to it?

core/src/lib/btime.cc
```diff
@@ -121,37 +121,96 @@ char* bstrutime(char* dt, int maxlen, utime_t utime)
...
+  if (datetime.tm_year <= 0 || datetime.tm_mon <= 0 || datetime.tm_mday <= 0
+      || datetime.tm_mon > 12 || datetime.tm_hour > 23 || datetime.tm_min > 59
+      || datetime.tm_sec > 59 || datetime.tm_mon < 0 || datetime.tm_hour < 0
+      || datetime.tm_min < 0 || datetime.tm_sec < 0) {
+    return false;
+  }
+
+  /*
+   * taking into context leap years in february:
+   * leap years are years that are devisible by 4 or are divisible by 4, 100 and
```
franku, line 140:
```suggestion
   * leap years are years that are divisible by 4 or are divisible by 4, 100 and
```

core/src/dird/ua_restore.cc
alaaeddineelamri: It is used within `UserSelectJobidsOrFiles(UaContext* ua, RestoreContext* rx)`.

After calling `get_date(UaContext* ua, char* date, int date_len)`, the result date is used with other functions that expect a date to fetch the actual jobs or data needed before that date, like `SelectBackupsBeforeDate(UaContext* ua, RestoreContext* rx, char* date)` for example.

arogge (approved):

PR #708, opened 2021-01-08
core: adapt debian/control to new version of debian/control.in

joergsteffens:
debian/control is generated from debian/control.in during cmake.
This change (adding 2 blank lines) prevents that this file differs from git after every run of cmake.
This change is required, as the optional package bareos-storage-droplet has been added recently.


franku:
Double checked on Fedora 33

PR #710, opened 2021-01-11
UCS: adapt download URL to new format (bareos >= 20)

joergsteffens:


joergsteffens:
Tested the change manually on an fresh UCS 4.4 system. As @arogge already agrees to this chance, I'll merge it.

PR #712, opened 2021-01-14
travis: use dist=bionic as default

joergsteffens:
In the current travis image, the default Postgresql database does not work.
Default is Postgresql 12.
However, Postgresql 13 is also included and does work.
This problem exists since 03 Nov 2020.
As it is still not fixed, we switch to the older bionic dist.
We still keep one job running on focal, however with Postgresql 13.


pstorz (approved):
Looks good to me

PR #715, opened 2021-01-15
run systemtests also on freebsd and solaris during build

pstorz:


core/src/tests/CMakeLists.txt
```diff
@@ -370,7 +370,10 @@ if(HAVE_EXECINFO_H
...
 endif() # HAVE_EXECINFO_H ..
 
-if(GMOCK_FOUND AND NOT client-only)
+if(GMOCK_FOUND
+   AND NOT client-only
+   AND NOT ${CMAKE_SYSTEM_NAME} MATCHES "SunOS"
+)
   bareos_add_test(
     sd_reservation LINK_LIBRARIES ${LINK_LIBRARIES} ${GMOCK_LIBRARIES}
   )
```
arogge, line 379: * shouldn't we add the test as disabled, if it cannot be run?
* maybe `bareos_add_test()` should have a parameter `DISABLE_IF` that takes a condition as an argument and disables the test if the condition evaluates to true. I think it would simplify a lot of the CMake-Code around the systemtests.

systemtests/scripts/functions
```diff
@@ -113,18 +113,24 @@ setup_data()
 {
   data_dir="${tmp}/data"
 
+  # gnu tar is required
+  if [ "$(uname -s)" == SunOS ]; then
+    GTAR='gtar'
+  else
+    GTAR='tar'
+  fi
+
```
arogge, line 122: maybe we could do this with `command -v gtar tar` to always use `gtar` if it exists or `tar` otherwise.

We should either have should have a find_command function that wraps this. As an alternative we could export the commands we found using cmake into a shellscript that could be sourced. That way we could have `$GTAR_EXECUTABLE` or something like that automatically set by cmake.

PR #717, opened 2021-01-21
dird: Fix Copy Job behaviour regarding Archive Jobs

fbergkemper:


systemtests/scripts/functions
```diff
@@ -301,6 +301,19 @@ check_files_written()
     fi
 }
 
+check_sd_files_written()
+{
+    LOG=$1
+    FILES="$(awk '/SD Files Written:/ { last=$4 } END { print last }' "$LOG")"
+
+    if [[ "$FILES" -lt 1 || -z "$FILES" ]]; then
+        print_debug "ERROR: Expected SD files written > 0, got $FILES."
```
franku, line 310:
```suggestion
        print_debug "ERROR: Expected number of SD files written > 0, got \"$FILES\"."
```

franku: In case $FILES is empty this makes more sense.

franku (approved):
One remark on a debug message, otherwise looks good. 
The test runs successful on my Fedora 33. 
Good work!

systemtests/scripts/functions
fbergkemper: Good suggestions, made the change. Thanks for the review.

PR #726, opened 2021-02-08
Revert "webui: change updateBvfsCache trigger condition in restore controller"

fbergkemper:
This reverts commit 9967eb01f7297735d2e838dce6569ce8273a0c5c.

### DIFF
--- a/webui/module/Restore/src/Restore/Controller/RestoreController.php
+++ b/webui/module/Restore/src/Restore/Controller/RestoreController.php
@@ -550,9 +550,10 @@ class RestoreController extends AbstractActionController
     $this->setRestoreParams();
     $this->getJobIds();
 
-    if($this->restore_params['id'] !== null && $this->restore_params['id'] !== "#") {
+    if($this->restore_params['id'] === null || $this->restore_params['id'] === "#") {
       $this->updateBvfsCache();
     }
+
     $this->getDirectories();
     $this->getFiles();
 


webui/module/Restore/src/Restore/Controller/RestoreController.php
```diff
@@ -550,9 +550,10 @@ private function buildSubtree()
     $this->setRestoreParams();
     $this->getJobIds();
 
-    if($this->restore_params['id'] !== null && $this->restore_params['id'] !== "#") {
+    if($this->restore_params['id'] == null || $this->restore_params['id'] == "#") {
```
arogge, line 553: you cannot sanely compare with null in PHP. You'll have to use `is_null()` to check if a value is null.
Question is: do you mean `is_null($this->restore_params['id'])`, `array_key_exists('id', $this->restore_params)` or maybe even both?

fbergkemper: $this->restore_params['id'] is always set and its default value is null, there is no need to check if it exists.
Values are mixed type, integer, string or null.

is_null($a) is same as $a === null whereas $a === null is bit faster than is_null($a) for saving one function call, but I don't want to argue about micro optimization here.

I've corrected the comparison operators to take the types into account.

franku (approved):
Good job!

PR #729, opened 2021-02-11
python plugins: fix problems with parallel jobs

pstorz:
Fix problems with parallel jobs using python plugins


sduehr (changes requested):
- The PR description text is still the template text
- The first line of commit message from 9b40aae is too long

sduehr (approved):
Looks good now, can be merged.

PR #731, opened 2021-02-15
Multicolumn prompts in selection mode

alaaeddineelamri:


core/src/dird/ua_select.cc
```diff
@@ -1107,6 +1107,56 @@ void AddPrompt(UaContext* ua, std::string&& prompt)
...
+      max_prompt_length = strlen(ua->prompt[i]);
+    }
+  }
+  max_line_length = max_prompt_length + prompt_max_index_length;
+  prompts_perline = (windowsize / max_line_length) - 1;
+  if (prompts_perline == 0) { prompts_perline = 1; }
+
+
+  max_line_length += 4;
+  char buffer[max_line_length];
```
franku, line 1136: pls use a std::vector<char> instead

core/src/dird/ua_select.cc
```diff
@@ -1107,6 +1107,56 @@ void AddPrompt(UaContext* ua, std::string&& prompt)
...
+
+  std::string output("");
```
franku, line 1138:
```suggestion
  std::string output;
```

core/src/tests/multicolumn_prompts.cc
```diff
@@ -0,0 +1,187 @@
...
+{
+  UaContext uacon;
+  UaContext* ua = &uacon;
+
+  const char* list[] = {_(""), _(" "), _("  "), NULL};
+
+  StartPrompt(ua, "start");
+  for (int i = 0; list[i]; ++i) { AddPrompt(ua, list[i]); }
+
+  std::string output("");
```
franku, line 180:
```suggestion
  std::string output;
```

core/src/tests/multicolumn_prompts.cc
```diff
@@ -0,0 +1,187 @@
...
+         _("Select backup for a client before a specified time"),
+         _("Enter a list of files to restore"),
+         _("Enter a list of files to restore before a specified time"),
+         _("Find the JobIds of the most recent backup for a client"),
+         _("Find the JobIds for a backup for a client before a specified time"),
+         _("Enter a list of directories to restore for found JobIds"),
+         _("Select full restore to a specified Job date"),
+         _("Cancel"),
+         NULL};
+  StartPrompt(ua, "start");
```
franku, line 60:
```suggestion
  StartPrompt(&uacon, "start"); // you can eliminate the ua pointer everywhere like this
```

core/src/tests/multicolumn_prompts.cc
```diff
@@ -0,0 +1,187 @@
...
+  for (int i = 0; list[i]; ++i) { AddPrompt(ua, list[i]); }
+
+  std::string output("");
```
franku, line 63:
```suggestion
  std::string output;
```

core/src/tests/multicolumn_prompts.cc
```diff
@@ -0,0 +1,187 @@
...
```
franku, line 106:
```suggestion
  std::string output;
```

core/src/tests/multicolumn_prompts.cc
```diff
@@ -0,0 +1,187 @@
...
+  output = FormatMulticolumnPrompts(ua, 80, 20);
```
franku, line 64:
```suggestion
  std::string output = FormatMulticolumnPrompts(ua, 80, 20);
```

core/src/tests/multicolumn_prompts.cc
```diff
@@ -0,0 +1,187 @@
...
+         _("Find the JobIds for a backup for a client before a specified time"),
+         _("Enter a list of directories to restore for found JobIds"),
+         _("Select full restore to a specified Job date"),
+         _("Cancel"),
+         NULL};
+  StartPrompt(ua, "start");
+  for (int i = 0; list[i]; ++i) { AddPrompt(ua, list[i]); }
+
+  std::string output("");
+  output = FormatMulticolumnPrompts(ua, 5000, 1);
```
franku, line 107:
```suggestion
  std::string output = FormatMulticolumnPrompts(ua, 5000, 1);
```

core/src/tests/multicolumn_prompts.cc
```diff
@@ -0,0 +1,187 @@
...
+      _("bareos1"),  _("bareos2"),  _("bareos3"),  _("bareos4"),  _("bareos5"),
+      _("bareos6"),  _("bareos7"),  _("bareos8"),  _("bareos9"),  _("bareos10"),
+      _("bareos11"), _("bareos12"), _("bareos13"), _("bareos14"), _("bareos15"),
+
+      NULL};
+
+  StartPrompt(ua, "start");
+  for (int i = 0; list[i]; ++i) { AddPrompt(ua, list[i]); }
+
+  std::string output("");
```
franku, line 141:
```suggestion
```

core/src/tests/multicolumn_prompts.cc
```diff
@@ -0,0 +1,187 @@
...
+  output = FormatMulticolumnPrompts(ua, 60, 10);
```
franku, line 142:
```suggestion
  std::string output = FormatMulticolumnPrompts(ua, 60, 10);
```

core/src/tests/multicolumn_prompts.cc
```diff
@@ -0,0 +1,187 @@
...
+         _("Select the most recent backup for a client"),
+         _("Select backup for a client before a specified time"),
+         _("Enter a list of files to restore"),
+         _("Enter a list of files to restore before a specified time"),
+         _("Find the JobIds of the most recent backup for a client"),
+         _("Find the JobIds for a backup for a client before a specified time"),
+         _("Enter a list of directories to restore for found JobIds"),
+         _("Select full restore to a specified Job date"),
+         _("Cancel"),
+         NULL};
```
franku, line 102:
```suggestion
         nullptr}; // do not use NULL anymore
```

core/src/dird/ua_select.cc
alaaeddineelamri: I need this c-string in order to use it within another function (just a few lines below this declaration) that asks for a c-string argument.

core/src/dird/ua_select.cc
```diff
@@ -1124,6 +1174,9 @@ int DoPrompt(UaContext* ua,
   PoolMem pmsg(PM_MESSAGE);
   BareosSocket* user = ua->UA_sock;
 
+  int windowsize = 80;
+  int min_lines_threshold = 20;
```
arogge, line 1178:
```suggestion
  constexpr int windowsize = 80;
  constexpr int min_lines_threshold = 20;
```
you can make these `constexpr` so they'll end up as a compile-time constant instead of an actual variable.

core/src/dird/ua_select.h
```diff
@@ -54,6 +54,9 @@ bool SelectClientDbr(UaContext* ua, ClientDbRecord* cr);
 void StartPrompt(UaContext* ua, const char* msg);
 void AddPrompt(UaContext* ua, const char* prompt);
 void AddPrompt(UaContext* ua, std::string&& prompt);
+std::string FormatMulticolumnPrompts(const UaContext* ua,
+                                     const int windowsize,
+                                     const int min_lines_threshold);
```
arogge, line 59:
```suggestion
                                     int windowsize,
                                     int min_lines_threshold);
```
we usually don't declare by-value function parameters `const`

arogge (changes requested):
I just tried it out and it is really great!
However, I cannot make it use up all of the 80 text-columns. This should be improved.
Also I have some minor remarks on the code itself.

core/src/tests/multicolumn_prompts.cc
```diff
@@ -0,0 +1,177 @@
+/*
+   BAREOS® - Backup Archiving REcovery Open Sourced
+
+   Copyright (C) 2007-2011 Free Software Foundation Europe e.V.
```
pstorz, line 4:
```suggestion
```

core/src/tests/multicolumn_prompts.cc
```diff
@@ -0,0 +1,177 @@
...
+   Copyright (C) 2011-2012 Planets Communications B.V.
```
pstorz, line 5:
```suggestion
```

core/src/tests/multicolumn_prompts.cc
```diff
@@ -0,0 +1,177 @@
...
+   Copyright (C) 2019-2021 Bareos GmbH & Co. KG
```
pstorz, line 6:
```suggestion
   Copyright (C) 2021-2021 Bareos GmbH & Co. KG
```

pstorz (commented):
As the file "muldicolumn_prompts.cc" is new, only our Bareos GmbH Copyright applies here.

core/src/dird/ua_select.cc
```diff
@@ -1107,6 +1107,58 @@ void AddPrompt(UaContext* ua, std::string&& prompt)
...
+  }
+
+  const int extra_formatting_characters = 4;
+  const int max_formatted_prompt_length = max_prompt_length
+                                          + max_prompt_index_length
+                                          + extra_formatting_characters;
+  const int prompts_perline
+      = std::max(1, windowsize / (max_formatted_prompt_length - 1));
+
+  char formatted_prompt[max_formatted_prompt_length];
```
arogge, line 1136: what @franku meant was that you could use `std::vector<char> formatted_prompt(max_formatted_prompt_length)` instead of a `char[]`.
Then you would use `formatted_prompt.data()` and `formatted_prompt.size()` as parameters to `snprintf()` below.
This may be advantageous because a `std::vector<char>` will never decay to a pointer.

For your specific use I think using `char[]` is allright, however it is worth keeping in mind that you can always use a `std::vector<T>` instead of a `T[]` to get a safe buffer that always knows its size and can be resized easily.

core/src/dird/ua_select.cc
```diff
@@ -1124,6 +1176,9 @@ int DoPrompt(UaContext* ua,
   PoolMem pmsg(PM_MESSAGE);
   BareosSocket* user = ua->UA_sock;
 
+  int windowsize = 80;
```
arogge, line 1179: when reading `windowssize` my first idea is with and height. However, your variable just contains the width, so maybe `window_width` might be a better name.

core/src/dird/ua_select.h
```diff
@@ -54,6 +54,9 @@ bool SelectClientDbr(UaContext* ua, ClientDbRecord* cr);
 void StartPrompt(UaContext* ua, const char* msg);
 void AddPrompt(UaContext* ua, const char* prompt);
 void AddPrompt(UaContext* ua, std::string&& prompt);
+std::string FormatMulticolumnPrompts(const UaContext* ua,
+                                     int windowsize,
```
arogge, line 58: see above

arogge (commented):
besides a trivial rename and a clarification of @franku's previous comment, the code looks great.

arogge (approved):
Great work. Thank you!
I ran bareos-check-sources and it fixed 3 minor things. I also wrapped a long line in your commit message.

PR #744, opened 2021-02-23
stored: use "archive_device_string" variable for archive device

pstorz:
The "archive device" directive was stored in the variable "device_name"
in the "DeviceResource" Class and in "dev_name" in the "Device" Class.

This is misleading as the device name (i.e. the name of the device
resource) is stored in the "resource_name_" member of the device resource.

#### Consider the following chapters in the Bareos Developer Documentation

- [x] [Git Workflow](https://docs.bareos.org/DeveloperGuide/gitworkflow.html)
- [x] [Automatic Sourcecode Formatting](https://docs.bareos.org/DeveloperGuide/generaldevel.html#automatic-sourcecode-formatting)
- [x] [Check your commit messages](https://docs.bareos.org/DeveloperGuide/gitworkflow.html#commits)

#### Add some information

- [x] Add your name to the AUTHORS file

#### Keep spirit!

### DIFF
--- a/core/src/plugins/stored/scsicrypto/scsicrypto-sd.cc
+++ b/core/src/plugins/stored/scsicrypto/scsicrypto-sd.cc
@@ -2,7 +2,7 @@
    BAREOS® - Backup Archiving REcovery Open Sourced
 
    Copyright (C) 2012 Planets Communications B.V.
-   Copyright (C) 2013-2020 Bareos GmbH & Co. KG
+   Copyright (C) 2013-2021 Bareos GmbH & Co. KG
 
    This program is Free Software; you can redistribute it and/or
    modify it under the terms of version three of the GNU Affero General Public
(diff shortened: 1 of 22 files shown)


core/src/plugins/stored/scsicrypto/scsicrypto-sd.cc
```diff
@@ -340,7 +340,7 @@ static bRC do_set_scsi_encryption_key(void* value)
      * Check if encryption key is needed for reading this volume.
      */
     P(crypto_operation_mutex);
-    if (!NeedScsiCryptoKey(dev->fd(), dev->dev_name, true)) {
+    if (!NeedScsiCryptoKey(dev->fd(), dev->archive_device, true)) {
```
franku, line 343:
```suggestion
    if (!NeedScsiCryptoKey(dev->fd(), dev->archive_device_name, true)) {
```

franku (approved):
Good idea to give a meaningful name.. 

However, please update the changelog.

core/src/plugins/stored/scsicrypto/scsicrypto-sd.cc
pstorz: After a second thought I think that "archive_device" is the right name, as the directive that sets the value is "archive device",
See stored_conf.cc:

{"ArchiveDevice", CFG_TYPE_STRNAME, ITEM(res_dev, device_name), 0, CFG_ITEM_REQUIRED, NULL, NULL, NULL},

franku: Suggestion: archive_device_string

PR #754, opened 2021-02-25
Backport: python-bareos: fallback to ssl.PROTOCOL_SSLv23

franku:
This is a backport from master.

By default we set ssl.PROTOCOL_TLS.
However, this requires Python >= 2.7.13, which is not available by default on RHEL/CentOS 7.
Therefore we added a fallback to ssl.PROTOCOL_SSLv23.

(cherry picked from commit a3897f5ec36011153e1b05945819f57ac872c33b)

#### Keep spirit!


pstorz (approved):
Looks good.

PR #757, opened 2021-03-01
disable recent job results list for sparc and improve test

franku:
- rename test "recently-used-jobs" to "statefile"
- disable all "statefile" tests for sparc
- use temporary state files instead of original files in the source code which may be deleted during the test in case of an error

#### Checklist
- [x] [Git Workflow](https://docs.bareos.org/DeveloperGuide/gitworkflow.html)
- [x] [Automatic Sourcecode Formatting](https://docs.bareos.org/DeveloperGuide/generaldevel.html#automatic-sourcecode-formatting)
- [x] [Check your commit messages](https://docs.bareos.org/DeveloperGuide/gitworkflow.html#commits)

#### Add some information

- [x] Add your name to the AUTHORS file

#### Keep spirit!


core/src/tests/recent_job_results_list.cc
```diff
@@ -48,20 +48,50 @@ static bool StateFileExists(const char* dir, const char* name, int port)
   return stat(path.c_str(), &buf) == 0;
 }
 
+static bool CopyTestStateFileFromOriginal(std::string orig_path,
```
pstorz, line 51: As we now have c++17 available this should be easier with 
std::filesystem::copy

pstorz (changes requested):
I would suggest to change the name of the tests:

recent_job_results_list -> statefile:
read_job_results_from_flie -> read
write_job_results_to_file -> write
read_job_results_from_file_truncated_jobs -> handle_truncated_jobs
read_job_results_from_file_truncated_header -> handle_truncated_headers
read_job_results_from_file_not_exist -> handle_empty_file

core/src/tests/recent_job_results_list.cc
arogge: the filesystem interface is not really usable today. A lot of platforms require to link with a special lib and some don't support the interface yet.

core/src/tests/CMakeLists.txt
```diff
@@ -264,14 +266,29 @@ bareos_add_test(
...
-set_tests_properties(
-  gtest:recent_job_results_list.read_job_results_from_file PROPERTIES LABELS
-                                                                      "broken"
-)
-set_tests_properties(
-  gtest:recent_job_results_list.write_job_results_to_file PROPERTIES LABELS
-                                                                     "broken"
-)
+
+if(${CMAKE_HOST_SYSTEM_PROCESSOR} STREQUAL "sparc")
```
arogge, line 270: it doesn't make sense to check for the platform we're building on, we should check the platform we're building for.
```suggestion
if(${CMAKE_SYSTEM_PROCESSOR} STREQUAL "sparc")
```
Nevertheless, it should be rather simple to provide test-data for big endian, too. So there's no real point in disabling it.

franku: if(HAVE_BIG_ENDIAN) ..

pstorz (approved):
Looks good.

PR #792, opened 2021-04-12
gfapi-fd: Fix possible crash on second glfs_close()

sduehr:
Under rare circumstances, namely when a file was deleted during
backup after the glfs_open() and before the first glfs_close() call,
which then returns an error, the code assumed that the file was
closed and invoked a second glfs_close() call. This second call
could cause a segmentation fault.

The changed code no longer does a second glfs_close() call, the
first call already cleans up the glfs file descriptor properly,
even when an error occurs while trying to close the file.

See https://github.com/gluster/glusterfs/blob/34f854ad2f5d53578d7a4543f837cf0fb0f1a76a/api/src/glfs-fops.c#L513-L566

#### Please check

##### General

- [x] PR name is meaningful

##### Source code quality

- [x] `check-sources --since-merge` does not report any problems
- [ ] `git status` should not report modifications in the source tree after building and testing

### DIFF
--- a/core/src/plugins/filed/gfapi/gfapi-fd.cc
+++ b/core/src/plugins/filed/gfapi/gfapi-fd.cc
@@ -2,7 +2,7 @@
    BAREOS® - Backup Archiving REcovery Open Sourced
 
    Copyright (C) 2014-2017 Planets Communications B.V.
-   Copyright (C) 2014-2020 Bareos GmbH & Co. KG
+   Copyright (C) 2014-2021 Bareos GmbH & Co. KG
 
    This program is Free Software; you can redistribute it and/or
    modify it under the terms of version three of the GNU Affero General Public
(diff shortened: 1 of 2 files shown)


core/src/plugins/filed/gfapi/gfapi-fd.cc
```diff
@@ -1738,11 +1738,6 @@ static bRC pluginIO(PluginContext* ctx, struct io_pkt* io)
 
   switch (io->func) {
     case IO_OPEN:
-      /*
-       * Close the gfd when it was not closed before.
-       */
-      if (p_ctx->gfd) { glfs_close(p_ctx->gfd); }
```
pstorz, line 1744: Is this codepath really not needed anymore?
As far as I understand, p_ctx->gfd would be NULL in the case of the error that is fixed, 
so this glfs_close() call would be skipped anyway.

sduehr: Yes, as p_ctx->gfd is set to NULL unconditonally now on IO_CLOSE, the mentioned codepath would be never called.

PR #828, opened 2021-05-31
Fix acquire deadlock on migrate/copy

arogge:
Previously migrate/copy would acquire the device for reading before the
device for writing. This could lead to deadlock situation when the
volume that should be read was mounted in the device reserved for
writing.
By acquiring the device for writing first any volume in that device will
be unloaded unless it is the volume we are going to write to. This
should fix the deadlock described above.

#### Please check

##### General

- [x] PR name is meaningful

##### Source code quality


pstorz (changes requested):
This change looks good.

However I think it would make sense to add one line to the changelog that the order of acquisition of read device and append device was reversed and why.

pstorz (approved):
Good work!

PR #855, opened 2021-06-16
docs: change installation source to HTTPS

bluecmd:
I discovered this when following the installation guide for Debian and Ubuntu. It suggested to download the repository signing key over `http://` which in my book is a big no-no. Changing it to `https://` was fine, even for `apt-get update` and such.

I am guessing that this works for FreeBSD as well from what I can read, but not sure for zypper so I leave that as is.

#### Please check

- [ ] ~Your name is present in the AUTHORS file (optional)~

##### General

- [x] PR name is meaningful
- [ ] ~Separate commit for this PR in the CHANGELOG.md, PR number referenced is same~

### DIFF
--- a/docs/manuals/source/IntroductionAndTutorial/InstallingBareos.rst
+++ b/docs/manuals/source/IntroductionAndTutorial/InstallingBareos.rst
@@ -36,7 +36,7 @@ This will start a very basic Bareos installation which will regularly backup a d
 Decide about the Bareos release to use
 --------------------------------------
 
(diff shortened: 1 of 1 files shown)


arogge:
Would you mind updating the PR so it goes to the master branch? We will backport the change later.
This will work for `zypper addrepo`, too (I just tried it on SLES 12 SP5). So you can adapt that, too.

bluecmd:
@arogge Done!

arogge:
Thank you very much for taking the time to contribute to Bareos.
We really appreciate it!

PR #860, opened 2021-06-24
Console automatic reconnect to database without catalog directive

alaaeddineelamri:
#### Description

When a connection to the database is dropped, the director automatically tries to reconnect without checking for the "Reconnect" directive in the catalog.

#### Please check

##### General

- [x] PR name is meaningful

##### Source code quality

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ and since Bareos version 20 this project adheres to [Semantic Versioning](https:
 
 ## [Unreleased]
 
+### Breaking Changes
+
+- If you are relying on the fact that Bareos doesn't try to reconnect automatically on a database drop, you now have to specify it explicitly in the Catalog configuration with a `Reconnect = no` directive. [PR #860]
+
 ### Fixed
 - docs: Adapted the documentation of the VMware plugin due to update to VDDK 7 [PR #844]
 - fix a bug in VMware plugin where VMDK Files were created with wrong size when using the option localvmdk=yes [PR #826]
@@ -100,6 +104,7 @@ and since Bareos version 20 this project adheres to [Semantic Versioning](https:
 - version information was moved from core/cmake/ and webui/cmake/ directories into the toplevel cmake/ directory [PR #861]
 - add chromedriver options to improve reliability of selenium tests [PR #920]
 - docs: Describe how to get debugging info when using the VMware plugin [PR #921]
+- reconnecting to the database is now automatic per default without the need to specify it in the catalog [PR #860]
 
 ### Deprecated
 
@@ -186,6 +191,7 @@ and since Bareos version 20 this project adheres to [Semantic Versioning](https:
 [PR #844]: https://github.com/bareos/bareos/pull/844
 [PR #850]: https://github.com/bareos/bareos/pull/850
 [PR #858]: https://github.com/bareos/bareos/pull/858
+[PR #860]: https://github.com/bareos/bareos/pull/860
 [PR #861]: https://github.com/bareos/bareos/pull/861
 [PR #868]: https://github.com/bareos/bareos/pull/868
 [PR #869]: https://github.com/bareos/bareos/pull/869
--- a/core/src/cats/cats.h
+++ b/core/src/cats/cats.h
@@ -497,7 +497,7 @@ class BareosDb : public BareosDbQueryEnum {
   uint32_t ref_count_ = 0;                /**< Reference count */
   bool connected_ = false;                /**< Connection made to db */
   bool have_batch_insert_ = false;        /**< Have batch insert support ? */
-  bool try_reconnect_ = false;   /**< Try reconnecting DB connection ? */
+  bool try_reconnect_ = true;    /**< Try reconnecting DB connection ? */
   bool exit_on_fatal_ = false;   /**< Exit on FATAL DB errors ? */
   char* db_driver_ = nullptr;    /**< Database driver */
   char* db_driverdir_ = nullptr; /**< Database driver dir */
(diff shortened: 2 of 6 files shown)


core/src/cats/cats.h
```diff
@@ -497,7 +498,7 @@ class BareosDb : public BareosDbQueryEnum {
   uint32_t ref_count_ = 0;                /**< Reference count */
   bool connected_ = false;                /**< Connection made to db */
   bool have_batch_insert_ = false;        /**< Have batch insert support ? */
-  bool try_reconnect_ = false;   /**< Try reconnecting DB connection ? */
+  bool try_reconnect_ = true;    /**< Try reconnecting DB connection ? */
```
arogge, line 501: as every user of BareosDb is required to pass in `try_reconnect` when constructing, this won't make a difference

arogge (changes requested):
You have removed the parameter as discussed. However, as far as I can tell there are two ways to get a database connection in the director, one is `OpenDb()` in ua_db.cc and the other one is `GetDatabaseConnection()` in get_database_connection.c.

In the end, all connections for the console, scheduler and things like that should have `try_reconnect = true` and all connections for a job (each job will have its own database connection) should have `try_reconnect = false`.
You will have to find out where each of the above functions is called. I think we want `try_reconnect = true` in `OpenDb()` and `try_reconnect = false` in GetDatabaseConnection()`. However, I'm not sure this is correct in every case, so you'll have to take a deeper look.

core/src/dird/dird_conf.cc
```diff
@@ -327,8 +327,6 @@ static ResourceItem cat_items[] = {
    /* Turned off for the moment */
   { "MultipleConnections", CFG_TYPE_BIT, ITEM(res_cat, mult_db_connections), 0, 0, NULL, NULL, NULL },
   { "DisableBatchInsert", CFG_TYPE_BOOL, ITEM(res_cat, disable_batch_insert), 0, CFG_ITEM_DEFAULT, "false", NULL, NULL },
-  { "Reconnect", CFG_TYPE_BOOL, ITEM(res_cat, try_reconnect), 0, CFG_ITEM_DEFAULT, "false",
-     "15.1.0-", "Try to reconnect a database connection when its dropped" },
```
arogge, line 331: this will make the system:config-dump test fail (as you removed the patameter). So you'll need to update the test, too.

core/src/cats/cats.h
```diff
@@ -42,7 +42,8 @@
 
 #include <string>
 #include <vector>
-template <typename T> class dlist;
+template <typename T>
+class dlist;
```
arogge, line 46: a recent version of clang-format should format this it in a single line with our configuration. Maybe you need an update? You can always use the clang-format in devtools/clang-format

core/src/dird/dird_conf.h
```diff
@@ -250,8 +251,8 @@ class CatalogResource : public BareosResource {
   uint32_t mult_db_connections = 0; /**< Set if multiple connections wanted */
   bool disable_batch_insert
       = false;                /**< Set if batch inserts should be disabled */
-  bool try_reconnect = false; /**< Try to reconnect a database connection when
-                         its dropped */
+  bool try_reconnect = true;  /**< Try to reconnect a database connection when
+                          it is dropped */
```
arogge, line 255: when the parameter from the configuration is removed, this should probably be removed, too. Which will make the compiler complain if you overlooked any used of `try_reconnect` somewhere.
It will at least require a change in the other call to `DbSqlGetPooledConnection()`.

core/src/dird/get_database_connection.cc
```diff
@@ -32,15 +32,15 @@ namespace directordaemon {
...
   return DbSqlGetPooledConnection(
       jcr, jcr->impl->res.catalog->db_driver, jcr->impl->res.catalog->db_name,
       jcr->impl->res.catalog->db_user,
       jcr->impl->res.catalog->db_password.value,
       jcr->impl->res.catalog->db_address, jcr->impl->res.catalog->db_port,
       jcr->impl->res.catalog->db_socket,
       jcr->impl->res.catalog->mult_db_connections,
-      jcr->impl->res.catalog->disable_batch_insert,
-      jcr->impl->res.catalog->try_reconnect,
+      jcr->impl->res.catalog->disable_batch_insert, try_reconnect,
```
arogge, line 43: you can also save the variable completely and write
```suggestion
      jcr->impl->res.catalog->disable_batch_insert, 
      false /* try_reconnect */,
```

CHANGELOG.md
```diff
@@ -82,6 +82,7 @@ and since Bareos version 20 this project adheres to [Semantic Versioning](https:
 - packages: Build also for Fedora_34 [PR #869]
 - postgresql filedaemon plugin: switched from psycopg2 to pg8000, dropped support for python2.
 - version information was moved from core/cmake/ and webui/cmake/ directories into the toplevel cmake/ directory [PR #861]
+- reconnecting to the database is now automatic without the need to specify it in the catalog, except for ongoing jobs, reconnection for those is disabled.
```
arogge, line 85: I would also add a record under "removed" that tells the reader the catalog's reconnect parameter has been removed.

arogge (changes requested):
Looks promising. I'd remove the `try_reconnect` member from `CatalogResource` and set it explicitly to the values we want.
You'll have to look at the test system:config-dump and also building the documentation will currently fail as the json-Files with the resource parameters will need an update when a parameter is added, changed or removed.

arogge:
To check if the JSON schema files for the documentation are up-to-date you can run `make check-git`.
To update these files, you'll have to configure bareos with `cmake -Ddocs-build-json=yes [...]` and then `make`. This should update the files to not contain the parameter anymore.

Concerning the systemtest, you'll just have to update the reference-file that the dumped configuration is compared against.

arogge (commented):
Probably worth to merge the first two commits. Other than that, and with the knowledge you/we gained on the way, I believe it is ready to be merged.

arogge (changes requested):
Besides the now incorrect changelog record, that should be removed, this PR looks great to me. Improves the default behaviour a lot.

arogge (approved):

PR #923, opened 2021-09-14
Adapt Python DIR and SD plugin Baseclasses to the modernized Python plugin API

sduehr:
This PR adapts the Python DIR and SD plugin Baseclasses to the modernized Python plugin API, see
https://docs.bareos.org/TasksAndConcepts/Plugins.html#modernization-of-the-python-plugin-api

It also contains a fix for the DIR plugin, where DebugMessage after certain events caused
a plugin_ctx is unset runtime error.

#### Please check

##### General

- [x] PR name is meaningful
- [ ] Separate commit for this PR in the CHANGELOG.md, PR number referenced is same

##### Source code quality

### DIFF
--- a/systemtests/tests/py2plug-dir/python-modules/bareos-dir-test.py
+++ b/systemtests/tests/py2plug-dir/python-modules/bareos-dir-test.py
@@ -1,6 +1,6 @@
 # BAREOS - Backup Archiving REcovery Open Sourced
 #
-# Copyright (C) 2019-2020 Bareos GmbH & Co. KG
+# Copyright (C) 2019-2021 Bareos GmbH & Co. KG
 #
 # This program is Free Software; you can redistribute it and/or
 # modify it under the terms of version three of the GNU Affero General Public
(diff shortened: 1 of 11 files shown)


systemtests/tests/py2plug-dir/python-modules/bareos-dir-test.py
```diff
@@ -19,59 +19,18 @@
 #
 # Author: Tobias Plum
 #
-from bareosdir import *
+import bareosdir
+import BareosDirWrapper
+from BareosDirWrapper import *  # noqa
```
pstorz, line 24: Can this * import be avoided?

systemtests/tests/py2plug-sd/python-modules/bareos-sd-test.py
```diff
@@ -19,86 +19,18 @@
 #
 # Author: Tobias Plum
 #
-from bareossd import *
+import bareossd
+import BareosSdWrapper
+from BareosSdWrapper import *  # noqa
```
pstorz, line 24: Same here, is the * import needed?

pstorz (changes requested):
Good work. 

Only two comments:
* If the asterisk imports cannot be avoided 
* the CHANGELOG.md entry is missing

systemtests/tests/py2plug-dir/python-modules/bareos-dir-test.py
sduehr: Not really, we could use
`from BareosDirWrapper import parse_plugin_definition, handle_plugin_event`
but that's not much nicer, and flake8 would still complain.

For a better solution which would only need the `import BareosDirWrapper` (or may be no wrapper module at all) we would have to change bareosdir.cc (may be also pythton-dir.cc), as they lookup  the functions in the module namespace. But that's another story, I don't think that we should do that within this PR.

systemtests/tests/py2plug-sd/python-modules/bareos-sd-test.py
sduehr: See comment above, same situation here.

pstorz (changes requested):
Thanks you for the explanation regarding the * imports. 

What about the CHANGELOG.md entry?

sduehr:
Added CHANGELOG.md entry.

pstorz (approved):

PR #948, opened 2021-10-01
build: remove warnings during build

pstorz:
#### Please check

- [x] merge the following PR also when merging: https://git.bareos.com/bareos/CD/pulls/37

##### General

- [x] PR name is meaningful

##### Source code quality


core/src/droplet/libdroplet/src/backend/cdmi/replyparser.c
```diff
@@ -586,7 +586,7 @@ dpl_status_t dpl_cdmi_parse_list_bucket(dpl_ctx_t* ctx,
     goto end;
   }
 
-  children = json_object_object_get(obj, "children");
+  json_object_object_get_ex(obj, "children", &children);
```
arogge, line 589: return code is book and should probably be checked. The docs doesn't state what `children` will be set to on error, so we should probably assume it can be undefined and not rely on comparing with `NULL` only.

core/src/lib/crypto_openssl.cc
```diff
@@ -1477,13 +1477,15 @@ static int crypto_initialized = false;
  */
 int InitCrypto(void)
 {
-  int status;
+  int status = 0;
 
+#  if (OPENSSL_VERSION_NUMBER <= 0x10000000L)
```
arogge, line 1482: You cannot remove that for OpenSSL 1.0.0. Otherwise locking won't work any threading will break.
However, you can probably remove it for OpenSSL 1.1.0, but please make really sure none of the removed calls is required (we had broken that in the past already).

core/src/plugins/dird/python/CMakeLists.txt
```diff
@@ -25,7 +25,7 @@ if(Python2_FOUND)
   )
   install(TARGETS python-dir DESTINATION ${plugindir})
   target_link_libraries(python-dir ${Python2_LIBRARIES} bareos)
-  target_include_directories(python-dir PUBLIC SYSTEM ${Python2_INCLUDE_DIRS})
+  target_include_directories(python-dir SYSTEM PUBLIC ${Python2_INCLUDE_DIRS})
```
arogge, line 28: while this will silence the warning for strict-aliasing, we should probably also set `-fno-strict-aliasing` as a compile option so the compiler will not generate broken code.

arogge (commented):
still needs some work, some of the `read()` and `write()` calls are still not checked thoroughly.
But good progress anyway!

arogge (approved):

PR #965, opened 2021-10-26
director: fix crash in status scheduler when client is not set

pstorz:
The `status scheduler` command did not check the `job->client` pointer before dereferencing. This is fixed with this PR.
#### Please check

##### General

- [x] PR name is meaningful

##### Source code quality

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -63,6 +63,7 @@ and since Bareos version 20 this project adheres to [Semantic Versioning](https:
 - Fixed libdroplet xattr.h include issue by using sys/xattr.h [PR #985]
 - Fixed crash on bconsole when using autcomplete with tab [PR #969]
 - [Issue #1374] Include zero-file incremental backups in always-incremental consolidation [PR #995]
+- fix crash in "status scheduler" command when job->client is unset [PR #965]
 
 ### Added
 - systemtests: make database credentials configurable [PR #950]
--- a/core/src/dird/ua_status.cc
+++ b/core/src/dird/ua_status.cc
@@ -662,7 +662,7 @@ start_again:
       // List specific schedule.
       if (job) {
         if (job->schedule && job->schedule->enabled && job->enabled
-            && job->client->enabled) {
+            && job->client && job->client->enabled) {
           if (!show_scheduled_preview(ua, job->schedule, overview,
                                       &max_date_len, time_to_check)) {
             goto start_again;
@@ -674,7 +674,7 @@ start_again:
           if (!ua->AclAccessOk(Job_ACL, job->resource_name_)) { continue; }
 
           if (job->schedule && job->schedule->enabled && job->enabled
-              && job->client == client && job->client->enabled) {
+              && job->client && job->client == client && job->client->enabled) {
             if (!show_scheduled_preview(ua, job->schedule, overview,
                                         &max_date_len, time_to_check)) {
               job = NULL;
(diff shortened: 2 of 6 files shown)


arogge (dismissed):
Just tried with your changes.
Configured a copy-job with a disabled client and ran `status scheduler job=copy`.
Threw the direcotor in an infinite loop eating CPU, so something is obviously still broken.

systemtests/scripts/setup
```diff
@@ -13,6 +13,9 @@ print_debug "creating database \"${db_name}\" (${DBTYPE})"
 "${scripts}/make_bareos_tables" "${DBTYPE}" >/dev/null 2>&1
 "${scripts}/grant_bareos_privileges" "${DBTYPE}" >/dev/null 2>&1
 
+# Fill ${BackupDirectory} with data.
+setup_data
+
```
arogge, line 18: this will run setup_data for every systemtest that calls `${rscripts}/setup` is this intended?

arogge (commented):
in ua_status.cc we have lines 658 to 716 with 7 empty or comment lines.
That's a net amount of 51 lines of code.

These contain:
* 16 branches (14 `if`s, 2 `else`s)
* 3 loops (1 `while`, 2 `foreach_res`)
* 9 jumps (3 `goto`s, 5 `continue`s, 1 `break`)

So there is a loop, branch or some other kind of jump every 2 lines.

What I'm trying to say is: I would love to review the code, but I need to understand it.
Don't you think this should be refactored completely so we can understand what is going on without drawing a complex diagram?

systemtests/scripts/setup
alaaeddineelamri: In the old way (testrunner without subtests), the testrunner would setup the environment as well as the database, setup the custom data, and then finally run the bareos binaries.

When we do the subtest, the routine changes: the `test-setup` script sets up the environment as well as the database, but starts running the binaries without setting up the custom data, which falls in the hands of the new `testrunner-<your-test-name>`. At first hand, this seemed not important, as other systemtests that need data have no problem with it. But with the `scheduler-backup` it is different. When running the test, it complains about not finding the custom data. As far as I've understood, this happens because the first job of the scheduler runs as soon as the director starts. That happens quickly before the testrunner deploys the custom data. So, when trying the find the data to backup, bareos stalls, as there seems to be an issue with the sd and fd, and things get stuck there, which makes the `wait` command stuck too, which in turn gets the whole test stuck.

Setting up the custom data for the test *before* running the binaries, makes sure the first scheduled job runs properly on the start of director, and the rest of the test goes smoothly.

Other tricks like disabling and enabling the scheduler, had no impact, and then doing this change made the most sense for me.

It felt setting up the data in the setup script would make sense. Even if other tests don't need data, it doesn't cost too much in setup time, plus it could be useful for future ramifications of the test or extra subtests added.

Another solution would have been to leave the `scheduler-backup` systemtest as is, and create a whole new test, but I think for the functionality we're adding, it feels too much overkill, and converting the whole thing into subtests felt better.

let me know what you think of this

alaaeddineelamri: Another useful benefit of having `setup_data` in the `setup` script I just encountered, is that within any systemtest, you could run `test-setup` and have an environment ready to use, without having to run the entire testrunner. This gives more flexibility when experimenting within the environment of a certain test, or for debugging or anything of that sort

alaaeddineelamri: After a conversation with Joerg, I could handle this issue without having to setup the data in the `${rscripts}/setup` script.

arogge (approved):

PR #966, opened 2021-10-28
console: prune command gained support to prune multiple volumes at once

alaaeddineelamri:
#### Description:

This PR enables the prune command to prune 
 * all volumes of a specific pool
 * all volumes of all pools

#### Please check

##### General

- [x] PR name is meaningful

##### Source code quality

### DIFF
--- a/core/src/cats/sql_get.cc
+++ b/core/src/cats/sql_get.cc
@@ -3,7 +3,7 @@
 
    Copyright (C) 2000-2012 Free Software Foundation Europe e.V.
    Copyright (C) 2011-2016 Planets Communications B.V.
-   Copyright (C) 2013-2020 Bareos GmbH & Co. KG
+   Copyright (C) 2013-2021 Bareos GmbH & Co. KG
 
    This program is Free Software; you can redistribute it and/or
    modify it under the terms of version three of the GNU Affero General Public
--- a/core/src/cats/sql_get.cc
+++ b/core/src/cats/sql_get.cc
@@ -1079,6 +1079,27 @@ bool BareosDb::GetQueryDbids(JobControlRecord* jcr,
 
 
 /**
+ * Get all volumes names
+ *
+ * Returns: false: on failure
+ *          true:  on success
+ */
+bool BareosDb::GetAllVolumeNames(db_list_ctx* volumenames)
+{
+  PoolMem query(PM_MESSAGE);
+  volumenames->clear();
+  Mmsg(query,
+       "SELECT DISTINCT Media.VolumeName FROM Media ORDER BY VolumeName");
+
+  if (!SqlQueryWithHandler(query.c_str(), DbListHandler, volumenames)) {
+    Emsg1(M_ERROR, 0, "Could not retrieve volume names: ERR=%s\n",
+          sql_strerror());
+    return false;
+  }
+  return true;
+}
+
+/**
  * Get Media Record
  *
  * Returns: false: on failure
--- a/core/src/dird/ua_output.cc
+++ b/core/src/dird/ua_output.cc
@@ -1627,8 +1627,7 @@ void UaContext::vSendMsg(int signal,
   PoolMem message;
   send->SendBuffer();
   if (signal) {
-    BareosSocket* bs = UA_sock;
-    if (bs && api) bs->signal(signal);
+    if (UA_sock && api) UA_sock->signal(signal);
   }
   message.Bvsprintf(fmt, arg_ptr);
   if (console_is_connected) {
--- a/core/src/dird/ua_output.cc
+++ b/core/src/dird/ua_output.cc
@@ -1632,7 +1632,12 @@ void UaContext::SendMsg(const char* fmt, ...)
   va_start(arg_ptr, fmt);
   message.Bvsprintf(fmt, arg_ptr);
   va_end(arg_ptr);
-  send->message(NULL, message);
+
+  if (console_is_connected) {
+    send->message(NULL, message);
+  } else { /* No UA, send to Job */
+    Jmsg(jcr, M_INFO, 0, "%s", message.c_str());
+  }
 }
 
 void UaContext::SendRawMsg(const char* msg) { SendMsg(msg); }
@@ -1655,7 +1660,11 @@ void UaContext::ErrorMsg(const char* fmt, ...)
   va_start(arg_ptr, fmt);
   message.Bvsprintf(fmt, arg_ptr);
   va_end(arg_ptr);
-  send->message(MSG_TYPE_ERROR, message);
+  if (console_is_connected) {
+    send->message(MSG_TYPE_ERROR, message);
+  } else { /* No UA, send to Job */
+    Jmsg(jcr, M_INFO, 0, "%s", message.c_str());
+  }
 }
 
 /**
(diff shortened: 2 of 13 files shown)


core/src/cats/sql_get.cc
```diff
@@ -1078,6 +1078,23 @@ bool BareosDb::GetQueryDbids(JobControlRecord* jcr,
...
+ */
+bool BareosDb::GetAllVolumeNames(db_list_ctx* volumenames)
+{
+  PoolMem query(PM_MESSAGE);
+  volumenames->clear();
+  Mmsg(query,
+       "SELECT DISTINCT Media.VolumeName FROM Media ORDER BY VolumeName");
+  SqlQueryWithHandler(query.c_str(), DbListHandler, volumenames);
+
+  return true;
```
arogge, line 1095: if you never fail, why return a value a all?

core/src/dird/ua_prune.cc
```diff
@@ -230,6 +231,53 @@ bool PruneCmd(UaContext* ua, const char* cmd)
...
+          } else if (!ConfirmRetention(ua, &mr.VolRetention, msg.c_str())) {
+            result = false;
+            continue;
+          }
+
+          result &= PruneVolume(ua, &mr);
+        }
+
+        return result;
+      }
```
arogge, line 279: so `prune all` does nothing at all? Not even a warning or error message?

core/src/dird/ua_prune.cc
```diff
@@ -230,6 +231,53 @@ bool PruneCmd(UaContext* ua, const char* cmd)
...
+                _("Cannot prune Volume \"%s\" because it is archived.\n"),
+                mr.VolumeName);
+            result = false;
+            continue;
+          }
+
+          std::string msg("Volume");
+          msg.append(" (" + volume + ")");
+          if (pool && pool->VolRetention > 0) {
+            if (!ConfirmRetention(ua, &pool->VolRetention, msg.c_str())) {
```
arogge, line 266: `pool->VolRetention` may differ from `mr.VolRetention`. In that case `mr.VolRetention` takes precedence.
Thus, checking only `mr.VolRetention` is probably the right thing to do.

core/src/dird/ua_prune.cc
```diff
@@ -230,6 +231,53 @@ bool PruneCmd(UaContext* ua, const char* cmd)
...
+              result = false;
+              continue;
+            }
+          } else if (!ConfirmRetention(ua, &mr.VolRetention, msg.c_str())) {
+            result = false;
+            continue;
+          }
+
+          result &= PruneVolume(ua, &mr);
```
arogge, line 275:
```suggestion
          if (ConfirmRetention(ua, &mr.VolRetention, msg.c_str())) {
            result &= PruneVolume(ua, &mr);
          } else {
            result = false;
          }
```

core/src/dird/ua_prune.cc
```diff
@@ -593,7 +642,8 @@ static int JobSelectHandler(void* ctx, int num_fields, char** row)
   res->ClientId = str_to_int64(row[4]);
   lst->append(res);
 
-  // Dmsg2(150, "row=%d val=%d\n", del->num_ids-1, del->JobId[del->num_ids-1]);
+  // Dmsg2(150, "row=%d val=%d\n", del->num_ids-1,
+  // del->JobId[del->num_ids-1]);
```
arogge, line 646:
```suggestion
```

core/src/dird/ua_select.cc
```diff
@@ -862,23 +865,35 @@ bool SelectStorageDbr(UaContext* ua, StorageDbRecord* sr, const char* argk)
 }
 
 // Select a Media (Volume) record from the database
-bool SelectMediaDbr(UaContext* ua, MediaDbRecord* mr)
+bool SelectMediaDbr(UaContext* ua, MediaDbRecord* mr, const char* volumename)
```
arogge, line 868: I think this function already did way too much before you added another feature.
The problem is `SelectPoolAndMediaDbr()` which clearly violates SRP (the name already gives it away).
I'd refactor `SelectPoolAndMediaDbr()` into `SelectPoolForMedia()` by just removing the call to `SelectMediaDbr()` at the beginning and requiring the caller to have `MediaDbRecord*` already filled.
You'll then need to add a call to call `SelectMediaDbr(ua, mr)` before the one existing use of `SelectPoolAndMediaDbr()`.

After refactoring like that, you can then focus on what you really need and implement a new function that will do the following:
* reset `mr`
* fill `mr->VolumeName`
* call `ua->db->GetMediaRecord(ua->jcr, mr)`

I guess something like this should be sufficient (and way easier to understand):
```
bool SelectMediaByName(UaContext* ua, MediaDbRecord* mr, const char* volumename) {
  new (mr) MediaDbRecord();
  bstrncpy(mr->VolumeName, volumename, sizeof(mr->VolumeName));
  if (!ua->db->GetMediaRecord(ua->jcr, mr)) {
    ua->ErrorMsg("%s", ua->db->strerror());
    return false;
  }
  return true;
}
```
If you want to go the extra mile, you can add a new constructor `MediaDbRecord(const char*)` so that you can pass `volumename` as part of the placement-new.

arogge (changes requested):
I haven't checked the functionality yet, but I have a few comments about the code.

core/src/cats/sql_get.cc
alaaeddineelamri: actually forgot to run a check there, thanks!

core/src/dird/ua_prune.cc
alaaeddineelamri: it effectively does nothing ... thanks!

core/src/dird/ua_prune.cc
alaaeddineelamri: makes sense

pstorz (changes requested):
While testing I found some problems.
The most important is that the output from pruning is stored in the messages buffer, so that now we can call "prune all volume yes", and then we got "you have messages". Then we need to execute "messages" to see what  is going on.

I changed the prune and purge functions to log differently so that they now print their results immediately when running in console, and logging to the joblog if a job is the initiator of the pruning. (See last commit)

Besides of this, I added some commtent.

Good work!

core/src/dird/ua_output.cc
```diff
@@ -1617,22 +1617,37 @@ bool bsendmsg(void* ctx, const char* fmt, ...)
...
+                         const char* messagetype,
+                         const char* fmt,
+                         va_list arg_ptr)
+{
+  /* send current buffer */
+  PoolMem message;
+  send->SendBuffer();
+  if (signal) {
+    BareosSocket* bs = UA_sock;
+    if (bs && api) bs->signal(signal);
```
arogge, line 1631: no need for a pointer alias.
```suggestion
    if (UA_sock && api) UA_sock->signal(signal);
```

pstorz: Good point

pstorz (approved):

pstorz (changes requested):
I just noticed, that we need to update the "prune command" section in the docs:

https://docs.bareos.org/TasksAndConcepts/BareosConsole.html#index-34

pstorz (approved):
Good work!

PR #986, opened 2021-11-15
Task Solution for Bareos test for the offer : Backend Developer

alaamzoughi:
Contributor : Alaa Mzoughi 
Action : Fix the The problem of the Trace in setdebug 
The tasks solutions and screenshots was sended on : jobs@bareos.com


pstorz (changes requested):
Hello,
thank you very much for your PR.
However, I think that something went wrong as your PR changes more than 5.000
Files.
Could you change your PR to only contain the changes that you did?
Thank you very much.
If you need help with this task please ask.
best regards,
Philipp

PR #1021, opened 2021-12-09
stored: enable labeling of tapes in  devices even when `autoselect = no`

alaaeddineelamri:
#### Description:

When devices are set with the option `autoselect=no`, labeling and releasing is not possible as the device is skipped in the search. This PR handles this issue by not accounting for the `autoselect` attribute when searching devices. This has the same effect on the `mount`, `unmount`, as well as `readlabel` and `autochanger` commands.  Backups are not affected. 

#### Please check

##### General

- [x] PR name is meaningful

##### Source code quality

##### Tests

- [x] Decision taken that a system- or unittest is required (if not, then remove this paragraph)
- [x] The decision towards a systemtest is reasonable compared to a unittest
- [x] Output of the test leads quickly to the origin of the fault

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ and since Bareos version 20 this project adheres to [Semantic Versioning](https:
 - webui: remove an unnecessary .bvfs_get_jobids and buildSubtree() call [PR #1050]
 - git: set merge strategy for CHANGELOG.md to union [PR #1062]
 - webui: add timeline chart by jobs [PR #1059]
+- stored: enable labeling of tapes in drives even if `autoselect=no` [PR #1021] 
 
 ### Deprecated
 
--- a/core/src/stored/dir_cmd.cc
+++ b/core/src/stored/dir_cmd.cc
@@ -3,7 +3,7 @@
 
    Copyright (C) 2001-2012 Free Software Foundation Europe e.V.
    Copyright (C) 2011-2012 Planets Communications B.V.
-   Copyright (C) 2013-2021 Bareos GmbH & Co. KG
+   Copyright (C) 2013-2022 Bareos GmbH & Co. KG
 
    This program is Free Software; you can redistribute it and/or
    modify it under the terms of version three of the GNU Affero General Public
--- a/core/src/stored/dir_cmd.cc
+++ b/core/src/stored/dir_cmd.cc
@@ -920,11 +920,7 @@ static DeviceControlRecord* FindDevice(JobControlRecord* jcr,
                  device_resource->resource_name_, devname.c_str());
             continue;
           }
-          if (!device_resource->dev->autoselect) {
-            Dmsg1(100, "Device %s not autoselect skipped.\n", devname.c_str());
-            continue; /* device is not available */
-          }
-          if (drive == kInvalidDriveNumber
+          if ((drive == kInvalidDriveNumber && device_resource->dev->autoselect)
               || drive == device_resource->dev->drive) {
             Dmsg1(20, "Found changer device %s\n",
                   device_resource->resource_name_);
--- a/systemtests/tests/autochanger/test-setup
+++ b/systemtests/tests/autochanger/test-setup
@@ -1,6 +1,8 @@
+#!/bin/bash
+
 #   BAREOS® - Backup Archiving REcovery Open Sourced
 #
-#   Copyright (C) 2021-2021 Bareos GmbH & Co. KG
+#   Copyright (C) 2022-2022 Bareos GmbH & Co. KG
 #
 #   This program is Free Software; you can redistribute it and/or
 #   modify it under the terms of version three of the GNU Affero General Public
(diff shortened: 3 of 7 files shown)


CHANGELOG.md
```diff
@@ -151,6 +151,7 @@ and since Bareos version 20 this project adheres to [Semantic Versioning](https:
 - webui: format total jobs and total files language-sensitive [PR #996]
 - webui: optimize bvfs update cache calls [PR #999]
 - Replace complicated memory pool allocator with plain malloc() [PR #997]
+- Allow commands `label`, `release`, `mount`, `unmount`, also `readlabel` and `autochanger` to run when `autoselect=no` [PR #1021]
```
pstorz, line 154:
```suggestion
- enable labeling of tapes in drives even if `autoselect=no` [PR #1021] 
```

systemtests/tests/autochanger/testrunner
```diff
@@ -61,6 +79,17 @@ for i in $(seq ${NUMBER_OF_POOLS}); do
   done
 done
 
+#resetting only 2 devices to `autoselect = yes` to check that devices with `autoselect = no` don't get used.
```
pstorz, line 82: This change now only allows two jobs to run on two drives in parallel. Before, four jobs were writte on four tapes which makes sense to test also.
I think the functionality that devices set to autoselect=no are not selected is already tested in the multiplied-devices test and so does not need to be tested here.

pstorz: The whole idea of this change is that we want to be able to label tapes even if all drives are busy.
For that, we want to free one drive by disabling the autoselect, then be able to label a tape in the free drive, and 
after the label is done reenable the drive so that it will be used for  backups again.

So we would need something like this:
- all drives are busy with jobs the whole time
- while this is the case, we set autoselect to no one drive.
- wait until the drive is not in use anymore
- label a new tape in this drive
- set autoselect to yes again
- check if the next jobs will use the drive again
- It might make sense to do this in an own test.

pstorz (dismissed):
Excellent work!
It might make sense to refactor this test to have setup and teardown, and to have two autochanger tests (the original one and the new one testing the label functionality)

systemtests/tests/autochanger/testrunner
alaaeddineelamri: Thanks!

done!

core/src/stored/dir_cmd.cc
```diff
@@ -920,10 +920,6 @@ static DeviceControlRecord* FindDevice(JobControlRecord* jcr,
                  device_resource->resource_name_, devname.c_str());
             continue;
           }
-          if (!device_resource->dev->autoselect) {
-            Dmsg1(100, "Device %s not autoselect skipped.\n", devname.c_str());
-            continue; /* device is not available */
-          }
           if (drive == kInvalidDriveNumber
```
arogge, line 923: When you call `FindDevice()` with `drive == kInvalidDriveNumber` it will simply return the first drive (i.e. kInvalidDriveNumber means "I don't care which drive").
If the caller doesn't care, we should probably only return a device with `autoselect = yes`.
```suggestion
          if ((drive == kInvalidDriveNumber && device_resource->dev->autoselect)
```

arogge (commented):
This looks like a good idea, I just wouldn't return a device with `autoselect = no` unless the caller explicitly asks for that one.

systemtests/tests/autochanger/test-setup
```diff
@@ -0,0 +1,53 @@
+#!/bin/bash
+
+#   BAREOS® - Backup Archiving REcovery Open Sourced
+#
+#   Copyright (C) 2021-2021 Bareos GmbH & Co. KG
```
arogge, line 5:
```suggestion
#   Copyright (C) 2022-2022 Bareos GmbH & Co. KG
```

systemtests/tests/autochanger/testrunner-label-release-autoselect
```diff
@@ -0,0 +1,136 @@
...
+      spooling=""
+    fi
+  done
+done
+
+# start all the jobs
+bin/bconsole < bconsole_backup_jobs
+
+testdevice0=tapedrive-0-.dev.tape.by-id.scsi-350223344ab000100-nst
+testdevice1=tapedrive-1-.dev.tape.by-id.scsi-350223344ab000200-nst
```
arogge, line 59: The device names can easily change. These are just the devices mhvtl provides, but you can change the WWNs in mhvtl. Or maybe you run the tests with a real library, which will have different WWNs, too.
Also, people might have just one drive (esp. with a real library) so you cannot always rely on having two drives available - maybe we need to disable this (sub)test if there is only one drive available.

arogge (changes requested):

systemtests/tests/autochanger/testrunner-label-release-autoselect
alaaeddineelamri: you are right, I've changed the devices from hardcoded ones to being extracted from the config. I also changed it so it uses only one device.

arogge (changes requested):
bareos-check-sources is not happy yet.
Other than that, it looks great and can be merged.

arogge (approved):

PR #1040, opened 2021-12-30
dir, stored: start statistics threads only if needed

alaaeddineelamri:
#### Description:
The director statistics thread starts even when not needed. In order to optimize and have less unnecessary workload, the statistics thread for the director is not started by default, unless  `collect statistics` and `statistics collect interval` are both specified.

#### Please check

##### General

- [x] PR name is meaningful

##### Source code quality


core/src/dird/stats.cc
```diff
@@ -322,7 +330,7 @@ int StartStatisticsThread(void)
 
   statistics_initialized = true;
 
-  return 0;
+  return 1;
```
pstorz, line 333: Why do we have a return code here at all? All calls to "StartStatisticsThread()" that i saw never checked the return code. so can't we just change ther return code to void?

And can't we do the same with StartStatisticsThread() in sd_stats.cc also?

core/src/tests/addresses_and_ports.cc
```diff
@@ -135,8 +85,8 @@ static bool test_sockets(int family, int port)
   bool result = true;
 
   if (family == 0) {
-    result = create_and_bind_v4socket(v4_fd, port)
-             && create_and_bind_v6socket(v6_fd, port);
+    result = create_and_bind_v4socket(v4_fd, port);
+    result &= create_and_bind_v6socket(v6_fd, port);
```
pstorz, line 89: Why this change? I guess to avoid short-cicuit evalutiont, right? However, &= is a bitwise AND and not logical AND.
For clarification it might make sense to introduce `bool result_v4` and `bool result_v6`and then have `result = restult_v4 && result_v6`

pstorz (changes requested):

core/src/dird/stats.cc
pstorz: In case the return code is required, we should check it in the code that calls StartStatisticsThread()

core/src/tests/addresses_and_ports.cc
alaaeddineelamri: good point

core/src/dird/stats.cc
alaaeddineelamri: You are right, the return value is not being used at all, but I believe we should make use of it. I already started using in the test, and I think we should even use the return value to send a message in case of a failed starting attempt. 
Let me know what you think of this

pstorz: Ok, then the return code of course needs to be checked and handeled everywhere. So the question is what to do if the start fails? Do we only log that or do we even stop the daemon in that case?

pstorz (changes requested):
The code that handles the return status of StartStatisticsThread() in both storage and director needs to be added.
Good work!

pstorz (approved):

core/src/dird/dird.cc
```diff
@@ -469,7 +469,12 @@ int main(int argc, char* argv[])
 
   //   InitDeviceResources();
 
-  StartStatisticsThread();
+  if (!StartStatisticsThread()) {
+    BErrNo be;
+    Emsg1(M_WARNING, 0,
```
pstorz, line 474:
```suggestion
    Emsg1(M_ERROR_TERM, 0,
```

core/src/dird/dird.cc
```diff
@@ -673,7 +678,12 @@ bool DoReloadConfig()
     } else {  // no jobs running
       FreeSavedResources(&prev_config);
     }
-    StartStatisticsThread();
+    if (!StartStatisticsThread()) {
+      BErrNo be;
+      Emsg1(M_WARNING, 0,
```
pstorz, line 683:
```suggestion
      Emsg1(M_ERROR_TERM, 0,
```

core/src/stored/stored.cc
```diff
@@ -353,7 +353,12 @@ int main(int argc, char* argv[])
         me->jcr_watchdog_time); /* start JobControlRecord watchdogs etc. */
   }
 
-  StartStatisticsThread();
+  if (!StartStatisticsThread()) {
+    BErrNo be;
+    Emsg1(M_WARNING, 0,
```
pstorz, line 358:
```suggestion
    Emsg1(M_ERROR_TERM, 0,
```

pstorz (changes requested):
Good work, however I would use M_ERROR_TERM instead of M_WARNING if the statistics thread is required but cannot be started. This probably is a serious problem.

Other than that, the last commit message could be reformatted as it has a single "s" in one line that needs to be removed.

Thank you!

core/src/dird/dird.cc
alaaeddineelamri: This way, the daemons will always shutdown, since the default behavior is now to not start.
I changed the message to be displayed from within the `StartStatisticsThread()`

pstorz (approved):
Good work!

PR #1056, opened 2022-01-25
consolidate: purge jobids without files before starting the consolidation

pstorz:

The consolidation job now purges jobs without files from the jobid list before starting the consolidation. This makes sure that jobs without are cleaned up before and consolidation of jobs that do not have any files is never tried.

Also, the current postgresql query code reacts to an PGRES_FATAL_ERROR either by exiting the daemon (if exit on fatal is set) or by simply returning a normal error returning the error string from the database.

#### Please check

##### General

- [x] PR name is meaningful

##### Source code quality

##### Tests

- [x] Decision taken that a system- or unittest is required (if not, then remove this paragraph)
- [x] The decision towards a systemtest is reasonable compared to a unittest
- [x] Output of the test leads quickly to the origin of the fault

### DIFF
--- a/core/src/cats/postgresql.cc
+++ b/core/src/cats/postgresql.cc
@@ -620,8 +620,10 @@ retry_query:
       retval = true;
       break;
     case PGRES_FATAL_ERROR:
-      Dmsg1(50, "Result status fatal: %s\n", query);
-      if (exit_on_fatal_) { Emsg0(M_ERROR_TERM, 0, "Fatal database error\n"); }
+      Dmsg1(50, "Result status fatal: %s, %s\n", query, sql_strerror());
+      if (exit_on_fatal_) {
+        Emsg1(M_ERROR_TERM, 0, "Fatal database error: %s\n", sql_strerror());
+      }
 
       if (try_reconnect_ && !transaction_) {
         /*
--- a/core/src/cats/sql_get.cc
+++ b/core/src/cats/sql_get.cc
@@ -1293,6 +1293,27 @@ bool BareosDb::GetUsedBaseJobids(JobControlRecord* jcr,
   return SqlQueryWithHandler(query.c_str(), DbListHandler, result);
 }
 
+/*
+ * Remove the jobs that have JobFiles == 0 from the supplied jobid list
+ * and return them in another list
+ */
+db_list_ctx BareosDb::FilterZeroFileJobs(db_list_ctx& jobids)
+{
+  std::string query{"SELECT JobId FROM Job WHERE JobFiles = 0 AND JobId IN ("};
+  query += jobids.Join(",") + ")";
+
+  db_list_ctx zero_file_jobs;
+  if (!SqlQueryWithHandler(query.c_str(), DbListHandler, &zero_file_jobs)) {
+    throw new BareosSqlError(sql_strerror());
+  }
+  for (auto& remove_jobid : zero_file_jobs) {
+    jobids.erase(std::remove(jobids.begin(), jobids.end(), remove_jobid),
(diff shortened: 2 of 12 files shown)


core/src/cats/postgresql.cc
```diff
@@ -634,7 +634,7 @@ bool BareosDbPostgresql::SqlQueryWithoutHandler(const char* query, int flags)
       retval = true;
       break;
     case PGRES_FATAL_ERROR:
-      Dmsg1(50, "Result status fatal: %s\n", query);
+      Dmsg1(50, "Result status fatal: %s, %s\n", query, sql_strerror());
       if (exit_on_fatal_) { Emsg0(M_ERROR_TERM, 0, "Fatal database error\n"); }
```
bruno-at-bareos, line 638: Quick question : shouldn't we also print the root cause if we exit_on_fatal__ so the exit cause is obvious even without debug level ?

bruno-at-bareos (commented):
Answer the question about sending the code in case of exit on fatal.
Changelog Commit needed.

core/src/cats/postgresql.cc
pstorz: Makes sense, good idea!

bruno-at-bareos (commented):
I let one suggestion open.

bruno-at-bareos (commented):
Let see if last modification build and tests pass.

bruno-at-bareos (approved):
Everything reviewed, tested, and build correctly.
Nice work.

core/src/cats/sql_get.cc
```diff
@@ -1293,6 +1293,42 @@ bool BareosDb::GetUsedBaseJobids(JobControlRecord* jcr,
...
+ *  - The job never had any files
+ *  - The job files were purged or pruned
+ */
+bool BareosDb::DeleteZeroFileJobsinList(JobControlRecord* jcr,
+                                        JobDbRecord* jr,
+                                        db_list_ctx* jobids)
+{
+  db_list_ctx deleted_jobids;
+  std::string query{
+      "DELETE FROM Job where (JobFiles=0 OR PurgedFiles=1) AND JobID in ("};
```
arogge, line 1307: do we really want to implicitly purge jobs when their files were purged?
I don't think that's right.

arogge: also, just doing an SQL `DELETE` will potentially leave stale records in tables JobMedia, Log, basefiles, PathVisibility, NDMPJobEnvironment, JobStats and RestoreObject.

core/src/cats/sql_get.cc
```diff
@@ -1293,6 +1293,42 @@ bool BareosDb::GetUsedBaseJobids(JobControlRecord* jcr,
...
+  query += jobids->GetAsString().c_str();
+  query += ") RETURNING *";
+  Jmsg(jcr, M_INFO, 0,
+       "Checking for jobids that contain no files in jobid range %s\n",
+       jobids->GetAsString().c_str());
+
+  if (!SqlQueryWithHandler(query.c_str(), DbDeleteListHandler,
+                           &deleted_jobids)) {
+    Jmsg(jcr, M_ERROR, 0,
+         "Deleting jobids with no files from list %s failed : ERR=%s\n",
```
arogge, line 1317: shouldn't it be "Purging" instead of "Deleting"?

core/src/cats/sql_get.cc
arogge: I guess we need to determine the list and then do whatever"delete jobid=XXX" would be doing.

arogge (dismissed):

PR #1099, opened 2022-03-03
dird: director tries to connect to client when connection is disabled

alaaeddineelamri:
### Description

When `Connection From Director To Client = no`, the director tries anyway to connect but fails. Connection should not be even tried in the first place. This PR fixes the issue.

#### Please check

##### General
- [ ] PR name is meaningful
- [ ] Separate commit for this PR in the CHANGELOG.md, PR number referenced is same
- [ ] If backport: add original PR number and target branch at top of this file: **Backport of PR#000 to bareos-2x**

##### Source code quality

- [ ] `git status` should not report modifications in the source tree after building and testing

##### Tests

- [ ] Decision taken that a system- or unittest is required (if not, then remove this paragraph)
- [ ] The decision towards a systemtest is reasonable compared to a unittest
- [ ] Output of the test leads quickly to the origin of the fault

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,7 @@ and since Bareos version 20 this project adheres to [Semantic Versioning](https:
 - dir: fix crash when there are no jobs to consolidate [PR #1131]
 - webui: get volume and pool params from query instead of route [PR #1139]
 - FreeBSD packages: add missing ddl/update 2171_2192 and 2192_2210 files [PR #1147]
+- Fix director connects to client while `Connection From Director To Client` is disabled. [PR #1099]
 
 ### Changed
 - Qmsg: in case of syslog logging use adapted log priority instead of always LOG_ERR [PR #1134]
--- a/core/src/dird/fd_cmds.cc
+++ b/core/src/dird/fd_cmds.cc
@@ -127,7 +127,7 @@ static bool connect_outbound_to_file_daemon(JobControlRecord* jcr,
   utime_t heart_beat;
 
   if (!IsConnectingToClientAllowed(jcr)) {
-    Dmsg1(120, "connecting to client \"%s\" is not allowed.\n",
+    Dmsg1(120, "Connection to client \"%s\" is not allowed.\n",
           jcr->impl->res.client->resource_name_);
     return false;
   }
@@ -227,6 +227,11 @@ bool ConnectToFileDaemon(JobControlRecord* jcr,
                          bool verbose,
                          UaContext* ua)
 {
+  if (!IsConnectingToClientAllowed(jcr) && !IsConnectFromClientAllowed(jcr)) {
+    Emsg1(M_WARNING, 0, _("Connecting to %s is not allowed.\n"),
+          jcr->impl->res.client->resource_name_);
+    return false;
+  }
   bool success = false;
   bool tcp_connect_failed = false;
   int connect_tries
@@ -1328,8 +1333,7 @@ void* HandleFiledConnection(ConnectionPool* connections,
 
   if (!IsConnectFromClientAllowed(client_resource)) {
     Emsg1(M_WARNING, 0,
-          "Client \"%s\" tries to connect, "
-          "but does not have the required permission.\n",
+          "Connection from client \"%s\" to director is not allowed.\n",
           client_name);
     return NULL;
   }
--- a/systemtests/scripts/functions
+++ b/systemtests/scripts/functions
@@ -977,6 +977,33 @@ case $1 in
 esac
 }
 
+expect_grep()
+{
+
+local usage="Usage: ${FUNCNAME[0]} <expected_result> <log_file> [error message]\n    Find expected_result in log_file. Optionally display custom error_message if not found."
+local expected_result="$1"
+local target_file="$2"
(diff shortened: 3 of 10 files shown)


CHANGELOG.md
```diff
@@ -21,6 +21,7 @@ and since Bareos version 20 this project adheres to [Semantic Versioning](https:
 - Fix for wrong update message when updating all volumes from all pools with no existing volumes [PR #1015]
 - Fix context confusion in Director's Python plugins [PR #1047]
 - Fix several cases of undefined behaviour, memory corruption and memory leaks [PR #1060]
+- Fix case when director tries to connect to a client when it is not allowed to do so. [PR #1099]
```
pstorz, line 24:
```suggestion
- Fix director connects to client while `Connection From Director To Client` is disabled. [PR #1099]
```

core/src/dird/fd_cmds.cc
```diff
@@ -127,7 +127,7 @@ static bool connect_outbound_to_file_daemon(JobControlRecord* jcr,
   utime_t heart_beat;
 
   if (!IsConnectingToClientAllowed(jcr)) {
-    Dmsg1(120, "connecting to client \"%s\" is not allowed.\n",
+    Dmsg1(120, "Connection from director to client \"%s\" is not allowed.\n",
```
pstorz, line 130:
```suggestion
    Dmsg1(120, "Connecting to client \"%s\" is not allowed.\n",
```

pstorz: I would suggest to keep the wording unified (You added "Connecting to %s is not allowed" in line 231, so I would suggest to call it like this here also. Or we would need to update all other occurrences to contain the word "director" also.

core/src/dird/fd_cmds.cc
```diff
@@ -227,6 +227,11 @@ bool ConnectToFileDaemon(JobControlRecord* jcr,
                          bool verbose,
                          UaContext* ua)
 {
+  if (!IsConnectingToClientAllowed(jcr) && !IsConnectFromClientAllowed(jcr)) {
+    Jmsg(jcr, M_ERROR, 0, _("\nConnecting to %s is not allowed.\n"),
```
pstorz, line 231:
```suggestion
    Jmsg(jcr, M_ERROR, 0, _("Connecting to %s is not allowed.\n"),
```

pstorz: Jmsgs never start with a \n.

pstorz (changes requested):
I think it would make sense to create a gtest instead of a system test.
We only need to check if the connection request is denied when the configuration is configured correctly.

core/src/dird/fd_cmds.cc
alaaeddineelamri: That's the only occurrence of `IsConnectingToClientAllowed(jcr)` where there is a message returned. Adding a `from director to client` makes sense in my opinion to specify the exact reason on that occasion. But `IsConnectFromClientAllowed(jcr)` has a few occurrences that I can easily update.

For line 231, I did not add a `from` since both types of connection would be closed. If the check is verified, there is no way to connect to the client at all.

pstorz: The "from director" is not needed as dmesg log entries are prefixed witht he daemon name that logs.

pstorz (changes requested):

systemtests/scripts/functions
```diff
@@ -977,6 +977,33 @@ case $1 in
 esac
 }
 
+expect_grep()
+{
+
+local usage="Usage: ${FUNCNAME[0]} <expected_result> <log_file> [error message]\n    Find expected_result in log_file. Optionnaly display custom error_message if not found."
```
pstorz, line 983:
```suggestion
local usage="Usage: ${FUNCNAME[0]} <expected_result> <log_file> [error message]\n    Find expected_result in log_file. Optionally display custom error_message if not found."
```

systemtests/scripts/functions
```diff
@@ -977,6 +977,33 @@ case $1 in
...
+if [ ! -f ${target_file} ]; then
+    echo "\"${target_file}\" does not exist or is not a file."
+    exit 1
+fi
+
+if ! grep -q "${expected_result}" "${target_file}"; then
+  echo "Expected line \"${expected_result}\" was not found in log file \"${target_file}\"." >&2
+  estat=1
+  if [ "${error_message}" != "" ]; then
+    echo "-->${error_message}" >&2
```
pstorz, line 1002: If running in jenkins, we probably want to print out the whole log file in the error case to be able to understand what is going on. I think we can detect if running in jenkins by checking if JENKINS_HOME is set.

pstorz (changes requested):
I think printing out the logfiles in case of error when running inside of jenkins will help us to find problems faster.

systemtests/scripts/functions
alaaeddineelamri: Everything is usually displayed on Jenkins, so adding the content of the file would be duplicating the output even more. 
Adding the expected line at least would help where to look for in the output.

pstorz (approved):

PR #1126, opened 2022-03-15
cats:lib: removed comma formatting from jobId attribute in list jobs output.

HediBenFraj:
#### Description 

`List jobs` shows `jobId` with comma formatting. 
This PR fixes the issue by removing the commas from `jobid` attribute in the command output".

#### Please check

##### General
- [x] PR name is meaningful
- [x] If backport: add original PR number and target branch at top of this file: **Backport of PR#000 to bareos-2x**

##### Source code quality

##### Tests

- [ ] Decision taken that a system- or unittest is required (if not, then remove this paragraph)
- [ ] The decision towards a systemtest is reasonable compared to a unittest
- [ ] Output of the test leads quickly to the origin of the fault

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -87,6 +87,7 @@ and since Bareos version 20 this project adheres to [Semantic Versioning](https:
 - cats: management scripts remove db_driver support [PR #1081]
 - bconsole: multiple identical successive commands are only added to history once [PR #1162]
 - build: Now use solaris 11.4-11.4.42.0.0.111.0 [PR #1189]
+- bconsole: removed commas from jobid attribute in list jobs and llist jobs outputs [PR #1126]
 
 ### Deprecated
 - make_catalog_backup.pl is now a shell wrapper script which will be removed in version 23.
@@ -114,6 +115,7 @@ and since Bareos version 20 this project adheres to [Semantic Versioning](https:
 - clarifies Sphinx bareos-extension parallel_read_safe status to False [PR #1037]
 - fix incorrect link in contrib PythonFdPlugin [BUG #1450] [PR #1065]
 
+[PR #698]: https://github.com/bareos/bareos/pull/698
 [PR #768]: https://github.com/bareos/bareos/pull/768
 [PR #1010]: https://github.com/bareos/bareos/pull/1010
 [PR #1013]: https://github.com/bareos/bareos/pull/1013
@@ -162,6 +164,7 @@ and since Bareos version 20 this project adheres to [Semantic Versioning](https:
 [PR #1115]: https://github.com/bareos/bareos/pull/1115
 [PR #1120]: https://github.com/bareos/bareos/pull/1120
 [PR #1122]: https://github.com/bareos/bareos/pull/1122
+[PR #1126]: https://github.com/bareos/bareos/pull/1126
 [PR #1127]: https://github.com/bareos/bareos/pull/1127
 [PR #1131]: https://github.com/bareos/bareos/pull/1131
 [PR #1134]: https://github.com/bareos/bareos/pull/1134
@@ -176,6 +179,7 @@ and since Bareos version 20 this project adheres to [Semantic Versioning](https:
 [PR #1154]: https://github.com/bareos/bareos/pull/1154
 [PR #1155]: https://github.com/bareos/bareos/pull/1155
 [PR #1159]: https://github.com/bareos/bareos/pull/1159
+[PR #1162]: https://github.com/bareos/bareos/pull/1162
 [PR #1163]: https://github.com/bareos/bareos/pull/1163
 [PR #1169]: https://github.com/bareos/bareos/pull/1169
 [PR #1170]: https://github.com/bareos/bareos/pull/1170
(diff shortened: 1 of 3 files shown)


core/src/lib/edit.cc
```diff
@@ -625,6 +625,29 @@ char* add_commas(char* val, char* buf)
...
+  q = p + nc;
+  *q-- = *p--;
+  for (; nc; nc--) {
+    for (i = 0; i < 3; i++) { *q-- = *p--; }
+    *q-- = ' ';
+  }
+
+  return buf;
+}
+
```
arogge, line 650: Is duplicating the code the best way to do this or do you think there is a better way to do solve this?
Do you see any room for improvement in the function declaration itself (i.e. parameter and return types)?

HediBenFraj: > Is duplicating the code the best way to do this or do you think there is a better way to do solve this? Do you see any room for improvement in the function declaration itself (i.e. parameter and return types)?

yes I can do that, and that's much better I think, there is a slight issue since I'll need to maybe change the function calls in everyplace for add_commas or i can make it so default function call formats with commas.

HediBenFraj: > Is duplicating the code the best way to do this or do you think there is a better way to do solve this? Do you see any room for improvement in the function declaration itself (i.e. parameter and return types)?

thankfully there weren't a lot of function calls to 'add_commas'  so I changed the function to add_separator

core/src/cats/sql.cc
```diff
@@ -566,7 +566,7 @@ int BareosDb::ListResult(void* vctx, int nb_col, char** row)
           value.bsprintf(" %-*s |", max_len, "NULL");
         } else if (SqlFieldIsNumeric(field->type) && !jcr->gui
                    && IsAnInteger(row[i])) {
-          value.bsprintf(" %*s |", max_len, add_commas(row[i], ewc));
+          value.bsprintf(" %*s |", max_len, add_separator(row[i], ewc,','));
```
arogge, line 569: There's a space missing.
We have an automated tool `devtools/bareos-check-sources` for things like that. Take a look at the documentation in `devtools/pip-tools/README.md` on how to get it running.

```suggestion
          value.bsprintf(" %*s |", max_len, add_separator(row[i], ewc,','));
```

bruno-at-bareos:
Just a question, but in our backlog, from what I read : the initial idea was to be able to copy and paste jobids so I don't see a point to replace comma by space ?

HediBenFraj:
> Just a question, but in our backlog, from what I read : the initial idea was to be able to copy and paste jobids so I don't see a point to replace comma by space ?

yes it was suggested to remove the commas: they make it hard to copy and paste jobIds to use them in other commands (will require to manually remove the commas, I think ) 
we can also remove the separation altogether if necessary.

HediBenFraj:
@arogge By the way didn't need the trim method.
also i had to use the 
``` if (jobid){
remove_commas
}else{
add_commas
}
```

CHANGELOG.md
```diff
@@ -52,6 +52,7 @@ and since Bareos version 20 this project adheres to [Semantic Versioning](https:
 - webui: new login screen background and adapted logo to support Ukraine  [PR #1122]
 - console: multicolumn output: fill columns first [PR #1072]
 - cats: include only jobtypes in `list jobtotals` that write data to volumes [PR #1135]
+- bconsole: replaced commas in list jobs output with spaces for numbers over 999 [PR #1126]
```
arogge, line 55: that doesn't describe the change correctly anymore

core/src/cats/sql.cc
```diff
@@ -565,7 +565,8 @@ int BareosDb::ListResult(void* vctx, int nb_col, char** row)
         if (row[i] == NULL) {
           value.bsprintf(" %-*s |", max_len, "NULL");
         } else if (SqlFieldIsNumeric(field->type) && !jcr->gui
-                   && IsAnInteger(row[i])) {
```
arogge, line 568: the file's header doesn't have the right copyright years set.

core/src/lib/edit.h
```diff
@@ -1,7 +1,7 @@
 /*
    BAREOS® - Backup Archiving REcovery Open Sourced
 
-   Copyright (C) 2018-2020 Bareos GmbH & Co. KG
+   Copyright (C) 2018-2022 Bareos GmbH & Co. KG
```
arogge, line 4: as there's no other change in this file, we should also not change the header.

arogge (changes requested):
I noticed a formatting issue with the table. It looks like the calculation for the width of the column still calculates it as if the commas were still these.
Note the excess spaces in this example:
```
+------------+------------------+-----------+---------------------+----------+------+-------+----------+----------+-----------+
| jobid      | name             | client    | starttime           | duration | type | level | jobfiles | jobbytes | jobstatus |
+------------+------------------+-----------+---------------------+----------+------+-------+----------+----------+-----------+
|   10000000 | backup-bareos-fd | bareos-fd | 2022-03-28 11:46:48 | 00:00:01 | B    | F     |       73 |  144,145 | T         |
```

Other than that we may want to not add commas to `priorjobid` as that will also contain a `jobid` value. (AFAIR only `llist job` displays that field)

arogge (approved):

PR #1166, opened 2022-04-20
Add prompt for tracing

sdicke:
Changed processing of setdebug to ask for tracing wanted when interactive input is fetched.


joergsteffens:
Unfortunately, this code does not compile. I'll close this PR for now.

PR #1206, opened 2022-07-19
daemons: fix issue with group and users commandline options

alaaeddineelamri:
#### Description:
After the introduction of the new CLI options, the `-p`, `-u` and `-g` options can make daemons crash. This PR fixes the issue.
Also adding some helpful messages for the changed options.

#### Please check

##### General
- [x] PR name is meaningful
- [x] If backport: add original PR number and target branch at top of this file: **Backport of PR#000 to bareos-2x**

##### Source code quality

##### Tests


core/src/dird/dird.cc
```diff
@@ -301,11 +297,17 @@ int main(int argc, char* argv[])
...
 #if !defined(HAVE_WIN32)
-  if (!test_config && !foreground && pidfile_path) {
-    pidfile_fd = CreatePidFile("bareos-dir", pidfile_path);
+  if (!test_config && !foreground && !pidfile_path.empty()) {
+    pidfile_fd = CreatePidFile("bareos-dir", pidfile_path.c_str());
   }
 #endif
   // See if we want to drop privs.
+  char* uid = nullptr;
+  if (!user.empty()) { uid = user.data(); }
```
pstorz, line 306: Here, `str.data()` is used, while more down, `str.c_str()` is used. I think it would make sense to use the same function always, and as we do want to have a nullterminated c string, I would opt for using `c_str()` everywhere.

While nowadays both are identical, historically only the `c_str()` was guaranteed to have a \0 at the end.

pstorz (changes requested):
Looks good, only unified usage of c_str() would make sense.

core/src/dird/dird.cc
alaaeddineelamri: while I would agree about keeping it the same, the problem goes back to the type returned. 
Each situation demands a different type that can only be fulfilled by either one of the two functions. 
For this specific part, we need a `char*` to give to `uid` that can be only obtained with `.data()` while later in the code, we need a `const char*` that can only be obtained with `.c_str()`

pstorz: Ah, I see. That of course makes sense then. My checks for the difference between the two functions told me that  `The returned array is null-terminated, that is, data() and [c_str()](https://en.cppreference.com/w/cpp/string/basic_string/c_str) perform the same function.`

systemtests/tests/commandline-options/testrunner
```diff
@@ -0,0 +1,107 @@
...
+#Testing the creation and deletion of pid files with -p
+pid_file="${tmp}"/pid_file.out
+
+# Director
+
+${rscripts}/bareos-ctl-dir start "-p ${pid_file} -d 100" && sleep 5
+
+if [ ! -s ${pid_file} ]; then
+    echo "Pid file was not written."
+    estat=1
```
pstorz, line 42: Setting the estat to an individual value for each error could serve to quicker understand what exactly failed

core/src/lib/cli.cc
```diff
@@ -142,12 +143,43 @@ void InitCLIApp(CLI::App& app, std::string description, int fsfyear)
...
+}
+
+void AddDeprecatedExportOptionsHelp(CLI::App& app)
+{
+  app.add_option(
+         "-x",
+         [&app](std::vector<std::string> vals) {
+           app.failure_message(CLI::FailureMessage::simple);
+           throw CLI::ParseError(
+               "The -xc and -xs options have changed. Use --xc and --xs as "
```
pstorz, line 156: I would like to suggest the following format for the syntax change hints:
```
The -dt option has changed.                                                        
Use --dt as given in the help.                                                     
                       
Run with --help for more information.
```
This way, the old and new syntax is printed one below the other and it is easier to see the difference.

pstorz (changes requested):
Good work, Please see the comments.

pstorz (approved):

PR #1249, opened 2022-09-06
dird: fix tls reload crash

pstorz:
The director reload logic was not aware of the fact that the tls-psk mechanism's callback also accesses the configuration resources.

The tls_openssl_private objects now also have a shared_ptr to the configuration resources to make sure that these are not freed between the tls-psk callback is setup and the callback is really called.

The following patch can be used to trigger the problem when built with sanitizers:
```
diff --git a/core/src/lib/tls_openssl_private.cc b/core/src/lib/tls_openssl_private.cc
index 3b47510a2..ece45fa6e 100644
--- a/core/src/lib/tls_openssl_private.cc
+++ b/core/src/lib/tls_openssl_private.cc
@@ -468,7 +468,7 @@ unsigned int TlsOpenSslPrivate::psk_server_cb(SSL* ssl,
   Dmsg1(100, "psk_server_cb. identitiy: %s.\n", lst.JoinReadable().c_str());
 
   std::string configured_psk;
-
+  Bmicrosleep(5, 0);
   ConfigurationParser* config
       = static_cast<ConfigurationParser*>(SSL_CTX_get_ex_data(
           openssl_ctx,

```
- Reload the config every second e.g. with
            `kill -HUP $(pidof bareos_dir-reload)`
- start bconsole

#### Please check

##### General
- [x] PR name is meaningful

##### Source code quality

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,7 @@ and since Bareos version 20 this project adheres to [Semantic Versioning](https:
 - build: run a build and test with sanitizers enabled [PR #1244]
 
 ### Fixed
+- dird: fix possible crash in tls context on configuration reload [PR #1249]
 - dird: RunScript fixes [PR #1217]
   - fix show command output for RunScript RunsOnClient
   - fix show verbose for RunScripts


core/src/lib/tls_openssl.cc
```diff
@@ -75,7 +75,11 @@ void TlsOpenSsl::SetTlsPskServerContext(ConfigurationParser* config)
   } else if (!config) {
     Dmsg0(50, "Could not prepare TLS_PSK SERVER callback (no config)\n");
   } else {
-    Dmsg0(50, "Preparing TLS_PSK SERVER callback\n");
+    // store a shared pointer to the resources table currently used in the
+    // private data so that it is freed when the TLS Private Context is freed
```
arogge, line 79:
```suggestion
    // keep a shared_ptr to the current config, so a reload won't
    // free the memory we're going to use in the private context
```

core/src/lib/tls_openssl.cc
```diff
@@ -75,7 +75,11 @@ void TlsOpenSsl::SetTlsPskServerContext(ConfigurationParser* config)
...
+    d_->config_table_ = config->GetResourcesTablePointer();
+    Dmsg1(50, "Preparing TLS_PSK SERVER callback, config table pointer is %p\n",
+          d_->config_table_->configuration_resources_);
```
arogge, line 82: Do you really think we will ever need that for debugging? If so, please increase the loglevel. Otherwise, remove the Dmsg().

core/src/lib/parse_conf.cc
```diff
@@ -507,6 +503,13 @@ ConfigurationParser::BackupResourceTable()
   return backup_table;
 }
 
+std::shared_ptr<ConfigResourcesContainer>
+ConfigurationParser::GetResourcesTablePointer()
```
arogge, line 507: It is `BackupResourceTable()` and `RestoreResourceTable()` so it should be singular here, too.
The function does not actually return a pointer, so `...Pointer()` feels misleading. Maybe `...Handle()` is better or maybe drop the suffix completely and just call it `GetResourceTable()`.

core/src/lib/tls_openssl_private.cc
```diff
@@ -61,6 +61,7 @@ TlsOpenSslPrivate::TlsOpenSslPrivate()
     , pem_callback_(nullptr)
     , pem_userdata_(nullptr)
     , verify_peer_(false)
+    , config_table_(nullptr)
```
arogge, line 64: we could (and probably should) move all these zero-initializations to the class member's definition in the header.
This is less error-prone as you then cannot forget to initialize a member when you write another constructor. Also when reading the code you won't need to wonder if this member is guaranteed to be zero-initialized or not.

arogge (commented):
Looks pretty good, I just have some stylistic comments.

core/src/lib/parse_conf.cc
pstorz: Makes sense, done.

core/src/lib/tls_openssl.cc
pstorz: Agreed, removed the Dmsg1() call

core/src/lib/tls_openssl_private.cc
pstorz: agreed and done.

arogge (approved):

PR #1262, opened 2022-09-16
dird: cats: remove copy and migration jobs with no data from catalog

alaaeddineelamri:
#### Description

Copy and migration jobs always leave empty jobs behind in the catalog even after volume pruning. Those jobs can only be removed manually or with job pruning. 
This PR makes sure:
* the "administrative job" or "migration/copy control job" that spawns the migration/copy jobs is deleted upon completion,
* the copy jobs are deleted when actual related copies are upgraded to backups,
* the migration/copy jobs are deleted when their new migrated/copied jobs are pruned

#### Please check

##### General
- [x] PR name is meaningful
- [x] Check backport line
- [x] Separate commit for CHANGELOG.md ("update CHANGELOG.md"). The PR number is correct.

##### Source code quality

##### Tests


core/src/cats/sql_update.cc
```diff
@@ -563,6 +563,14 @@ void BareosDb::UpgradeCopies(const char* jobids)
...
+  // remove the worker job that copied it
+  Mmsg(query,
+       "DELETE FROM Job "
+       "WHERE Type='%c' "
+       "AND priorjobid IN ( SELECT JobId FROM cpy_tmp )",
+       JT_COPY);
+  SqlQuery(query.c_str());
+
   Mmsg(query, "DROP TABLE cpy_tmp");
   SqlQuery(query.c_str());
```
pstorz, line 575:
```suggestion
  // remove the worker job that copied it
  SqlQuery(
       "DELETE FROM Job "
       "WHERE Type='JT_COPY' "
       "AND priorjobid IN ( SELECT JobId FROM cpy_tmp )");
```

pstorz, line 575: Also here we have a fixed string, which could be directly given to SqlQuery, right?

core/src/dird/migrate.cc
```diff
@@ -1682,18 +1666,26 @@ void MigrationCleanup(JobControlRecord* jcr, int TermCode)
   PoolMem query(PM_MESSAGE);
 
   Dmsg2(100, "Enter migrate_cleanup %d %c\n", TermCode, TermCode);
-  UpdateJobEnd(jcr, TermCode);
+
```
pstorz, line 1669: Why was UpdateJobEnd moved down?

core/src/dird/ua_prune.cc
```diff
@@ -868,6 +868,25 @@ int GetPruneListForVolume(UaContext* ua,
...
+    for (auto jobid : prune_list) { copy_migrate_jobs_list << jobid; }
+    Mmsg(query,
+         "SELECT JobId from Job "
+         "WHERE type in ('%c','%c') AND priorjobid in (%s)",
+         JT_COPY, JT_MIGRATE, copy_migrate_jobs_list.Join(',').c_str());
+
+    if (!ua->db->SqlQuery(query.c_str(), FileDeleteHandler,
+                          static_cast<void*>(&prune_list))) {
+      if (ua->verbose) { ua->ErrorMsg("%s", ua->db->strerror()); }
+      Dmsg0(050, "Count failed\n");
```
pstorz, line 885: Is this message correct? Why "Count failed"?

docs/manuals/source/TasksAndConcepts/MigrationAndCopy.rst
```diff
@@ -39,7 +39,7 @@ A migration job can be started manually or from a Schedule, like a backup job. I
 
 Normally three jobs are involved during a migration:
 
--  The Migration control Job which starts the migration child Jobs.
+-  The Migration control Job which starts the migration child Jobs (removed upon completion).
```
pstorz, line 42: Is the job also removed if completes with error?

pstorz (changes requested):
Very good work! Just some small remarks

Suggestion for commit  descripton of b578a1e84c10228ac55cc96224a85618706c8529 :

"docs: MigrationAndCopy.rst: clarify that control job gets removed upon completition"

docs/manuals/source/TasksAndConcepts/MigrationAndCopy.rst
pstorz: Also, do we want to add the information that the "migration child jobs" will be removed when the original jobid is removed from the database?

alaaeddineelamri: when the job enters `MigrationCleanup(...)` in migrate.cc:1659 at the end of the job, it will be deleted.
A migration cleanup is part of the natural cycle of the job, so whether successful or failed, it gets called, and thus the control job gets removed.

core/src/cats/sql_update.cc
alaaeddineelamri: I guess we can do it here for `DROP TABLE cpy_tmp`

alaaeddineelamri: most of it is fixed, but `JT_COPY` is not. If for any reason in the future we decide to change what `JT_COPY` refers to, the result would be wrong

core/src/dird/migrate.cc
alaaeddineelamri: it is technically the same position, just moved into the `if` statement. 
It used to be executed for both jobs. It now gets executed only for the new migrated job, while the control one gets deleted in the `else` statement, so no need for it to be updated.

docs/manuals/source/TasksAndConcepts/MigrationAndCopy.rst
alaaeddineelamri: > Also, do we want to add the information that the "migration child jobs" will be removed when the original jobid is removed from the database?

I'll see where I can fit that information in the docs

core/src/cats/sql_update.cc
pstorz: Agreed

pstorz (commented):
Good work. Reading the docs about the migration jobs I was not happy with how everything was described so I updated the docs. Could you please check if what I wrote makes sense?

pstorz (approved):

PR #1272, opened 2022-10-06
sd: refactor the SD's backend interface

arogge:
This changes the SD Backend and Device interface in a way that is supposed to make it easier to add new backends.

* file device is now also built and loaded dynamically
* `Device Type` setting will accept any string
* presence of a configured storage backend checked after loading the configuration
* systemtests fail early if a daemon config doesn't pass a test with `-t`
* Remove `IsFile()` and `IsFifo()` from `Device` (in favor of new checks)
* Increase CMake required version to 3.17
* static linking of chunked, droplet and gentape into the backends that use them

#### Please check

##### General
- [x] PR name is meaningful
- [x] Check backport line
- [x] Separate commit for CHANGELOG.md ("update CHANGELOG.md"). The PR number is correct.
- [x] Merge corresponding PR in git.bareos.com/ci-scripts

##### Source code quality


core/src/lib/plugin_registry.h
```diff
@@ -0,0 +1,75 @@
+/*
```
pstorz, line 1: Maybe better call it "backend registry" to avoid confusion with fd, sd and dir plugins?

pstorz (changes requested):
_Do we still want to ship libbareossd-chunked.so and libbareossd-gentape.so?_
**No, we want to link chunked and gentape statically into the backends (i.e. droplet and tape) instead of shipping them as shared libraries.**

core/src/stored/device_resource.h
```diff
@@ -43,7 +43,7 @@ class DeviceResource : public BareosResource {
   char* changer_command;        /**< Changer command  -- external program */
   char* alert_command;          /**< Alert command -- external program */
   char* spool_directory;        /**< Spool file directory */
-  DeviceType dev_type;          /**< device type */
+  std::string dev_type;         /**< device type */
```
arogge, line 46:
```suggestion
  std::string device_type;
```

core/src/lib/plugin_registry.h
arogge: renamed it to ImplementationFactory

core/CMakeLists.txt
```diff
@@ -30,6 +30,7 @@ endif()
 
 cmake_policy(SET CMP0054 NEW)
 cmake_policy(SET CMP0057 NEW)
+cmake_policy(SET CMP0079 NEW)
```
pstorz, line 33:
```suggestion
cmake_policy(SET CMP0054 NEW) # Only interpret if() arguments as variables or keywords when unquoted.
cmake_policy(SET CMP0057 NEW) # Support new if() IN_LIST operator.
cmake_policy(SET CMP0079 NEW) # target_link_libraries() allows use with targets in other directories.
```

pstorz: Also I think we should request  `cmake_minimum_required(VERSION 3.17)`

pstorz (changes requested):
Just minimal changes, please check my comments.

core/CMakeLists.txt
arogge: we do that in the toplevel CMakeLists.txt

arogge: I unified cmake_policy to require 3.17 and enable all policies up to and including 3.19.
We also require CMake 3.17 now and don't use `cmake_minimum_required()` everywhere anymore.

pstorz (approved):
Good work!

PR #1294, opened 2022-10-25
systemtests: make them usably with Windows binaries on Linux using Wine

joergsteffens:

Make the Windows binaries testable via systemtests on Linux using Wine.

#### Please check

##### General
- [ ] PR name is meaningful
- [ ] Check backport line
- [ ] Has https://git.bareos.com/bareos/CD-buildimages/pulls/12 been merged?
- [ ] Has https://git.bareos.com/bareos/CD/pulls/53/ been merged?

##### Source code quality

##### Tests


systemtests/CMakeLists.txt
```diff
@@ -74,34 +99,38 @@ message(STATUS "POSTGRES_BIN_PATH is ${POSTGRES_BIN_PATH}")
 
 find_program(PERL perl)
 
+set(SYSTEMTESTS_USE_WINE
+    ""
+    CACHE STRING "systemtests do use WINE to run Bareos"
+)
```
arogge, line 105: If you want a boolean cache variable, you can use option.
```suggestion
option(SYSTEMTESTS_USE_WINE
       "systemtests do use WINE to run Bareos"
       OFF
)
```

joergsteffens: Applied.

joergsteffens:
> That this PR alters more thant 600 `.in` files just to add a prefix to an exisiting variable is not a good solution. Probably the variables need to be prefixed themselves by cmake instead.

The problem is, that we need this variable in 2 variants:
* Unix path format (for inside cmake and shell scripts)
* Windows path format. When used as Bareos configuration, as the Daemon do run in Windows/Wine. In Bareos we can use the same path, but must prefix it with `@wine_drive_prefix@`.

If using `@wine_drive_prefix@` in the Bareos configuration files is not an option, I still have to create a second variable for every path variable used in the Bareos configuration. Example:

Instead of using
* `@archivedir@` and
* `@wine_drive_prefix@@archivedir@`
I could create a second variable to be used in the Bareos configuration:
* `@archivedir@` and
* `@bcfg_archivedir@` (= `@wine_drive_prefix@@archivedir@`)

Still, the same amount if `*.in` files must be modified. Do you think, the second variant is better? Would you agree to it?

joergsteffens:
> That this PR alters more thant 600 `.in` files just to add a prefix to an exisiting variable is not a good solution. Probably the variables need to be prefixed themselves by cmake instead.

The problem is, that we need this variable in 2 variants:
* Unix path format (for inside cmake and shell scripts)
* Windows path format. When used as Bareos configuration, as the Daemon do run in Windows/Wine. In Bareos we can use the same path, but must prefix it with `@wine_drive_prefix@`.

If using `@wine_drive_prefix@` in the Bareos configuration files is not an option, I still have to create a second variable for every path variable used in the Bareos configuration. Example:

Instead of using
* `@archivedir@` and
* `@wine_drive_prefix@@archivedir@`
I could create a second variable to be used in the Bareos configuration:
* `@archivedir@` and
* `@bcfg_archivedir@` (= `@wine_drive_prefix@@archivedir@`)

Still, the same amount if `*.in` files must be modified. Do you think, the second variant is better? Would you agree to it?

bruno-at-bareos:
Closing, a try will be make to run test in native builds.

PR #1311, opened 2022-11-11
dird: update JobStatus to DB in cases we wait for media

banuchka:
Update JobStatus as **wait for media** to DB in cases where job is waiting for valid media to write.
From my point of view if job is in wait for media status it can't be marked as Running job(right technically it is running job from director perspective but it will do nothing until the issue with valid media will be resolved).
   
##### General
- [ ] PR name is meaningful
- [ ] Check backport line

##### Source code quality

##### Tests

### DIFF
--- a/core/src/dird/getmsg.cc
+++ b/core/src/dird/getmsg.cc
@@ -83,6 +83,12 @@ static void SetJcrSdJobStatus(JobControlRecord* jcr, int SDJobStatus)
   }
   jcr->dir_impl->SDJobStatus = SDJobStatus;
 
+  jcr->setJobStatus(SDJobStatus);
+
+  if (!jcr->db->UpdateJobStartRecord(jcr, &jcr->impl->jr)) {
+    Jmsg(jcr, M_FATAL, 0, "%s", jcr->db->strerror());
+  }
+
   // Some SD Job status setting are propagated to the controlling Job.
   switch (jcr->dir_impl->SDJobStatus) {
     case JS_Incomplete:
--- a/core/src/dird/getmsg.cc
+++ b/core/src/dird/getmsg.cc
@@ -85,7 +85,7 @@ static void SetJcrSdJobStatus(JobControlRecord* jcr, int SDJobStatus)
 
   jcr->setJobStatus(SDJobStatus);
 
-  if (!jcr->db->UpdateJobStartRecord(jcr, &jcr->impl->jr)) {
+  if (!jcr->db->UpdateJobStartRecord(jcr, &jcr->dir_impl->jr)) {
     Jmsg(jcr, M_FATAL, 0, "%s", jcr->db->strerror());
   }
 


pstorz:
Hello,

thank you for your contribution. As the master `impl` pointer was altered in master, I just rebased and solved the conflict.

best regards,

Philipp

alaaeddineelamri (commented):
Thank you for your contribution!

Do you have a functional reason to update the status in the catalog? is a functionality misbehaving because of it? or maybe does this hinder your ability to follow up your jobs?

The `status director` command can already show the status of running jobs and tell you if the job is in a waiting status (and the reason for it) or not.

banuchka:
@alaaeddineelamri main reason for me here is an ability to use monitoring tools which are able to show correct information about each job status.
For me it is good to know about full cycle of a job. I need such information to fix any issues with volumes, retention policy etc for different pools which I have a lot.
I do known about `status dir` command. I've tried to use it for my purposes before.
And as I can see for example Bareos WebUI also rely on Job status information in DB/Catalog.

core/src/dird/getmsg.cc
```diff
@@ -83,6 +83,12 @@ static void SetJcrSdJobStatus(JobControlRecord* jcr, int SDJobStatus)
   }
   jcr->dir_impl->SDJobStatus = SDJobStatus;
 
+  jcr->setJobStatus(SDJobStatus);
```
arogge, line 86: TBH I'm not sure how you would replicate the SDJobStatus (and also the FDJobStatus) into the Director's job status correctly.
Just copying it -- like you did here -- may or may not be the right way.

arogge (changes requested):
Updating the status in the database is great. I think it will help us a lot. We should probably do that every time the status changes and not just when the SD sends a message.
However, as currently the status in the director is usually just 'R' while the job runs, we may need to review all currently used SQL queries.
Also, we would need a sane way to combine multiple statuses (i.e. Status A in SD and Status B in FD when current Director status is X will result in Director Status Y).

banuchka (commented):
ok for me

banuchka:
> There are probably three ways you could pursue:
> 
> * basically your proposal, but with a defined way of determining the overall status based on status of Director, SD and FD
>   
>   * probably pretty hard to do as there are a lot of places that rely on the current overall status and some that look at the catalog database.
probably but sounds a bit overpriced for such _simple feature_
> * add new columns to the job table and keep SD and FD status separately
>   
>   * may be a good idea, but if it doesn't make it into 22 (which is due in December) the next chance for a schema update is in Bareos 23.
not sure it is a right way here but it is only from my point of view
> * improve `status dir` to provide the information you're looking for in json-api mode
>   
>   * probably the easiest and least invasive method, but it won't reflect the status in the catalog DB.
doesnt work for me. Now it is possible to use `status dir` and it doesnt really matter with or without JSON output. Based on my experience with Bareos it is always cheaper to ask DB server than to bother DIR with bconsole especially from monitoring tools. 

Maybe there is a way when it is possible to change JobStatus(in Catalog) from job's SD status only for certain scenarios, for example:
- Job *Waiting for new media*
- Job *Waiting for media mount*

Because 2 scenarios above are the most recent cases when the Job technically not running(it doesnt do anything).
I can understand your point that it maybe not the best option to change Job status in Catalog for every intermediate state but anyway as I've seen debug output and log all sql queries with different scenarios, it doesnt look as bad as I may imagine.

arogge:
> Because 2 scenarios above are the most recent cases when the Job technically not running(it doesnt do anything). I can understand your point that it maybe not the best option to change Job status in Catalog for every intermediate state but anyway as I've seen debug output and log all sql queries with different scenarios, it doesnt look as bad as I may imagine.

I would be totally happy with updating the status in the catalog every time the status changes. Only doing it for one or two specific status changes is what actually bothers me. It would be much easier to understand, to document and a lot less surprising if it happens always (or never as it is right now).

What I'm also not happy with is unconditionally overwriting the Director's job status with the SD's job status.

Last but not least all the "list all running jobs" queries will need to consider the newly introduced status identifiers (i.e. the ones from the SD we didn't write to the Catalog previously), too. For example, when I apply your change and restart the director while a job is waiting for a volume, that job will not be marked "failed" in the catalog.

banuchka:
> Last but not least all the "list all running jobs" queries will need to consider the newly introduced status identifiers (i.e. the ones from the SD we didn't write to the Catalog previously), too. For example, when I apply your change and restart the director while a job is waiting for a volume, that job will not be marked "failed" in the catalog.

Exactly! I've just found such case and it is bother me a lot. Hope I will find and fix it for my specific reason.
Regarding each and every status change on SD I'm not sure what is better:
- update on every and each status change
- use only specific statuses which needs to be in DB, let say for not it is two of them: "m" and "M".

And I'm not sure that I need to change DIR job's status at all but in such scenario you are absolutely right because the rest of sql queries rely on R status in database.

What do you think if I only update job status in catalog but leave DIR job's status as it is. How bad such idea?

Thanks for you attention to current PR and for your time by the way!

arogge:
Hi,
sorry for keeping you waiting so long.
One of the reasons I kept you waiting so long is that I still don't have a feasible way to solve your requirement.
We discussed this topic in one of our developer meetings and nobody thought updating the job status in the Catalog is a good idea.

Changing the job status in the catalog and leaving it unchanged in the director itself is probably worse. It will break catalog queries just like the original implementation, but is also pretty surprising (as everybody expects this to be in sync).

I really don't have any advice how we can move forward here :(

arogge:
As there aren't any news for two weeks and I don't see how we can move forward, I'll close the PR. Feel free to reopen, to ping me or to bring something up on the users or devel mailing list.

PR #1280, opened 2022-10-17
webui: introduce subscription backup unit report

fbergkemper:
Introduces an additional view to the director module where a backup unit report can be retrieved and downloaded in JSON format.

#### Please check

- [x] Short description and the purpose of this PR is present _above this paragraph_
- [x] Your name is present in the AUTHORS file (optional)

If you have any questions or problems, please give a comment in the PR.

### Helpful documentation and best practices

- [Git Workflow](https://docs.bareos.org/DeveloperGuide/gitworkflow.html)
- [Automatic Sourcecode Formatting](https://docs.bareos.org/DeveloperGuide/generaldevel.html#automatic-sourcecode-formatting)
- [Check your commit messages](https://docs.bareos.org/DeveloperGuide/gitworkflow.html#commits)

##### General
- [x] PR name is meaningful
- [x] Purpose of the PR is understood
- [x] Commit descriptions are understandable and well formatted
- [x] Check backport line
- [x] Is the PR title usable as CHANGELOG entry?
- [x] Separate commit for CHANGELOG.md ("update CHANGELOG.md"). The PR number is correct.

##### Source code quality

- [x] Source code changes are understandable
- [x] Variable and function names are meaningful
- [x] Code comments are correct (logically and spelling)
- [ ] Required documentation changes are present and part of the PR
- [x] `bareos-check-sources --since-merge` does not report any problems

##### Tests

- [ ] Decision taken that a test is required (if not, then remove this paragraph)
- [ ] The choice of the type of test (unit test or systemtest) is reasonable
- [ ] Testname matches exactly what is being tested
- [ ] On a fail, output of the test leads quickly to the origin of the fault

PR #1312, opened 2022-11-15
status subscription: extend output

joergsteffens:
While working on https://github.com/bareos/bareos/pull/1280 some improvements to the `status subscription` command did come up and are now handled in this separate PR.

- improve the JSON output, reducing the depth of the structure and make it better understand- and parseable 
- add information Bareos version (also in case the calculation of units will change with other versions), current date and simple checksum.
- add ACL test.

#### Please check

##### General
- [x] PR name is meaningful
- [x] Check backport line
- ~Separate commit for CHANGELOG.md ("update CHANGELOG.md"). The PR number is correct.~

##### Source code quality

##### Tests

### DIFF
--- a/core/src/cats/sql_list.cc
+++ b/core/src/cats/sql_list.cc
@@ -46,7 +46,7 @@ bool BareosDb::ListSqlQuery(JobControlRecord* jcr,
                             const char* query,
                             OutputFormatter* sendit,
                             e_list_type type,
-                            bool verbose)
+                            const bool verbose)
 {
   return ListSqlQuery(jcr, query, sendit, type, "query", verbose);
 }
@@ -55,7 +55,7 @@ bool BareosDb::ListSqlQuery(JobControlRecord* jcr,
                             SQL_QUERY query,
                             OutputFormatter* sendit,
                             e_list_type type,
-                            bool verbose)
+                            const bool verbose)
 {
   return ListSqlQuery(jcr, query, sendit, type, "query", verbose);
 }
@@ -65,7 +65,8 @@ bool BareosDb::ListSqlQuery(JobControlRecord* jcr,
                             OutputFormatter* sendit,
                             e_list_type type,
                             const char* description,
-                            bool verbose)
+                            const bool verbose,
+                            const CollapseMode collapse)
 {
   DbLocker _{this};
 
@@ -75,9 +76,24 @@ bool BareosDb::ListSqlQuery(JobControlRecord* jcr,
     return false;
   }
 
-  sendit->ArrayStart(description);
-  ListResult(jcr, sendit, type);
-  sendit->ArrayEnd(description);
+  if ((collapse == CollapseMode::Collapse) && (SqlNumRows() > 1)) {
+    Mmsg(errmsg,
+         _("Query returned %d rows. In collapsed mode, only one row is "
+           "accepted.\n"),
+         SqlNumRows());
+    if (verbose) { sendit->Decoration(errmsg); }
+    return false;
+  }
+
+  if (collapse == CollapseMode::Collapse) {
+    sendit->ObjectStart(description);
+    ListResult(jcr, sendit, type);
+    sendit->ObjectEnd(description);
+  } else {
+    sendit->ArrayStart(description);
+    ListResult(jcr, sendit, type);
+    sendit->ArrayEnd(description);
+  }
   SqlFreeResult();
 
   return true;
@@ -88,10 +104,11 @@ bool BareosDb::ListSqlQuery(JobControlRecord* jcr,
                             OutputFormatter* sendit,
                             e_list_type type,
                             const char* description,
-                            bool verbose)
+                            const bool verbose,
+                            const CollapseMode collapse)
 {
   return ListSqlQuery(jcr, get_predefined_query(query), sendit, type,
-                      description, verbose);
+                      description, verbose, collapse);
 }
 
 void BareosDb::ListPoolRecords(JobControlRecord* jcr,
--- a/core/src/dird/ua_status.cc
+++ b/core/src/dird/ua_status.cc
@@ -515,6 +515,14 @@ static bool DoSubscriptionStatus(UaContext* ua)
   }
 
   if (kw_all || kw_detail) {
+    ua->send->ObjectKeyValue(
+        "version", "Bareos version: ", kBareosVersionStrings.Full, "%s");
+    ua->send->ObjectKeyValue("os", kBareosVersionStrings.GetOsInfo(),
+                             " (%s)\n");
+    ua->send->ObjectKeyValue("binary-info",
+                             "Binary info: ", kBareosVersionStrings.BinaryInfo,
+                             "%s\n");
+    ua->send->ObjectKeyValue("report-time", "Report time: ", now, "%s\n");
     ua->SendMsg(_("\nDetailed backup unit report:\n"));
     ua->db->ListSqlQuery(
         ua->jcr,
--- a/core/src/dird/ua_status.cc
+++ b/core/src/dird/ua_status.cc
@@ -478,6 +478,24 @@ static bool show_scheduled_preview(UaContext*,
   return true;
 }
 
+std::string get_subscription_status_checksum_source_text(UaContext* ua,
+                                                         const char* timestamp)
+{
+  const std::string salt("SECRETSALT");
+  PoolMem subscriptions(PM_MESSAGE);
+  PoolMem query(PM_MESSAGE);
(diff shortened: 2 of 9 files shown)


core/src/dird/ua_status.cc
```diff
@@ -513,46 +532,53 @@ static bool DoSubscriptionStatus(UaContext* ua)
...
+
   if (kw_all || kw_detail) {
+    ua->send->ObjectKeyValue(
+        "version", "Bareos version: ", kBareosVersionStrings.Full, "%s");
+    ua->send->ObjectKeyValue("os", kBareosVersionStrings.GetOsInfo(),
+                             " (%s)\n");
+    ua->send->ObjectKeyValue("binary-info",
+                             "Binary info: ", kBareosVersionStrings.BinaryInfo,
+                             "%s\n");
+    ua->send->ObjectKeyValue("date", "Date: ", now, "%s\n");
```
arogge, line 546: maybe we want that part really only when "all" was set?

core/src/dird/ua_status.cc
```diff
@@ -478,6 +478,24 @@ static bool show_scheduled_preview(UaContext*,
   return true;
 }
 
+std::string get_subscription_status_signature_source_text(UaContext* ua,
+                                                          const char* timestamp)
+{
+  std::string salt("SECRETSALT");
```
arogge, line 484:
```suggestion
  const std::string salt{"SECRETSALT"};
```

core/src/dird/ua_status.cc
```diff
@@ -478,6 +478,24 @@ static bool show_scheduled_preview(UaContext*,
...
+  PoolMem subscriptions(PM_MESSAGE);
+  PoolMem query(PM_MESSAGE);
+  OutputFormatter output_text
+      = OutputFormatter(pm_append, (void*)&subscriptions, nullptr, nullptr);
```
arogge, line 488: casts to `void*` are implicit and c-style casts are evil.
```suggestion
      = OutputFormatter(pm_append, &subscriptions, nullptr, nullptr);
```

core/src/dird/ua_status.cc
```diff
@@ -478,6 +478,24 @@ static bool show_scheduled_preview(UaContext*,
   return true;
 }
 
+std::string get_subscription_status_signature_source_text(UaContext* ua,
```
arogge, line 481: you're only using `ua->db`, so maybe consider `BareosDb*` instead of `UaContext*` or maybe `BareosDb&`.

core/src/dird/ua_status.cc
```diff
@@ -513,46 +532,53 @@ static bool DoSubscriptionStatus(UaContext* ua)
...
         query, BareosDb::SQL_QUERY::subscription_select_backup_unit_total_1,
         me->subscriptions);
-    ua->db->ListSqlQuery(ua->jcr, query.c_str(), ua->send, VERT_LIST, true);
-    ua->send->ObjectEnd("total-units-required");
+    ua->db->ListOneRowSqlQuery(ua->jcr, query.c_str(), ua->send, VERT_LIST,
+                               "total-units-required", true);
+    std::string signature_source
+        = get_subscription_status_signature_source_text(ua, now);
+    std::string hash = compute_hash(signature_source.c_str());
+    ua->send->ObjectKeyValue("signature", "Signature: ", hash.c_str(), "%s\n");
```
arogge, line 564: I really wouldn't call it a signature, it is just a hash or checksum, so we should probably name it as such.

core/src/dird/ua_status.cc
arogge, line 546: It is called date, but contains the current time.
Maybe something like "report-time" and "Report time:" is more reasonable.

arogge (changes requested):
Pretty good, but we can probably do better :)

core/src/dird/ua_status.cc
joergsteffens: Why? As "all" can produce confusing outputs (unknown section), currently we switched WebUI to show "status subscription detail" instead of "status subscription all" and I'd like to see this information in the WebUI report.

core/src/cats/sql_list.cc
```diff
@@ -94,6 +96,51 @@ bool BareosDb::ListSqlQuery(JobControlRecord* jcr,
                       description, verbose);
 }
 
+bool BareosDb::ListOneRowSqlQuery(JobControlRecord* jcr,
```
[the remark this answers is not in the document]
joergsteffens: Good point.

core/src/lib/crypto_openssl.cc
```diff
@@ -1463,6 +1466,33 @@ const char* crypto_digest_name(DIGEST* digest)
   return crypto_digest_name(digest->type);
 }
 
+std::string compute_hash(const std::string& unhashed,
+                         const std::string& digestname)
+{
+  std::stringstream string;
```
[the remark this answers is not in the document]
joergsteffens: Done.

core/src/lib/crypto_openssl.cc
```diff
@@ -1463,6 +1466,33 @@ const char* crypto_digest_name(DIGEST* digest)
...
+  const EVP_MD* type = EVP_get_digestbyname(digestname.c_str());
+  if (type) {
+    EVP_MD_CTX* context = EVP_MD_CTX_create();
+    if (context != NULL) {
+      if (EVP_DigestInit_ex(context, type, NULL)) {
+        if (EVP_DigestUpdate(context, unhashed.c_str(), unhashed.length())) {
+          unsigned char hash[EVP_MAX_MD_SIZE];
+          unsigned int lengthOfHash = 0;
+          if (EVP_DigestFinal_ex(context, hash, &lengthOfHash)) {
+            string << "{" << digestname << "}";
```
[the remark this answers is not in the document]
joergsteffens: Done.

core/src/lib/crypto_openssl.cc
```diff
@@ -1463,6 +1466,33 @@ const char* crypto_digest_name(DIGEST* digest)
...
+            string << std::hex << std::setw(2) << std::setfill('0');
+            for (unsigned int i = 0; i < lengthOfHash; ++i) {
+              string << (int)hash[i];
```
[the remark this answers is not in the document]
joergsteffens: Done.

core/src/dird/ua_status.cc
joergsteffens: done.

core/src/cats/sql_list.cc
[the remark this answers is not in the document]
joergsteffens: done.

core/src/dird/ua_status.cc
joergsteffens: I named it Checksum now.

core/src/dird/ua_status.cc
joergsteffens: done.

joergsteffens:
I addressed all mentioned suggestions. Without new feedback, the only I maybe can address is replacing `compute_hash` by the already existing function. Anyhow, I'll not have the time for this before next week. But maybe this is not even required, as this function also works on all our supported platforms.

core/src/dird/ua_status.cc
```diff
@@ -513,46 +532,58 @@ static bool DoSubscriptionStatus(UaContext* ua)
...
+    ua->db->ListSqlQuery(ua->jcr, query.c_str(), ua->send, VERT_LIST,
+                         "total-units-required", true,
+                         BareosDb::CollapseMode::Collapse);
+    std::string checksum_source
+        = get_subscription_status_checksum_source_text(ua, now);
+    std::optional<std::string> checksum = compute_hash(checksum_source.c_str());
+    if (checksum) {
+      ua->send->ObjectKeyValue("checksum",
+                               "Checksum: ", checksum.value().c_str(), "%s\n");
+    }
```
arogge, line 568: * i would probably have used `auto` instead of `std::optional<std::string>` here.
* `compute_hash()` takes a `const std::string&` as first parameter. You're passing the underlying `const char*` of your existing `std::string` to it, which will lead to an implicit conversion to std::string() which implies another copy. Just passing the `std::string` would be better
* there are two ways to get the value out of a `std::optional`: calling `value()` or `operator*`/`operator->`. With `value()` you get a check that there is a value in your optional (and an exception if there is none), while with the operators behaviour is undefined if there is no value. As you checked that there is a value, you can safely use the operators, simplifying `checksum.value().c_str()` to just `checksum->c_str()`.

arogge (approved):
I made a few observations, but no objections. Looks good to me now!

PR #1318, opened 2022-11-22
dird: config: console: prohibit PAM usage with user ACL and Profiles in consoles

alaaeddineelamri:
#### Description

This PR makes Bareos consider malformed, consoles whose configuration having simultaneously `use PAM authentication` and ACL commands, or a Profile having ACL.

#### Please check

##### General
- [x] PR name is meaningful
- [x] Check backport line

##### Source code quality

##### Tests

### DIFF
--- a/core/src/dird/reload.cc
+++ b/core/src/dird/reload.cc
@@ -186,15 +186,14 @@ bool CheckResources()
   BStringList consoles_with_auth_problems;
   foreach_res (console, R_CONSOLE) {
     if (console->use_pam_authentication_
-        && (console->user_acl.ACL_lists[0] || console->user_acl.profiles)) {
-      consoles_with_auth_problems.push_back(
-          std::string(console->resource_name_));
+        && (console->user_acl.HasAcl() || console->user_acl.profiles)) {
+      consoles_with_auth_problems.emplace_back(console->resource_name_);
     }
   }
   if (!consoles_with_auth_problems.empty()) {
     Jmsg(nullptr, M_FATAL, 0,
-         "Console(s) `%s` using `Use Pam Authentication` while "
-         "having ACL or a Profile activated\n",
+         "Combining `Use Pam Authentication` with ACL commands or `Profile` in "
+         "Console(s) `%s` is not allowed\n",
          consoles_with_auth_problems.Join(',').c_str());
     UnlockRes(my_config);
     return false;
--- a/core/src/dird/reload.cc
+++ b/core/src/dird/reload.cc
@@ -34,7 +34,6 @@ namespace directordaemon {
  */
 bool CheckResources()
 {
-  bool OK = true;
   JobResource* job;
   const std::string& configfile = my_config->get_base_config_path();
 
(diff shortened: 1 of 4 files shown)


core/src/dird/reload.cc
```diff
@@ -48,8 +47,8 @@ bool CheckResources()
          _("No Director resource defined in %s\n"
            "Without that I don't know who I am :-(\n"),
          configfile.c_str());
-    OK = false;
-    goto bail_out;
+    UnlockRes(my_config);
```
arogge, line 50: if we had GSL, you could set this up using `finally()` or if `ConfigurationParser` would satisfy _BasicLockable_ one could just use  a `lock_guard lock{my_config}`. This would remove the need for the explicit `UnlockRes(my_config)`

core/src/dird/reload.cc
```diff
@@ -183,20 +182,35 @@ bool CheckResources()
...
-    }
-    if (me->log_timestamp_format) {
-      SetLogTimestampFormat(me->log_timestamp_format);
+  ConsoleResource* console;
+  BStringList consoles_with_auth_problems;
+  foreach_res (console, R_CONSOLE) {
+    if (console->use_pam_authentication_
+        && (console->user_acl.ACL_lists[0] || console->user_acl.profiles)) {
+      consoles_with_auth_problems.push_back(
+          std::string(console->resource_name_));
```
arogge, line 191: With `emplace_back()` you can just pass your arguments into the constructor of the contained type (and save a copy operation):
```suggestion
consoles_with_auth_problems.emplace_back(console->resource_name_);
```

core/src/dird/reload.cc
```diff
@@ -183,20 +182,35 @@ bool CheckResources()
...
     }
   }
+  if (!consoles_with_auth_problems.empty()) {
+    Jmsg(nullptr, M_FATAL, 0,
+         "Console(s) `%s` using `Use Pam Authentication` while "
+         "having ACL or a Profile activated\n",
```
arogge, line 197: that describes a situation, but does not name the problem.

core/src/dird/reload.cc
```diff
@@ -183,20 +182,35 @@ bool CheckResources()
...
-    if (me->secure_erase_cmdline) {
-      SetSecureEraseCmdline(me->secure_erase_cmdline);
-    }
-    if (me->log_timestamp_format) {
-      SetLogTimestampFormat(me->log_timestamp_format);
+  ConsoleResource* console;
+  BStringList consoles_with_auth_problems;
+  foreach_res (console, R_CONSOLE) {
+    if (console->use_pam_authentication_
+        && (console->user_acl.ACL_lists[0] || console->user_acl.profiles)) {
```
arogge, line 189: if I understood it correctly, then `ACL_lists[0]` is only checking for Job_ACL, but not for the others like Catalog_ACL or Command_ACL.

arogge (changes requested):
We should really improve `LockRes()`/`UnlockRes()` as it is too easy to misuse.
I'm not sure if your check for the ACL is correct/complete, but overall it looks good.

core/src/dird/reload.cc
arogge: We can probably keep the old behaviour, but allow newer and refactored code to utilize RAII here.

Besides that, you can always do it like this:
```
void afunction() {
  {
     std::lock_guard lock(my_config);
     do_something
  }
  do_something_else
  {
     std::lock_guard lock(my_config);
     doxxx
  }
}
```
which isn't nice, but still a lot better than requiring manual unlock (which is btw. one of the main reasons a lot of our code is not exception-safe)

arogge (approved):

arogge (approved):
Great work!
Besides the Changelog, I'm happy with it.

PR #1330, opened 2022-12-11
fix file count mismatch on restores that use recycled volumes

alaaeddineelamri:
#### Description

When using recycled volumes, the restore process may report wrong expected number of files.
This PR fixes the issue.

#### Please check

##### General
- [x] PR name is meaningful
- [x] Check backport line
- ~Separate commit for CHANGELOG.md ("update CHANGELOG.md"). The PR number is correct.~

##### Source code quality

##### Tests

### DIFF
--- a/core/src/cats/sql_create.cc
+++ b/core/src/cats/sql_create.cc
@@ -102,21 +102,18 @@ bool BareosDb::CreateJobRecord(JobControlRecord* jcr, JobDbRecord* jr)
  */
 bool BareosDb::CreateJobmediaRecord(JobControlRecord* jcr, JobMediaDbRecord* jm)
 {
-  int count;
-  char ed1[50], ed2[50], ed3[50];
-
   DbLocker _{this};
 
   /* clang-format off */
   Mmsg(cmd,
-       "UPDATE JobMedia SET LastIndex=%lu, EndFile=%lu, EndBlock=%lu, JobBytes=%s "
-       "WHERE JobId=%s AND MediaId=%s",
+       "UPDATE JobMedia SET LastIndex=%lu, EndFile=%lu, EndBlock=%lu, JobBytes=%llu "
+       "WHERE JobId=%lu AND MediaId=%lu",
        jm->LastIndex,
        jm->EndFile,
        jm->EndBlock,
-       edit_uint64(jm->JobBytes, ed3),
-       edit_int64(jm->JobId, ed1),
-       edit_int64(jm->MediaId, ed2));
+       jm->JobBytes,
+       jm->JobId,
+       jm->MediaId);
   /* clang-format on */
 
   bool update_result = false;
@@ -128,9 +125,8 @@ bool BareosDb::CreateJobmediaRecord(JobControlRecord* jcr, JobMediaDbRecord* jm)
           _("Update JobMedia record %s failed: ERR=%s\n Trying to insert: \n"),
           cmd, sql_strerror());
 
-    Mmsg(cmd, "SELECT count(*) from JobMedia WHERE JobId=%s",
-         edit_int64(jm->JobId, ed1));
-    count = GetSqlRecordMax(jcr);
+    Mmsg(cmd, "SELECT count(*) from JobMedia WHERE JobId=%lu", jm->JobId);
+    int count = GetSqlRecordMax(jcr);
     if (count < 0) { count = 0; }
     count++;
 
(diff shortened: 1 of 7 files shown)


core/src/dird/restore.cc
```diff
@@ -433,17 +432,19 @@ void NativeRestoreCleanup(JobControlRecord* jcr, int TermCode)
 
   if (JobCanceled(jcr)) { CancelStorageDaemonJob(jcr); }
 
+  if (jcr->dir_impl->ExpectedFiles != jcr->JobFiles) { TermCode = JS_Warnings; }
```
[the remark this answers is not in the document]
pstorz: Strictly thinking, we have two possibilities here: 
1: TermCode is currently **JS_Terminated**, then we update the TermCode to  **JS_Warnings** and the `-- warnings file count mismatch` message
2.: Term Code is already **JS_Warning** and we have also a `file count mismatch` . Then we probably want to have `-- with warnings and file count mismatch` as message.

core/src/cats/sql_create.cc
```diff
@@ -102,21 +102,25 @@ bool BareosDb::CreateJobRecord(JobControlRecord* jcr, JobDbRecord* jr)
...
+       "WHERE JobId=%lu AND MediaId=%lu",
+       jm->FirstIndex,
+       jm->StartBlock,
+       jm->StartFile,
        jm->LastIndex,
        jm->EndFile,
        jm->EndBlock,
-       edit_uint64(jm->JobBytes, ed3),
-       edit_int64(jm->JobId, ed1),
-       edit_int64(jm->MediaId, ed2));
```
pstorz, line 119: Why did we have these edit_xxx calls here?

core/src/cats/sql_create.cc
```diff
@@ -102,21 +102,25 @@ bool BareosDb::CreateJobRecord(JobControlRecord* jcr, JobDbRecord* jr)
...
   DbLocker _{this};
 
   /* clang-format off */
   Mmsg(cmd,
-       "UPDATE JobMedia SET LastIndex=%lu, EndFile=%lu, EndBlock=%lu, JobBytes=%s "
-       "WHERE JobId=%s AND MediaId=%s",
+       "UPDATE JobMedia SET "
+       "FirstIndex = (CASE when firstindex=0 then %lu else firstindex end), "
+       "StartBlock = (CASE when startblock=0 then %lu else startblock end), "
+       "StartFile  = (CASE when startfile=0 then %lu else startfile end), "
```
pstorz, line 112: Are we sure that `firstindex`, `startblock` and `startfile` are always > 0 in normal operation, and only the dummy place holder sets them to 0?

pstorz (changes requested):
See comments.

core/src/cats/sql_create.cc
alaaeddineelamri: only the dummy placeholder creates zero-filled jobmedia record, no other place in the code requires those fields to be zero.

core/src/cats/sql_create.cc
alaaeddineelamri: direct heritage from the first ever commit, no specific reason mentioned.

All variables around it use the dedicated placeholders, there is no obvious reason for not changing those too.

my guess would be that it used to be useful back when 32-bit systems were supported.

core/src/dird/restore.cc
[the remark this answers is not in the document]
alaaeddineelamri: how about I just make the file count mismatch a joblog warning instead of a termination message?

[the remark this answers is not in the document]
arogge: If we can do that, it would be great. "Restore OK -- with warnings" with a reasonable error message would work for me.

core/src/cats/sql_create.cc
arogge: as we're overwriting placeholder data with real data, I would consider this safe (even though it isn't nice).

pstorz (approved):

core/src/dird/restore.cc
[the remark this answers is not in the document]
alaaeddineelamri: > If we can do that, it would be great. "Restore OK -- with warnings" with a reasonable error message would work for me.

already done.

core/src/cats/sql_create.cc
```diff
@@ -102,21 +102,25 @@ bool BareosDb::CreateJobRecord(JobControlRecord* jcr, JobDbRecord* jr)
...
   DbLocker _{this};
 
   /* clang-format off */
   Mmsg(cmd,
-       "UPDATE JobMedia SET LastIndex=%lu, EndFile=%lu, EndBlock=%lu, JobBytes=%s "
-       "WHERE JobId=%s AND MediaId=%s",
+       "UPDATE JobMedia SET "
+       "FirstIndex = (CASE WHEN firstindex=0 THEN %lu ELSE firstindex END), "
+       "StartBlock = (CASE WHEN startblock=0 THEN %lu ELSE startblock END), "
+       "StartFile  = (CASE WHEN startfile=0 THEN %lu ELSE startfile END), "
```
arogge, line 112: I just had an idea...
We could make the zero-Record contain NULL instead of 0. That would still provide the purge-protection we want, but we could easily check if it is a zero-record here.
So we could then alter the WHERE of this UPDATE here to also contain `FirstIndex IS NOT NULL`, so it won't update the zero-record, eventually creating a new one.

arogge: yeah... as discussed face-to-face: just forget it

PR #1390, opened 2023-02-23
VMware Plugin: fix restore to different vmname

sduehr:
Bugfix for restore to different vmname in same folder failing with duplicate name error. Check for invalid plugin option also works correctly now for both plugin options and config file.

##### General
- [x] Check backport line

##### Source code quality


core/src/plugins/filed/python/vmware/BareosFdPluginVMware.py
```diff
@@ -2668,6 +2681,8 @@ def transform(self, target_datastore_name=None):
         ]
         config_spec.migrateEncryption = self.config_info["migrateEncryption"]
         config_spec.name = self.config_info["name"]
+        if target_vm_name:
```
bruno-at-bareos, line 2684: Shouldn't we add a line into documentation https://docs.bareos.org/bareos-22/TasksAndConcepts/Plugins.html#description-of-all-plugin-options if vmname is present during restore the vm will be restored with the new name ?
Seems obvious but ...

bruno-at-bareos (changes requested):
Thanks for the fix, we need to agree on commit/PR title and adding or not the documentation line about how to restore vm with another name.
We should maybe extend the tests to restore with another name, and check the vm list with `vmware_cbt_tool.py` so major changes wouldn't be forgot on that one too.

bruno-at-bareos (changes requested):
We need to integrate a picture for the webui, and also make coherent the option used

bruno-at-bareos (approved):
The PR is now ready for merge

PR #1396, opened 2023-02-28
stored: change default block size to 1 MiB

arogge:
Historically Bareos uses a default block size of 63k (64.512 bytes). This usually works fine, until you use a modern tape drive. We already have a whilepaper on how to achieve good performance with tape drives, but you need to reconfigure the block sizes and there are a lot of ways you can break Bareos that way.
With the default block size set to 1M (1.048.576 bytes), there is no need to configure a custom block size for tapes anymore, as this new default now provides optimum performance for recent tape drives out of the box.
The default has been chosen carefully to provide optimum performance while retaining maximum interoperability. As there are still hostadapters (e.g. smartpqi) that cannot handle I/O larger than 1M and tape performance does not show significant increase when using a blocksize larger than 1M, this seems to be the reasonable choice.

#### Please check

##### General
- [x] Check backport line

##### Source code quality

##### Tests

### DIFF
--- a/core/src/stored/append.cc
+++ b/core/src/stored/append.cc
@@ -371,6 +371,7 @@ bool DoAppendData(JobControlRecord* jcr, BareosSocket* bs, const char* what)
   /* Check if we can still write. This may not be the case
    * if we are at the end of the tape or we got a fatal I/O error. */
   if (ok || dev->CanWrite()) {
+    jcr->JobFiles = file_currently_processed.GetFileIndex();
     if (!WriteSessionLabel(jcr->sd_impl->dcr, EOS_LABEL)) {
       // Print only if ok and not cancelled to avoid spurious messages
       if (ok && !jcr->IsJobCanceled()) {
--- a/core/src/stored/stored_conf.cc
+++ b/core/src/stored/stored_conf.cc
@@ -187,7 +187,7 @@ static ResourceItem dev_items[] = {
   {"LabelBlockSize", CFG_TYPE_PINT32, ITEM(res_dev, label_block_size), 0, CFG_ITEM_DEFAULT,
       "64512" /* DEFAULT_BLOCK_SIZE */, NULL, NULL},
   {"MinimumBlockSize", CFG_TYPE_PINT32, ITEM(res_dev, min_block_size), 0, 0, NULL, NULL, NULL},
-  {"MaximumBlockSize", CFG_TYPE_MAXBLOCKSIZE, ITEM(res_dev, max_block_size), 0, 0, NULL, NULL, NULL},
+  {"MaximumBlockSize", CFG_TYPE_MAXBLOCKSIZE, ITEM(res_dev, max_block_size), 0, CFG_ITEM_DEFAULT, "1048576", NULL, NULL},
   {"MaximumFileSize", CFG_TYPE_SIZE64, ITEM(res_dev, max_file_size), 0, CFG_ITEM_DEFAULT, "1000000000", NULL, NULL},
   {"VolumeCapacity", CFG_TYPE_SIZE64, ITEM(res_dev, volume_capacity), 0, 0, NULL, NULL, NULL},
   {"MaximumConcurrentJobs", CFG_TYPE_PINT32, ITEM(res_dev, max_concurrent_jobs), 0, CFG_ITEM_DEFAULT, "1", NULL, NULL},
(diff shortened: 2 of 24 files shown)


core/src/stored/stored_conf.cc
```diff
@@ -186,7 +186,7 @@ static ResourceItem dev_items[] = {
   {"LabelBlockSize", CFG_TYPE_PINT32, ITEM(res_dev, label_block_size), 0, CFG_ITEM_DEFAULT,
       "64512" /* DEFAULT_BLOCK_SIZE */, NULL, NULL},
   {"MinimumBlockSize", CFG_TYPE_PINT32, ITEM(res_dev, min_block_size), 0, 0, NULL, NULL, NULL},
-  {"MaximumBlockSize", CFG_TYPE_MAXBLOCKSIZE, ITEM(res_dev, max_block_size), 0, 0, NULL, NULL, NULL},
+  {"MaximumBlockSize", CFG_TYPE_MAXBLOCKSIZE, ITEM(res_dev, max_block_size), 0, CFG_ITEM_DEFAULT, "1048576", NULL, NULL},
   {"MaximumFileSize", CFG_TYPE_SIZE64, ITEM(res_dev, max_file_size), 0, CFG_ITEM_DEFAULT, "1000000000", NULL, NULL},
```
pstorz, line 190: Maybe increase the default for Maximum File Size to something bigger for modern tape drives?

core/src/stored/append.cc
```diff
@@ -370,6 +370,7 @@ bool DoAppendData(JobControlRecord* jcr, BareosSocket* bs, const char* what)
   /* Check if we can still write. This may not be the case
    * if we are at the end of the tape or we got a fatal I/O error. */
   if (ok || dev->CanWrite()) {
+    jcr->JobFiles = file_currently_processed.GetFileIndex();
```
pstorz, line 373: What is this line for?

docs/manuals/source/TasksAndConcepts/AutochangerSupport.rst
```diff
@@ -505,6 +505,12 @@ Tapespeed and blocksizes
 :index:`\ <single: Tape; speed>`
 :index:`\ <single: Blocksize; optimize>`
 
+.. note::
+  As of Bareos 23, the default block size has been increased to 1 MiB (1.048.576 bytes).
+  This provides optimal tape performance in the default configuration, so there is no need to tune the block sizes anymore.
```
pstorz, line 510:
```suggestion
  This provides close to optimal tape performance in the default configuration, so there is probably no need to tune the block sizes anymore.
```

docs/manuals/source/manually_added_config_directive_descriptions/sd-device-MinimumBlockSize.rst.inc
```diff
@@ -1,6 +1,8 @@
 This statement applies only to non-random access devices (e.g. tape drives). Blocks written by the storage daemon to a non-random archive device will never be smaller than the given size. The Storage daemon will attempt to efficiently fill blocks with data received from active sessions but will, if necessary, add padding to a block to achieve the required minimum size.
 
-To force the block size to be fixed, as is the case for some non-random access devices (tape drives), set the Minimum block size and the Maximum block size to the same value. The default is that both the minimum and maximum block size are zero and the default block size is 64,512 bytes.
+To force the block size to be fixed, as is the case for some non-random access devices (tape drives), set the Minimum block size and the Maximum block size to the same value.
```
pstorz, line 3:
```suggestion
To force the block size to be fixed, as is the case for some non-random access devices (tape drives), set the Minimum block size and the Maximum block size to the same value. This is usually not required.
```

pstorz (changes requested):
Good work! Please see comments.

core/src/stored/stored_conf.cc
arogge: That's probably a good idea, but right now I don't have a great default value that will work well in any case.
The culprit is stopping the tape (which could be mitigated by making the filemark non-immediate). Once that's done 10 GiB or 25 GiB is probably a good default. Until then I'm not sure - maybe 100 GiB or 250 GiB. We're trading restore seek-time for backup-speed here :(

docs/manuals/source/TasksAndConcepts/AutochangerSupport.rst
arogge: You're probably right, but that might sounds a bit too vague.
What about "This should provide optimal tape performance in the default configuration and remedy the need to tune block sizes."?

pstorz: Agreed!

pstorz (changes requested):
Only two points are left:
1: Adapt the documentation with your last suggestion of the sentence
2: Merge the latest docs commits into one.

Good work!

pstorz (approved):

PR #1454, opened 2023-04-25
daemons: update network handling when IP protocols unavailable

alaaeddineelamri:
#### Description

Even though IPv6 has been widely available across platforms, there can be situations where it can be outright disabled. In that case, Bareos wouldn't be able to create IPv6 sockets and would abort, dumping core files along the way.

The above issue prompted this PR, which updates how Bareos daemons handle this kind of situation.

#### Please check

##### General
- [x] Check backport line

##### Source code quality

##### Tests

### DIFF
--- a/core/src/lib/address_conf.cc
+++ b/core/src/lib/address_conf.cc
@@ -377,9 +377,9 @@ bool CheckIfFamilyEnabled(IpFamily family)
   do {
     ++tries;
     if ((fd = socket(GetFamily(family).value(), SOCK_STREAM, 0)) < 0) {
-      Bmicrosleep(5, 0);
+      Bmicrosleep(15, 0);
     }
-  } while (fd < 0 && tries < 2);
+  } while (fd < 0 && tries < 3);
 
   if (fd < 0) {
     BErrNo be;
--- a/core/src/lib/address_conf.cc
+++ b/core/src/lib/address_conf.cc
@@ -3,7 +3,7 @@
 
    Copyright (C) 2004-2011 Free Software Foundation Europe e.V.
    Copyright (C) 2011-2012 Planets Communications B.V.
-   Copyright (C) 2013-2022 Bareos GmbH & Co. KG
+   Copyright (C) 2013-2023 Bareos GmbH & Co. KG
 
    This program is Free Software; you can redistribute it and/or
    modify it under the terms of version three of the GNU Affero General Public
--- a/core/src/lib/address_conf.cc
+++ b/core/src/lib/address_conf.cc
@@ -37,9 +37,6 @@
 #ifdef HAVE_ARPA_NAMESER_H
 #  include <arpa/nameser.h>
 #endif
-#ifdef HAVE_RESOLV_H
-//#include <resolv.h>
-#endif
 
 IPADDR::IPADDR()
     : type(R_UNDEFINED), saddr(nullptr), saddr4(nullptr), saddr6(nullptr)
--- a/core/src/lib/address_conf.cc
+++ b/core/src/lib/address_conf.cc
@@ -339,7 +339,17 @@ bool SetupPort(unsigned short& port,
   }
 }
 
-bool FamilyEnabled(int family)
+bool IsFamilyEnabled(int family)
+{
+  static bool ipv6_enabled = CheckIfFamilyEnabled(AF_INET6);
+  static bool ipv4_enabled = CheckIfFamilyEnabled(AF_INET);
+
+  if (family == AF_INET6) { return ipv6_enabled; }
+  if (family == AF_INET) { return ipv4_enabled; }
+  return false;
+}
+
+bool CheckIfFamilyEnabled(int family)
 {
   int tries = 0;
   int fd;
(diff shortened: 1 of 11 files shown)


core/src/lib/address_conf.cc
```diff
@@ -338,6 +339,33 @@ bool SetupPort(unsigned short& port,
...
+    family_string = "IPv6";
+  }
+
+  if (fd < 0) {
+    BErrNo be;
+    Emsg2(M_WARNING, 0, _("Cannot open %s stream socket. ERR=%s\n"),
+          family_string.c_str(), be.bstrerror());
+    isEnabled = false;
+  }
+  close(fd);
```
sebsura, line 365: Dont close the fd if its < 0!

core/src/lib/address_conf.cc
```diff
@@ -341,6 +339,42 @@ bool SetupPort(unsigned short& port,
...
 
+bool IsFamilyEnabled(int family)
+{
+  static bool ipv6_enabled = CheckIfFamilyEnabled(AF_INET6);
+  static bool ipv4_enabled = CheckIfFamilyEnabled(AF_INET);
+
+  if (family == AF_INET6) { return ipv6_enabled; }
+  if (family == AF_INET) { return ipv4_enabled; }
+  return false;
+}
```
sebsura, line 350: Instead of trying to let the code handle programmer mistakes, we should try to make those mistakes impossible at compile time.  For example here we could do something like this:
```
enum class IpFamily { V4, V6 };

template <IpFamily fam>
bool IsFamilyEnabled() {
  static bool family_enabled[] = {
    CheckIfFamilyEnabled<IpFamily::V4>(),
    CheckIfFamilyEnabled<IpFamily::V6>()
 };
 return family_enabled[static_cast<std::underlying_type_t<IpFamily>>(fam)];
}
```

sebsura: The same is true for the other functions.

sebsura (changes requested):

core/src/lib/address_conf.cc
sebsura: Alternatively we could also have something like:

```
bool IsFamilyEnabled(IpFamily fam) { ... }
std::optional<IpFamily> GetFamily(int family) { switch (family) {
 case AF_INET: return IpFamily::V4;
 case AF_INET6: return IpFamily::V6;
 default: return std::nullopt; } }
```
So that at least `IsFamilyEnabled` and `CheckIfFamilyEnabled` do not have to check whether the family is a useful value.

sebsura: Alternatively we could have

template <IpFamily fam>
bool IsFamilyEnabled() {
  static bool is_enabled = CheckIfFamilyEnabled<fam>();
  return is_enabled;
}

core/src/tests/addresses_and_ports_functions.cc
```diff
@@ -0,0 +1,113 @@
...
+             hostname_str, port_str, buf, sizeof(buf));
+
+  IPADDR* addr = nullptr;
+  foreach_dlist (addr, addresses) {
+    addr->build_address_str(buf, sizeof(buf), true);
+
+    std::string theaddress = std::string(buf);
+    theaddress.pop_back();
+
+    newaddresses.push_back(theaddress);
```
sebsura, line 51: This does not matter since this is a test, but you can prevent the extra allocation here by doing
```suggestion
    std::string theaddress = std::string(buf);
    theaddress.pop_back();

    newaddresses.emplace_back(std::move(theaddress));
```
or
```suggestion
    newaddresses.emplace_back(buf).pop_back();
```

core/src/tests/addresses_and_ports_functions.cc
```diff
@@ -0,0 +1,113 @@
...
+}
+
+TEST_F(AddressesAndPortsFunctions, IPv4AddressesAreCorrectlyProcessed)
+{
+  std::vector<std::string> addresses = CreateAddressesFromAddAddress(
+      IPADDR::R_DEFAULT, AF_INET, default_port, "", "");
+  std::vector<std::string> expected_addresses{"host[ipv4;0.0.0.0;9101]"};
+  EXPECT_EQ(addresses, expected_addresses);
+
+  EXPECT_EQ(addresses, expected_addresses);
```
sebsura, line 77: Remove one of them.

core/src/tests/addresses_and_ports_functions.cc
```diff
@@ -0,0 +1,113 @@
...
+Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+02110-1301, USA.
+*/
+
+#include "testing_dir_common.h"
+
+#include "lib/address_conf.h"
+#include "lib/bnet.h"
+
+const unsigned short int default_port = 9101;
```
sebsura, line 27: I think just `short` is enough here.

sebsura (changes requested):

core/src/lib/address_conf.cc
```diff
@@ -341,6 +339,58 @@ bool SetupPort(unsigned short& port,
...
+    case IpFamily::V4:
+      return AF_INET;
+    case IpFamily::V6:
+      return AF_INET6;
+    default:
+      return std::nullopt;
+  }
+}
+
+std::string FamilyName(IpFamily fam)
```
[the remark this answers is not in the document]
sebsura: I guess something like
```
std::string_view FamilyName(IpFamily fam) {
  using namespace std::literals;
  switch (fam) {
    case IpFamily::V4: return "IPv6"sv;
    case IpFamily::V6: return "IPv6"sv;
 }
}
```
also works while also providing quick size access.

core/src/lib/address_conf.cc
```diff
@@ -341,6 +339,58 @@ bool SetupPort(unsigned short& port,
...
+{
+  switch (fam) {
+    case IpFamily::V4:
+      return "IPv4";
+    case IpFamily::V6:
+      return "IPv6";
+    default:
+      return "*Unknown Protocol*";
+  }
```
sebsura, line 370: Should we throw an exception here instead of returning normally ?  If we change the function to constexpr, then the compiler will catch those mistakes!
See https://godbolt.org/z/qf5hadP9Y for an example

sebsura (approved):
Looks fine to me!

PR #1476, opened 2023-05-26
status storage: fix wrong data displayed about waiting jobs

alaaeddineelamri:
#### Description

When using the `status storage=xxx` command, bareos reports that certain running jobs are writing to the wrong volume or using the wrong device. This is only a reporting issue as those jobs aren't running or writing, but are actually waiting, and so the data displayed does not yet correctly reflect where the job will write, until the waiting condition is over.

This PR:
* Updates the `status storage` output text to reflect the actual status of the jobs and show an empty volume for those who did not find a volume yet. Device name can still be wrong though. Even though that information is still not accurate, by seeing the status (waiting) of the job on top of the empty volume, one can infer that the device might be wrong.
* removes the `status resources` option as it is just a placeholder and not implemented.
* reverts a previous commit 6c265ec52d from PR #1271 that could potentially create an issue with ansi-labels

#### Please check

##### General

##### Source code quality


core/src/stored/reserve.cc
```diff
@@ -427,7 +427,10 @@ bool FindSuitableDeviceForJob(JobControlRecord* jcr, ReserveContext& rctx)
 
       // Check with Director if this Volume is OK
       bstrncpy(dcr->VolumeName, vol->vol_name, sizeof(dcr->VolumeName));
-      if (!dcr->DirGetVolumeInfo(GET_VOL_INFO_FOR_WRITE)) { continue; }
+      if (!dcr->DirGetVolumeInfo(GET_VOL_INFO_FOR_WRITE)) {
+        dcr->VolumeName[0] = 0;
+        continue;
```
sebsura, line 432: What happens if `DirGetVolumeInfo` succeeds but we do not find a usable device that can write to that volume ? In that case the volume name should also be reset since we will not use that volume!

sebsura: I would suggest reverting this change and instead doing
```
  if (ok) {
    Dmsg1(debuglevel, "OK dev found. Vol=%s from in-use vols list\n",
          rctx.VolumeName);
    return true;
  } else {
    // no existing volume found; reseting volume name.
    dcr->VolumeName[0] = '\0';
  }
```
after the loop.

core/src/stored/status.cc
```diff
@@ -648,14 +640,13 @@ static void ListRunningJobs(StatusPacket* sp)
...
-  time_t now = time(NULL);
   PoolMem msg(PM_MESSAGE);
-  int len, avebps, bps, sec;
+  int len;
   char JobName[MAX_NAME_LENGTH];
   char b1[50], b2[50], b3[50], b4[50];
 
   if (!sp->api) {
-    len = Mmsg(msg, _("\nRunning Jobs:\n"));
+    len = Mmsg(msg, _("\nExisting Jobs:\n\n"));
```
sebsura, line 649: I think you can leave this as 'Running'.  'Existing' sounds weird.

sebsura (changes requested):

sebsura (changes requested):
The device will still be displayed "incorrectly" if none are found right ?

core/src/stored/status.cc
alaaeddineelamri: I agree that `Existing` is a bit weird too. But the problem is that `Running` is misleading, since some of the displayed jobs are not always really in a running state, some might be blocked, some others might have just been created but not started yet. 

The point of `Existing` is to say that the job exists in the SD regardless of its status.

I've changed it to `Job Inventory` but feel free to suggest another word you see more fitting.

alaaeddineelamri:
> The device will still be displayed "incorrectly" if none are found right ?

yes, unfortunately. Resetting the volume name was quite straight forward, but the device name would require some changes that could lead to unknown consequences.

sebsura:
```
07-Jul 07:46 bareos-sd JobId 1: Please mount append Volume "Full-0005" or label a new one for:
    Job:          backup-bareos-fd.2023-07-07_07.46.15_04
    Storage:      "FileStorage" (storage)
    Pool:         Full
    Media type:   File
*status storage
Automatically selected Storage: File
Connecting to Storage daemon File at localhost:30473
 Encryption: TLS_CHACHA20_POLY1305_SHA256 TLSv1.3

bareos-sd Version: 23.0.0~pre732.eb6e817f9 (07 July 2023) Fedora 37 
Daemon started 07-Jul-23 07:46. Jobs: run=0, running=0, self-compiled binary
 Sizes: boffset_t=8 size_t=8 int32_t=4 int64_t=8 bwlimit=0kB/s

There are WARNINGS for this storagedaemon's configuration!
See output of 'bareos-sd -t' for details.

Job inventory:

JobId=1 Level=Full Type=Backup Name=backup-bareos-fd Status=Waiting for mount
Writing: Volume="Full-0005"
    pool="Full" device="FileStorage" (storage)
    spooling=0 despooling=0 despool_wait=0
    Files=0 Bytes=0 AveBytes/sec=0 LastBytes/sec=0
    FDReadSeqNo=6 in_msg=6 out_msg=4 fd=7

====

Jobs waiting to reserve a drive:
====

Terminated Jobs:
 JobId  Level    Files      Bytes   Status   Finished        Name 
===================================================================
     1  Full          0         0   Cancel   07-Jul-23 07:46 backup-bareos-fd
====

Device status:

Device "FileStorage" (storage) is not open.
    Device is BLOCKED waiting for mount of volume "Full-0005",
       Pool:        Full
       Media type:  File
==
====

Used Volume status:
====

====

```
I used a File Device without Auto Label enabled and got this output.  Is this the expected output after your changes ?

sebsura (approved):

PR #1487, opened 2023-06-20
dird: show current and allowed console connections

pstorz:
When the allowed number of console connectinos was exceeded, before only the fact that it is exceeded was logged without numbers.

Now both the allowed maximum and the current value are printed.

#### Please check

##### General
- [x] Check backport line

##### Source code quality

##### Tests
- [-] Decision taken that a test is required (if not, then remove this paragraph)
- [-] The choice of the type of test (unit test or systemtest) is reasonable
- [-] Testname matches exactly what is being tested
- [-] On a fail, output of the test leads quickly to the origin of the fault

### DIFF
--- a/core/src/dird/authenticate_console.cc
+++ b/core/src/dird/authenticate_console.cc
@@ -37,6 +37,7 @@
 #include "lib/util.h"
 #include "lib/version.h"
 #include "include/version_numbers.h"
+#include "console_connection_lease.h"
 
 namespace directordaemon {
 
@@ -354,18 +355,6 @@ static void LogErrorMessage(std::string console_name, UaContext* ua)
         ua->UA_sock->port());
 }
 
-static uint32_t CurrentNumberOfConsoleConnections()
-{
-  JobControlRecord* jcr;
-  uint32_t cnt = 0;
-
-  foreach_jcr (jcr) {
-    if (jcr->is_JobType(JT_CONSOLE)) { cnt++; }
-  }
-  endeach_jcr(jcr);
-  return cnt;
-}
-
 static bool GetConsoleNameAndVersion(BareosSocket* ua_sock,
                                      std::string& name_out,
                                      BareosVersionNumber& version_out)
@@ -406,15 +395,7 @@ static ConsoleAuthenticator* CreateConsoleAuthenticator(UaContext* ua)
 
 bool AuthenticateConsole(UaContext* ua)
 {
-  uint32_t ConsoleConnections = CurrentNumberOfConsoleConnections();
-  if (ConsoleConnections >= me->MaxConsoleConnections) {
-    Emsg0(M_ERROR, 0,
-          _("Number of console connections exceeded "
-            "Maximum :%u, Current: %u\n"),
-          me->MaxConsoleConnections, ConsoleConnections);
-    return false;
-  }
-
+  auto num_leases = ConsoleConnectionLease::get_num_leases();
   std::unique_ptr<ConsoleAuthenticator> console_authenticator(
       CreateConsoleAuthenticator(ua));
   if (!console_authenticator) { return false; }
@@ -425,7 +406,21 @@ bool AuthenticateConsole(UaContext* ua)
       LogErrorMessage(console_authenticator->console_name_, ua);
       return false;
     }
+    if (num_leases > me->MaxConsoleConnections) {
+      Emsg0(M_INFO, 0,
+            _("Number of console connections exceeded\n"
+              "Maximum: %u, Current: %zu\n"),
+            me->MaxConsoleConnections, num_leases);
+    }
   } else {
+    if (num_leases > me->MaxConsoleConnections) {
+      Emsg0(M_ERROR, 0,
+            _("Number of console connections exceeded "
+              "Maximum :%u, Current: %zu\n"),
+            me->MaxConsoleConnections, num_leases);
+      return false;
+    }
+
     console_authenticator->AuthenticateNamedConsole();
     if (!console_authenticator->auth_success_) {
       LogErrorMessage(console_authenticator->console_name_, ua);
--- a/core/src/dird/authenticate_console.cc
+++ b/core/src/dird/authenticate_console.cc
@@ -354,17 +354,16 @@ static void LogErrorMessage(std::string console_name, UaContext* ua)
         ua->UA_sock->port());
 }
 
-static bool NumberOfConsoleConnectionsExceeded()
+static uint32_t CurrentNumberOfConsoleConnections()
 {
   JobControlRecord* jcr;
-  unsigned int cnt = 0;
+  uint32_t cnt = 0;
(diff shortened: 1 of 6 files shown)


bruno-at-bareos (commented):
Looks goods, see my remark

bruno-at-bareos:
Also if you change something maybe you can fix the check-source warning 
 ✗  Commit checks failed:
        f094a2ab4 dird: show current and allowed console connections: body contains line longer 72 chars

core/src/dird/authenticate_console.cc
```diff
@@ -354,17 +354,16 @@ static void LogErrorMessage(std::string console_name, UaContext* ua)
...
+  uint32_t cnt = 0;
 
   foreach_jcr (jcr) {
     if (jcr->is_JobType(JT_CONSOLE)) { cnt++; }
   }
   endeach_jcr(jcr);
-
-  return (cnt >= me->MaxConsoleConnections) ? true : false;
+  return cnt;
 }
```
sebsura, line 367: The problem with this function is that it can return wrong results if there is a lot of job contention.
There is a really ugly way to prevent this:

```C++
static std::uint32_t CurrentNumberOfConsoleConnections()
{
  JobControlRecord* jcr;

  std::vector<JobControlRecord*> vec;
  foreach_jcr (jcr) {
    if (jcr->is_JobType(JT_CONSOLE)) {
     // keep the jcr alive
      jcr->IncUseCount();
      vec.push(jcr);
    }
  }
  endeach_jcr(jcr);

  for (auto* jcr : vec) {
   // this looks weird but its just the opposite of IncUseCount()
    FreeJcr(jcr);
  }

  constexpr std::uint32_t max = std::numeric_limits<std::uint32_t>::max();
  if (vec.size() > max) { return max; }
  return vec.size();
}
```
This stops jcrs from getting destroyed after we have counted them.

core/src/dird/authenticate_console.cc
```diff
@@ -407,10 +406,12 @@ static ConsoleAuthenticator* CreateConsoleAuthenticator(UaContext* ua)
 
 bool AuthenticateConsole(UaContext* ua)
 {
-  if (NumberOfConsoleConnectionsExceeded()) {
+  uint32_t ConsoleConnections = CurrentNumberOfConsoleConnections();
+  if (ConsoleConnections >= me->MaxConsoleConnections) {
     Emsg0(M_ERROR, 0,
           _("Number of console connections exceeded "
-            "MaximumConsoleConnections\n"));
+            "Maximum :%u, Current: %u\n"),
```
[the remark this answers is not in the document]
sebsura: Is this run on a single thread ? Otherwise this does not really ensure that the invariant
`ConsoleConnections >= me->MaxConsoleConnectinos` hold.  But it should be true in most circumstances (think of the scenario where two consoles want to connect "at the same time" while there is only one spot left.

sebsura (changes requested):
Depending on what we want to accomplish this needs to be changed somewhat.

core/src/dird/authenticate_console.cc
sebsura: We could even be a bit less pessimistic and say
```
std::size_t ended = 0;
for (auto* jcr : vec) {
  if (jcr->UseCount() == 1) {
     ended += 1;
  }
  FreeJcr(jcr);
}

...

std::size_t num_counted = vec.size();
std::size_t at_most_alive = num_counted - ended;
if (at_most_alive > max) return max;
return at_most_alive;
```

sebsura: Actually this does not really help the situation.

arogge (dismissed):
I don't have a good idea how to make this nicer, but the current approach is not great. Too much mixing of concerns and just too complicated in my opinion.

core/src/dird/authenticate_console.cc
```diff
@@ -424,7 +406,21 @@ bool AuthenticateConsole(UaContext* ua)
       LogErrorMessage(console_authenticator->console_name_, ua);
       return false;
     }
+    if (num_leases > me->MaxConsoleConnections) {
+      Emsg0(M_INFO, 0,
+            _("Number of console connections exceeded\n"
+              "Maximum: %u, Current: %u\n"),
+            me->MaxConsoleConnections, num_leases);
```
sebsura, line 413:
```suggestion
            me->MaxConsoleConnections, static_cast<unsigned>(num_leases));
```
Sadly there is no `saturating_cast` but I do not foresee anybody having the limit close to the 32bit limit so that should not matter.

sebsura: If `Emsg` supports it one could also just use `%zu` instead.

core/src/dird/authenticate_console.cc
```diff
@@ -424,7 +406,21 @@ bool AuthenticateConsole(UaContext* ua)
...
+    }
   } else {
+    if (num_leases > me->MaxConsoleConnections) {
+      Emsg0(M_ERROR, 0,
+            _("Number of console connections exceeded "
+              "Maximum :%u, Current: %u\n"),
+            me->MaxConsoleConnections, num_leases);
```
sebsura, line 420: Same here.

core/src/dird/authenticate_console.cc
```diff
@@ -405,42 +393,71 @@ static ConsoleAuthenticator* CreateConsoleAuthenticator(UaContext* ua)
...
+authentication::authentication(std::size_t max_console_connections)
+    : max_cons{max_console_connections}
+{
+  num_cons = NumAuthenticatedConsoles.load(std::memory_order_relaxed);
+
+  while (!incremented_counter && (num_cons < max_cons)) {
+    incremented_counter = NumAuthenticatedConsoles.compare_exchange_weak(
+        num_cons, num_cons + 1, std::memory_order_relaxed,
+        std::memory_order_relaxed);
+  }
```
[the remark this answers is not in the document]
sebsura: Yeah that is what the loop does, but in a thread safe manner!  The class invariant thing is probably a good idea.  One might even return a Optional<Authentication> to make it more clear that it is not something guaranteed.

core/src/dird/authenticate_console.cc
```diff
@@ -405,42 +393,71 @@ static ConsoleAuthenticator* CreateConsoleAuthenticator(UaContext* ua)
   }
 }
 
-bool AuthenticateConsole(UaContext* ua)
+static std::atomic<std::size_t> NumAuthenticatedConsoles{0};
+
+authentication::authentication(std::size_t max_console_connections)
+    : max_cons{max_console_connections}
+{
+  num_cons = NumAuthenticatedConsoles.load(std::memory_order_relaxed);
```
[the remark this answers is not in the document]
sebsura: It was saved to prevent nonsensical outputs in the `Dmsg()`.

sebsura (approved):

PR #1496, opened 2023-06-28
stored: automatically increment tape block size in case the buffer is too small

arogge:
This change allows Bareos to read tape volumes even if the configured `Maximum Block Size` is too small.
You will still see an error message like this to notify you that something is not right:
```
28-Jun 16:10 bareos-sd JobId 15: Error: stored/block.cc:1057 Block length 162132 is greater than buffer 64512. Attempting recovery.
28-Jun 16:10 bareos-sd JobId 15: stored/block.cc:1079 Setting block buffer size to 162132 bytes.
```
However, restore will not fail anymore.

#### Please check

##### General
~~Check backport line~~

##### Source code quality

### DIFF
--- a/core/src/stored/backends/unix_tape_device.cc
+++ b/core/src/stored/backends/unix_tape_device.cc
@@ -52,20 +52,16 @@
 #include "generic_tape_device.h"
 #include "unix_tape_device.h"
 
-namespace {
-static constexpr size_t size_increment = 1024 * 1024;
-static constexpr size_t max_size_increment = 16 * size_increment;
-static constexpr size_t next_size_increment(size_t count)
-{
-  if (count < size_increment) {
-    return size_increment;
-  } else {
-    return (count / size_increment) * 2 * size_increment;
-  }
-}
-}  // namespace
-
 namespace storagedaemon {
+
+static std::vector<std::size_t> bufsizes(std::size_t count)
+{
+  constexpr std::size_t mb = 1024 * 1024;
+  std::vector<std::size_t> sizes{1 * mb, 2 * mb, 4 * mb, 8 * mb, 16 * mb};
+  sizes.erase(sizes.begin(),
+              std::upper_bound(sizes.begin(), sizes.end(), count));
+  return sizes;
+}
 
 int unix_tape_device::d_ioctl(int fd, ioctl_req_t request, char* op)
 {
@@ -80,21 +76,33 @@ unix_tape_device::unix_tape_device()
 ssize_t unix_tape_device::d_read(int fd, void* buffer, size_t count)
 {
   ssize_t ret = ::read(fd, buffer, count);
-  /* when the driver fails to `read()` with `ENOMEM`, then the provided buffer
-   * was too small by re-reading with a temporary buffer that is enlarged
-   * step-by-step, we can read the block and returns its first `count` bytes.
+  /* If the driver fails to `read()` with `ENOMEM`, then the provided buffer
+   * was too small. By re-reading with a temporary buffer that is enlarged
+   * step-by-step, we can read the block and return the first `count` bytes.
    * This allows the calling code to read the block header and resize its buffer
    * according to the size recorded in that header.
    */
-  for (size_t bufsize = next_size_increment(count);
-       ret == -1 && errno == ENOMEM && bufsize <= max_size_increment;
-       bufsize = next_size_increment(bufsize)) {
-    std::vector<char> tmpbuf(bufsize);
-    bsr(1);       // go back one block so we can re-read
-    block_num++;  // re-increment the block counter that bsr() just incremented
-    if (auto tmpret = ::read(fd, tmpbuf.data(), tmpbuf.size()); tmpret != -1) {
-      memcpy(buffer, tmpbuf.data(), count);
-      ret = std::min(tmpret, static_cast<ssize_t>(count));
+  if (ret == -1 && errno == ENOMEM && HasCap(CAP_BSR)) {
+    for (auto bufsize : bufsizes(count)) {
+      // first go back one block so we can re-read
+      if (!bsr(1)) {
+        /* when backward-spacing fails for some reason, there's not much we
+         * can do, so we just return the original ENOMEM and hope that the
+         * caller knows the device is in a non well-defined state.
+         */
+        errno = ENOMEM;
+        return -1;
+      }
+      block_num++;  // re-increment the block counter bsr() just decremented
+      std::vector<char> tmpbuf(bufsize);
+      if (auto tmpret = ::read(fd, tmpbuf.data(), tmpbuf.size());
+          tmpret != -1) {
+        memcpy(buffer, tmpbuf.data(), count);
+        ret = std::min(tmpret, static_cast<ssize_t>(count));
+        break;  // successful read
+      } else if (errno != ENOMEM) {
+        break;  // some other error occured
+      }
     }
   }
   return ret;
--- a/core/src/stored/backends/unix_tape_device.cc
+++ b/core/src/stored/backends/unix_tape_device.cc
@@ -40,6 +40,9 @@
  * @file
  * UNIX Tape API device abstraction.
  */
+#include <algorithm>
+#include <vector>
+
 #include <sys/ioctl.h>
 #include <unistd.h>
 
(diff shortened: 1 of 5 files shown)


core/src/stored/backends/unix_tape_device.cc
```diff
@@ -61,6 +77,29 @@ unix_tape_device::unix_tape_device()
...
+   * step-by-step, we can read the block and returns its first `count` bytes.
+   * This allows the calling code to read the block header and resize its buffer
+   * according to the size recorded in that header.
+   */
+  for (size_t bufsize = next_size_increment(count);
+       ret == -1 && errno == ENOMEM && bufsize <= max_size_increment;
+       bufsize = next_size_increment(bufsize)) {
+    std::vector<char> tmpbuf(bufsize);
+    bsr(1);       // go back one block so we can re-read
+    block_num++;  // re-increment the block counter that bsr() just incremented
```
pstorz, line 94:
```suggestion
    block_num++;  // re-increment the block counter that bsr() just decremented
```

core/src/stored/block.cc
```diff
@@ -1060,7 +1060,9 @@ DeviceControlRecord::ReadStatus DeviceControlRecord::ReadBlockFromDev(bool)
     // Attempt to Reposition to re-read the block
     if (dev->IsTape()) {
       Dmsg0(250, "BootStrapRecord for reread; block too big for buffer.\n");
-      if (!dev->bsr(1)) {
+      if (dev->bsr(1)) {
+        dev->block_num++;  // reincrement what bsr() decremented
```
pstorz, line 1064:
```suggestion
        dev->block_num++;  // re-increment what bsr() decremented
```

core/src/stored/backends/unix_tape_device.cc
```diff
@@ -49,6 +52,19 @@
 #include "generic_tape_device.h"
 #include "unix_tape_device.h"
 
+namespace {
+static constexpr size_t size_increment = 1024 * 1024;
```
pstorz, line 56: Does an anonymous namespace not already serve like the static specifier?

pstorz (commented):
Looks good; Please check my comment.

core/src/stored/backends/unix_tape_device.cc
arogge: My thought was that putting stuff in the global namespace isn't great (even though it is static), but putting it in the storagedaemon namespace wouldn't be great either.
And that's how it ended up in an anonymous namespace.

sebsura: might be a bit too cute but what about
```
class bufsizes
{
public:
bufsizes() = default;
bufsizes(std::size_t initial_size) : current_size{round_up(initial_size)} {}
bufsizes end() { return bufsizes{}; }
bufsizes begin() { return *this; }
bufsizes& operator++() {
  current_size = next(current_size);
  return *this;
}

std::size_t operator*() {
    return current_size;
}

friend bool operator!=(const bufsizes& lhs, const bufsizes& rhs) {
    return lhs.current_size != rhs.current_size;
}

private:
static constexpr std::size_t increment = 1024 * 1024;
static constexpr std::size_t max_size = 16 * increment;
std::size_t current_size{next(max_size)};
static constexpr std::size_t round_up(std::size_t val) {
    return ((val + increment - 1) / increment) * increment;
}

static constexpr std::size_t next(std::size_t val) {
    return 2 * val;
}
};
```

arogge: We can even simplify more: it is sufficient to have a `std::vector<size_t> bufsizes(size_t start)` that returns the sizes to try, no need for a new type...

arogge: What about the following implementation?
```
std::vector<std::size_t> bufsizes(std::size_t count) {
  constexpr std::size_t mb = 1024 * 1024;
  std::vector<std::size_t> sizes{1 * mb, 2 * mb, 4 * mb, 8 * mb, 16 * mb};
  sizes.erase(std::remove_if(sizes.begin(), sizes.end(),
                             [count](size_t s) { return s <= count; }),
              sizes.end());
  return sizes;
}
```

sebsura: That also works.  Its a shame that it allocated memory, but we can just replace it by span once C++20 lands and we touch the code again.

sebsura: This is splitting hairs but i think we can just use
```
std::erase(std::begin(sizes), std::upper_bound(sizes.begin(), sizes.end(), count));
```
since we know the vector is sorted.

arogge: Yep, that's definitely better.

Concerning the allocation I don't think it is a problem.
1. we're on the error-path
2. we're rewinding a tape once, which is probably one of the slowest operations possible

If we really want to get rid of the allocation we can use a `static constexpr std:array<size_t>` to iterate over and then just `if (x < y) continue;` within the loop.

arogge:
@sebsura I implemented what I think the both of us settled on. Would you mind taking another look?

sebsura (changes requested):
The code looks great. One small detail still needs work.
I think you should mention the build script fix in the PR text itself.

core/src/stored/backends/unix_tape_device.cc
```diff
@@ -61,6 +73,33 @@ unix_tape_device::unix_tape_device()
   SetCap(CAP_ADJWRITESIZE); /* Adjust write size to min/max */
 }
 
+ssize_t unix_tape_device::d_read(int fd, void* buffer, size_t count)
+{
+  ssize_t ret = ::read(fd, buffer, count);
+  /* If the driver fails to `read()` with `ENOMEM`, then the provided buffer
+   * was too small. By re-reading with a temporary buffer that is enlarged
+   * step-by-step, we can read the block and returns its first `count` bytes.
```
sebsura, line 81:
```suggestion
   * step-by-step, we can read the block and return the first `count` bytes.
```

core/src/stored/backends/unix_tape_device.cc
```diff
@@ -61,6 +73,33 @@ unix_tape_device::unix_tape_device()
...
+      bsr(1);       // go back one block so we can re-read
+      block_num++;  // re-increment the block counter bsr() just decremented
+      std::vector<char> tmpbuf(bufsize);
+      if (auto tmpret = ::read(fd, tmpbuf.data(), tmpbuf.size());
+          tmpret != -1) {
+        memcpy(buffer, tmpbuf.data(), count);
+        ret = std::min(tmpret, static_cast<ssize_t>(count));
+        break;  // successful read
+      } else if (errno != ENOMEM) {
+        break;  // we have a problem what wasn't a too small buffer
```
sebsura, line 96:
```suggestion
        break;  // we have a problem which isn't the small buffer
```

core/src/stored/backends/unix_tape_device.cc
```diff
@@ -61,6 +73,37 @@ unix_tape_device::unix_tape_device()
...
+   * This allows the calling code to read the block header and resize its buffer
+   * according to the size recorded in that header.
+   */
+  if (ret == -1 && errno == ENOMEM) {
+    for (auto bufsize : bufsizes(count)) {
+      // first go back one block so we can re-read
+      if(!bsr(1)) {
+        errno = EIO;
+        return ret;
+      }
```
sebsura, line 91: This should not return `EIO`, otherwise readblockfromdev might try to accidentally read the next block.

core/src/stored/backends/unix_tape_device.cc
```diff
@@ -61,6 +73,41 @@ unix_tape_device::unix_tape_device()
...
+   * This allows the calling code to read the block header and resize its buffer
+   * according to the size recorded in that header.
+   */
+  if (ret == -1 && errno == ENOMEM) {
+    for (auto bufsize : bufsizes(count)) {
+      // first go back one block so we can re-read
+      if(!bsr(1)) {
+        /* when backward-spacing fails for some reason, we there's not much we
+         * can do, so we just return the original ENOMEM and hope that the
+         * caller knows the device is in a non well-defined status.
```
sebsura, line 91:
```suggestion
        /* when backward-spacing fails for some reason, there's not much we
         * can do, so we just return the original ENOMEM and hope that the
         * caller knows the device is in a non well-defined state.
```

sebsura (approved):
Looks great!

PR #1499, opened 2023-06-30
dird: add prev and new jobid variables

SamuelBoerlin:

This PR adds two new RunScript variables %O and %N which contain the previous and new jobid of a migration/copy job.
Currently there is no convenient way to retrieve this information in RunScripts. This would be useful to e.g. easily archive copied or migrated jobs (especially with always incremental and in combination with https://github.com/bareos/bareos/pull/1372):
```
  Run Script {
        console = "update jobid=%N jobtype=A"
        Runs When = After
        Runs On Client = No
        Runs On Failure = No
  }
```

#### Please check

##### General
- [x] Check backport line

##### Source code quality

##### Tests

### DIFF
--- a/core/src/dird/dird_conf.cc
+++ b/core/src/dird/dird_conf.cc
@@ -3083,33 +3083,32 @@ bail_out:
  * %m = Prev Backup JobId
  * %M = New Backup JobId
  */
-extern "C" char* job_code_callback_director(JobControlRecord* jcr,
+extern "C" std::optional<std::string> job_code_callback_director(JobControlRecord* jcr,
                                             const char* param)
 {
-  static char yes[] = "yes";
-  static char no[] = "no";
-  
-  static char str[50];
+  static const std::string yes = "yes";
+  static const std::string no = "no";
+  static const std::string none = "*None*";
 
   switch (param[0]) {
     case 'f':
       if (jcr->dir_impl->res.fileset) {
-        return jcr->dir_impl->res.fileset->resource_name_;
+        return std::string(jcr->dir_impl->res.fileset->resource_name_);
       }
       break;
     case 'h':
       if (jcr->dir_impl->res.client) {
-        return jcr->dir_impl->res.client->address;
+        return std::string(jcr->dir_impl->res.client->address);
       }
       break;
     case 'p':
       if (jcr->dir_impl->res.pool) {
-        return jcr->dir_impl->res.pool->resource_name_;
+        return std::string(jcr->dir_impl->res.pool->resource_name_);
       }
       break;
     case 'w':
       if (jcr->dir_impl->res.write_storage) {
-        return jcr->dir_impl->res.write_storage->resource_name_;
+        return std::string(jcr->dir_impl->res.write_storage->resource_name_);
       }
       break;
     case 'x':
@@ -3117,7 +3116,7 @@ extern "C" char* job_code_callback_director(JobControlRecord* jcr,
     case 'C':
       return jcr->dir_impl->cloned ? yes : no;
     case 'D':
-      return my_name;
+      return std::string(my_name);
     case 'V':
       if (jcr) {
         /* If this is a migration/copy we need the volume name from the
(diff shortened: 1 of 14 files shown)


core/src/dird/dird_conf.cc
```diff
@@ -3080,12 +3080,16 @@ static void StoreRunscript(LEX* lc, ResourceItem* item, int index, int pass)
...
+ * %m = Prev Backup JobId
+ * %M = New Backup JobId
  */
 extern "C" char* job_code_callback_director(JobControlRecord* jcr,
                                             const char* param)
 {
   static char yes[] = "yes";
   static char no[] = "no";
+  
+  static char str[50];
```
SamuelBoerlin, line 3092: Is using a static array to store and return the result okay? I copied that from another similar callback, so I'm not sure.

arogge: I don't really like it, because it won't be thread-safe. i.e. if the threads for multiple jobs enter `job_code_callback()` around the same time, there's a possible data-race on `str` which may lead to the wrong value being returned.

However, there's no great way to work around this inside the function right now. I guess the best way to proceed would be to remove the requirement for `extern "C"` and return a `std::string` instead of a `char*`.
I'll see if I can make that change easily.

SamuelBoerlin: Alright, thanks for your feedback. I already have a branch where I've refactored this callback to return an std::optional<std::string>, should I merge that into this PR or do you want that somewhere else?

SamuelBoerlin: I've only refactored it up to util.cc#edit_job_codes though (https://github.com/SamuelBoerlin/bareos/commit/0b7498a18f66778f1612d003ad485285a0ee3164#diff-edeceb51c3de23d07b2d5d9e679d111e1238b74cc870a492143e5283e4386ee9R937-R943), after which point it gets more complicated.

SamuelBoerlin:
Just realized there seems to already exist a %m variable here: https://github.com/bareos/bareos/blob/master/core/src/filed/fileset.cc#L68-L69
but I couldn't find anything about this in the documentation. Technically this shouldn't be a problem because this %m and my proposed %m should never conflict, but it might be confusing from a configuration standpoint.

core/src/dird/dird_conf.cc
arogge: That looks promising! You can just add it to this PR as it is the refactoring required to implement the feature you're working on.

You could
- drop the `extern "C"` everywhere, having C-linkage if you need a `std::string` probably won't help much
- return the plain `char*`/`const char*`, compiler will take care of conversion on return

Your solution for `edit_job_codes()` is probably as good as it gets without refactoring the whole thing. However, for a plain `char*` `snprintf()` is better suited than `Bsnprintf()`, as the latter expects a resizable `POOLMEM*`.

arogge (commented):
btw. I cannot thank you enough for adding the test for your new feature!

arogge:
That looks really really good. The only issue I see is the letters you chose.
We already have `%m` on the FD and it is hard enough to understand without duplicate letters.
As the set of letters to select from is `agkoquyzAEGHIJKLMNOQRSTUWXYZ` my proposal is to go with `%O` for the original jobid and `%N` for new jobid.
But I'm happy with whatever other non-duplicate letters you come up with.

SamuelBoerlin:
Thanks! Okay, makes sense. I'll go with `%O` and `%N` then.

SamuelBoerlin:
I've also added %m to the docs because that one was undocumented as far as I can tell.

arogge:
Sorry for keeping you waiting so long.
That changes look good (I'm going to finish the review today). You can (and probably should) drop the CHANGELOG commit. We have a Bot that adds a ChangeLog record based on the PR title automatically.

arogge (approved):
These changes look great. Thank you for the effort you put into this!
You can fix the CHANGELOG.md conflict by just removing your change in that file, the Bot will take care of that.
Do you have a need to see this in Bareos 22? I think we could backport these changes if you have a need for them to become available in the next Bareos 22 release...

TheCyberBrick:
Great! I've removed the changelog changes again.
Yes, it would be nice to have this in the next Bareos 22 release. Currently I get the new jobid by just adding +1 to the old jobid, but that seems unsafe and prone to breaking, so it'd be nice to have this new variable instead.

SamuelBoerlin:
Great! I've removed the changelog changes again.
Yes, it would be nice to have this in the next Bareos 22 release. Currently I get the new jobid by just adding +1 to the old jobid, but that seems unsafe and prone to breaking, so it'd be nice to have this new variable instead.

PR #1533, opened 2023-08-22
filed: parallel compression/reading/sending

sebsura:

Instead of serially doing read chunk -> compress chunk -> send chunk we instead try to do as much as possible in parallel.

#### Please check

##### General
- [x] Check backport line

##### Source code quality

##### Tests


core/src/lib/thread_pool.h
```diff
@@ -0,0 +1,143 @@
...
+template <typename F>
+auto enqueue(thread_pool& pool, F&& f) -> std::future<std::invoke_result_t<F>>
+{
+  using result_type = std::invoke_result_t<F>;
+  // todo(ssura): in c++23 we can use std::move_only_function
+  //              and pass the promise directly into the function.
+  //              currently this approach does not work because
+  //              std::function requires the function to be copyable,
+  //              but std::promise obviously is not.
+  std::shared_ptr p = std::make_shared<std::promise<result_type>>();
```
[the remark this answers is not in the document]
arogge: Nevermind, I just understood the comment above (that I already read three times before...)

core/src/lib/thread_pool.h
```diff
@@ -0,0 +1,143 @@
...
+      if constexpr (std::is_same_v<result_type, void>) {
+        f();
+        mp->set_value();
+      } else {
+        mp->set_value(f());
+      }
+    } catch (...) {
+      mp->set_exception(std::current_exception());
+    }
+  });
```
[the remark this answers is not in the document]
sebsura: Its important to note that i package a function f returning a B into a wrapper that returns nothing while still returning a std::future\<B\>.
This is not pretty to do with `std::packaged_task` since you either
* return a std::future\<void\> with std::packaged_task\<void\>
* return a std::future\<B\> with std::packaged_task\<B\>, but you now have to somehow create a type erased wrapper for the task.  std::function will not work since it requires being copyable.

I have not found a way yet to make this with packaged_task while still being somewhat easy to understand and read.
It should be doable with `std::future::then` but thats still in experimental so not even in C++23.

core/src/filed/backup.cc
```diff
@@ -1030,6 +1037,233 @@ static inline bool SendPlainData(b_ctx& bctx)
...
+          // Update checksum if requested
+          if (checksum) {
+            CryptoDigestUpdate(checksum, (uint8_t*)buf->data(), buf->size());
+          }
+
+          // Update signing digest if requested
+          if (signing) {
+            CryptoDigestUpdate(signing, (uint8_t*)buf->data(), buf->size());
+          }
+        }));
```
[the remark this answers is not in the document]
sebsura: We await the future before submitting any new digest updating, thus ensuring that they are updated in the correct order (and without raceconditions on the digest!).  I should write a comment here to clarify!

core/src/include/jcr.h
```diff
@@ -242,6 +242,7 @@ class JobControlRecord {
 extern int GetNextJobidFromList(const char** p, uint32_t* JobId);
 extern bool InitJcrSubsystem(int timeout);
 extern JobControlRecord* new_jcr(JCR_free_HANDLER* daemon_free_jcr);
+extern void register_jcr(JobControlRecord* jcr);
```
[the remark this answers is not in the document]
sebsura: done; also removed a race condition on `num_jobs_run` here.

core/src/lib/jcr.cc
```diff
@@ -216,14 +216,19 @@ JobControlRecord* new_jcr(JCR_free_HANDLER* daemon_free_jcr)
   jcr = new (jcr) JobControlRecord();
 
   jcr->daemon_free_jcr = daemon_free_jcr;
+  return jcr;
+}
+
+void register_jcr(JobControlRecord* jcr)
+{
+  Dmsg0(debuglevel, "Enter register_jcr\n");
```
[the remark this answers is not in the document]
sebsura: The reason for this two-step initialization is because consumers of the jcr chain sometimes do not consider the fact the the jcr they are looking at is not initialized completely (and there is not really a good way to check for that; some do it by checking for jobid==0 but thats not great).

One  such example that does nothing to check for initialization is in `ListJobsWaitingOnReservation` in `stored/status.cc`:

```C++
  foreach_jcr (jcr) {
    if (!jcr->sd_impl->reserve_msgs) { continue; } // <- jcr->sd_impl might still be 0 here!
    SendDriveReserveMessages(jcr, sp);
  }
```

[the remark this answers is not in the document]
sebsura: Its probably best to only add jcrs to the chain that are completely set up (jobid and all) but this might break stuff so for now i just made sure that at least the `*_impl` pointers are setup correctly.

core/src/filed/backup.cc
```diff
@@ -1030,6 +1033,384 @@ static inline bool SendPlainData(b_ctx& bctx)
...
+
+  void set_header(std::uint64_t h)
+  {
+    has_header = true;
+    auto* ptr = header_ptr();
+    std::memcpy(ptr, &h, header_size);
+
+    for (std::size_t i = 0; i < header_size / 2; ++i) {
+      std::swap(ptr[i], ptr[header_size - 1 - i]);
+    }
```
arogge, line 1104: that would byteswap on a big-endian platform, too.

core/src/filed/backup.cc
```diff
@@ -1030,6 +1033,384 @@ static inline bool SendPlainData(b_ctx& bctx)
...
+  }
+
+  void resize(std::size_t new_size) { buffer.resize(data_offset + new_size); }
+
+  char* header_ptr() { return &buffer[bnet_size]; }
+  char* data_ptr() { return &buffer[bnet_size + header_size]; }
```
arogge, line 1110:
```suggestion
  char* data_ptr() { return &buffer[data_offset]; }
```

core/src/filed/backup.cc
```diff
@@ -1030,6 +1033,384 @@ static inline bool SendPlainData(b_ctx& bctx)
...
+  const char* header_ptr() const { return &buffer[bnet_size]; }
+  const char* data_ptr() const { return &buffer[bnet_size + header_size]; }
```
arogge, line 1112:
```suggestion
  const char* data_ptr() const { return &buffer[data_offset]; }
```

core/src/filed/backup.cc
```diff
@@ -1030,6 +1033,384 @@ static inline bool SendPlainData(b_ctx& bctx)
...
+
+  /* important: this is not actually a POOLMEM*; do not pass it to POOLMEM*
+   *            functions, except to pass it to BareosSocket::SendData(). */
+  POOLMEM* as_socket_message() const
+  {
+    if (has_header) {
+      return const_cast<char*>(header_ptr());
+    } else {
+      return const_cast<char*>(data_ptr());
+    }
```
arogge, line 1128: `header_ptr()` and `data_ptr()` both return `char*` so the cast isn't required.

arogge (changes requested):
Looks pretty complicated, but great!
I have some minor remarks, am still not happy with the parameter naming and you have an endianess issue.

core/src/filed/backup.cc
sebsura: `header_ptr()` and `data_ptr()` return a const pointer if the object itself is const.  `as_socket_message` cannot return a const pointer because the `send()` function has to write the message size into it.  Thats why i need the const cast here.

core/src/lib/thread_pool.h
```diff
@@ -0,0 +1,143 @@
...
+  return fut;
+}
```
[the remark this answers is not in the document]
sebsura: yep

core/src/filed/filed_conf.cc
```diff
@@ -92,6 +92,7 @@ static ResourceItem cli_items[] = {
   {"PluginNames", CFG_TYPE_PLUGIN_NAMES, ITEM(res_client, plugin_names), 0, 0, NULL, NULL, NULL},
   {"ScriptsDirectory", CFG_TYPE_DIR, ITEM(res_client, scripts_directory), 0, 0, NULL, NULL, NULL},
   {"MaximumConcurrentJobs", CFG_TYPE_PINT32, ITEM(res_client, MaxConcurrentJobs), 0, CFG_ITEM_DEFAULT, "20", NULL, NULL},
+  {"MaximumWorkersPerJob", CFG_TYPE_PINT32, ITEM(res_client, MaxWorkerThreads), 0, CFG_ITEM_DEFAULT, "1", NULL, NULL},
```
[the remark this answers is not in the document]
sebsura: Good point.  Changed the variable name to `MaxWorkersPerJob`

core/src/filed/backup.cc
sebsura: You are right! Rebased onto master and used the new `network_order` stuff.

core/src/filed/backup.cc
sebsura: done

core/src/filed/backup.cc
sebsura: done

core/src/filed/backup.cc
arogge: I see. But then you're lying about `const`-ness which is way worse than casting it away to satisfy an interface.

arogge (approved):

core/src/filed/backup.cc
```diff
@@ -1406,6 +1411,11 @@ static inline bool SendPlainData(b_ctx& bctx)
     bctx.jcr->ReadBytes += bytes_read; /* count bytes read */
   }
   sd->msg = bctx.msgsave; /* restore read buffer */
+
+  if (read_error) {
+    // the following code recognizes read errors
+    sd->message_length = -1;
```
arogge, line 1417: um... aren't negative message lengths signals? If that's the case here maybe `BNET_EOD` would show the intent better.

sebsura: Its not supposed to be a signal.  Its just how the code communicates sadly :(  I did not want to add a major change now.  This should get changed in a coming pr.

sebsura: See here:
```
  if (sd->message_length < 0) { /* error */
    BErrNo be;
    Jmsg(jcr, M_ERROR, 0, _("Read error on file %s. ERR=%s\n"), ff_pkt->fname,
         be.bstrerror(ff_pkt->bfd.BErrNo));
    if (jcr->JobErrors++ > 1000) { /* insanity check */
      Jmsg(jcr, M_FATAL, 0, _("Too many errors. JobErrors=%d.\n"),
           jcr->JobErrors);
    }
[...]
```

sebsura: I added the missing second half of that comment.

PR #1539, opened 2023-08-29
stored: buffer fd messages

sebsura:

This pr adds an asynchronous message buffer to the sd.  This should reduce the time the file daemon has to wait when sending data to the storage daemon.

#### Please check

##### General
- [x] Check backport line

##### Source code quality

##### Tests

### DIFF
--- a/core/src/include/baconfig.h
+++ b/core/src/include/baconfig.h
@@ -164,7 +164,7 @@ void InitWinAPIWrapper();
 #define AUTH_TIMEOUT 60 * 10
 
 // Default network buffer size
-#define DEFAULT_NETWORK_BUFFER_SIZE (256 * 1024)
+inline constexpr std::size_t DEFAULT_NETWORK_BUFFER_SIZE = 256 * 1024;
 
 // Tape label types -- stored in catalog
 #define B_BAREOS_LABEL 0
@@ -249,12 +249,12 @@ typedef uint16_t slot_flags_t;
 
 #include <limits>
 
-constexpr slot_number_t kInvalidSlotNumber
+inline constexpr slot_number_t kInvalidSlotNumber
     = std::numeric_limits<slot_number_t>::max();
-constexpr slot_number_t kInvalidDriveNumber
+inline constexpr slot_number_t kInvalidDriveNumber
     = std::numeric_limits<drive_number_t>::max();
 
-constexpr int kInvalidFiledescriptor = -1;
+inline constexpr int kInvalidFiledescriptor = -1;
 
 inline bool IsSlotNumberValid(slot_number_t slot)
 {
--- a/core/src/include/baconfig.h
+++ b/core/src/include/baconfig.h
@@ -164,7 +164,7 @@ void InitWinAPIWrapper();
 #define AUTH_TIMEOUT 60 * 10
 
 // Default network buffer size
-#define DEFAULT_NETWORK_BUFFER_SIZE (64 * 1024)
+#define DEFAULT_NETWORK_BUFFER_SIZE (256 * 1024)
 
 // Tape label types -- stored in catalog
 #define B_BAREOS_LABEL 0
--- a/core/src/lib/bsock.h
+++ b/core/src/lib/bsock.h
@@ -179,6 +179,9 @@ class BareosSocket {
   virtual void RestoreBlocking(int flags) = 0;
   virtual bool ConnectionReceivedTerminateSignal() = 0;
   // Returns: 1 if data available, 0 if timeout, -1 if error
+  static inline constexpr int DataAvailable = 1;
+  static inline constexpr int Timeout = 0;
+  static inline constexpr int Error = -1;
   virtual int WaitData(int sec, int usec = 0) = 0;
   virtual int WaitDataIntr(int sec, int usec = 0) = 0;
   bool fsend(const char*, ...);
--- a/core/src/stored/append.cc
+++ b/core/src/stored/append.cc
@@ -190,7 +190,7 @@ class MessageHandler {
     POOLMEM* save = fd->msg;
     bool cont = true;
     for (int res = 0; cont; res = fd->WaitData(0, 100'000)) {
-      if (res == 1) {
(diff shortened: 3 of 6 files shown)


core/src/stored/append.cc
```diff
@@ -111,6 +118,164 @@ static bool SaveFullyProcessedFilesAttributes(
...
+                  "MsgQueue Message Count = %llu, Signal Count = %llu, Size = "
+                  "%llu\n",
+                  message_count, signal_count, bytes_in - bytes_out);
+          }
+          not_empty.notify_one();
+        }
+      } else if (res == -1) {
+        cont = false;
+        error_while_reading = true;
+      }
```
arogge, line 262: will `fd->WaitData()` only return `1` or `-1`?

core/src/stored/append.cc
```diff
@@ -111,6 +118,164 @@ static bool SaveFullyProcessedFilesAttributes(
...
+  // channel::out<result_type> chan;
+
+  std::mutex mut{};
+  bool finished{false};
+  std::size_t bytes_out{0};
+  std::condition_variable not_empty{};
+  std::condition_variable element_removed{};
+  std::deque<result_type> results{};
+
+  std::deque<result_type> cache;
```
arogge, line 191:
```suggestion
  std::deque<result_type> cache{};
```
(or remove the `{}` from all members that are guaranteed to be constructed somehow)

core/src/stored/append.cc
```diff
@@ -372,7 +590,10 @@ bool DoAppendData(JobControlRecord* jcr, BareosSocket* bs, const char* what)
   /* Check if we can still write. This may not be the case
    * if we are at the end of the tape or we got a fatal I/O error. */
   if (ok || dev->CanWrite()) {
-    jcr->JobFiles = file_currently_processed.GetFileIndex();
+    // take into account the fact that GetFileIndex() may return -1
+    // if no file is currently getting processed.
+    // This should only happen if no file was send to begin with!
+    jcr->JobFiles = std::max(file_currently_processed.GetFileIndex(), 0);
```
arogge, line 596: was there a bug you wanted/needed to fix here?

arogge (changes requested):
Looks pretty solid. Nevertheless, I have a few remarks.

core/src/stored/append.cc
sebsura: Yes, but only a visual one.   If you start a job that does not back up any files (for some reason) then the joblog will say that the fd send 0 files but the sd wrote 4 billion of them.

core/src/stored/append.cc
```diff
@@ -111,6 +118,164 @@ static bool SaveFullyProcessedFilesAttributes(
...
+          bytes_in += sizeof(result);
+
+          element_removed.wait(l, [this, bytes_in] {
+            return bytes_in - bytes_out < 5ULL * 1024ULL * 1024ULL
+                   || end.load();
+          });
+
+          results.emplace_back(std::move(result));
+          if (results.size() > max_size) { max_size = results.size(); }
+          ASSERT(bytes_in >= full_message_count * 64ULL * 1024ULL);
```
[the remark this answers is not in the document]
sebsura: Note to self: do not hard code 64k. replace by network buffer size.

core/src/stored/append.cc
sebsura: maybe `WaitData()` should return an enum instead ? What do you think ?

core/src/include/baconfig.h
```diff
@@ -171,7 +171,7 @@ void InitWinAPIWrapper();
 #define AUTH_TIMEOUT 60 * 10
 
 // Default network buffer size
-#define DEFAULT_NETWORK_BUFFER_SIZE (64 * 1024)
+#define DEFAULT_NETWORK_BUFFER_SIZE (256 * 1024)
```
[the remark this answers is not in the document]
sebsura: Yes.   Since the change was small and we wanted to do it somewhere.

core/src/lib/bsock.h
```diff
@@ -179,6 +179,9 @@ class BareosSocket {
   virtual void RestoreBlocking(int flags) = 0;
   virtual bool ConnectionReceivedTerminateSignal() = 0;
   // Returns: 1 if data available, 0 if timeout, -1 if error
+  static inline constexpr int DataAvailable = 1;
+  static inline constexpr int Timeout = 0;
+  static inline constexpr int Error = -1;
```
arogge, line 184: if I'm not mistaken that `inline` is redundant, but of course not wrong.

core/src/stored/append.cc
```diff
@@ -342,18 +520,19 @@ bool DoAppendData(JobControlRecord* jcr, BareosSocket* bs, const char* what)
     // Restore the original data pointer.
     jcr->sd_impl->dcr->rec->data = rec_data;
 
-    if (bs->IsError()) {
+    if (auto* error = handler.error()) {
```
arogge, line 523: as `error` is passed to `Jmsg2()` which isn't type-checked at all, I would feel safer with a `const char*` instead of an `auto*`

arogge (approved):
Looks good. I have two minor remarks that you can address if (and *only* if) you need to touch the commits again for some reason.

core/src/lib/bsock.h
sebsura: Ill have to check again, but at least there is a difference between `static` and `static inline`.  Not sure if `constexpr` makes those the same again.

core/src/stored/append.cc
sebsura: Makes sense!

core/src/lib/bsock.h
sebsura: https://stackoverflow.com/a/54467409 global variables should be `inline constexpr` to fix odr related issues that might arise.   `static` here is just because they are defined inside the class scope.

sebsura: Seems like constexpr member variables are always inline.  Thanks for the hint!

PR #1665, opened 2024-01-19
filed: fix vss during client initiated connections

sebsura:

Using client initiated connections lead to the filedaemon trying to open files on the live system and not on the shadow copy.   This was caused by not setting up the thread local jcr pointer correctly.  This pr fixes that issue.

It also includes:
* Bad format specifier for jobid in a jmsg
* list additional files to include/exclude in dmsgs.
* removing the double newline in some windows error messages.

#### Please check

##### General
- [x] Check backport line

##### Source code quality

### DIFF
--- a/core/src/filed/dir_cmd.cc
+++ b/core/src/filed/dir_cmd.cc
@@ -3,7 +3,7 @@
 
    Copyright (C) 2000-2011 Free Software Foundation Europe e.V.
    Copyright (C) 2011-2012 Planets Communications B.V.
-   Copyright (C) 2013-2023 Bareos GmbH & Co. KG
+   Copyright (C) 2013-2024 Bareos GmbH & Co. KG
 
    This program is Free Software; you can redistribute it and/or
    modify it under the terms of version three of the GNU Affero General Public
@@ -398,9 +398,12 @@ JobControlRecord* create_new_director_session(BareosSocket* dir)
   return jcr;
 }
 
-void* process_director_commands(void* p_jcr)
+void* process_fd_initiated_director_commands(void* p_jcr)
 {
   JobControlRecord* jcr = (JobControlRecord*)p_jcr;
+
+  SetJcrInThreadSpecificData(jcr);
+
   return process_director_commands(jcr, jcr->dir_bsock);
 }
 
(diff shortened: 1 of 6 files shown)


core/src/filed/dir_cmd.cc
```diff
@@ -398,9 +398,12 @@ JobControlRecord* create_new_director_session(BareosSocket* dir)
   return jcr;
 }
 
-void* process_director_commands(void* p_jcr)
+void* process_initiated_director_commands(void* p_jcr)
```
pstorz, line 401:
```suggestion
void* process_incoming_director_commands(void* p_jcr)
```

pstorz: Does this name fit to what the function does?

core/src/win32/filed/vss_generic.cc
```diff
@@ -534,6 +536,176 @@ bool VSSClientGeneric::Initialize(DWORD dwContext, bool bDuringRestore)
       errno = b_errno_win32;
       return false;
     }
+
+#  define VSS_CALL(Obj, Name, ...)                                 \
```
pstorz, line 540: Why is it now required to do an loop here?

pstorz (changes requested):

core/src/win32/filed/vss_generic.cc
sebsura: which loop ?  Do you mean the `do { ... } while (0)` construct ? Thats used to turn some code of block into a statement.  One should always do that for function like macros (if they arent statements by themselves already).

Otherwise the semicolon afterwards becomes optional.

core/src/filed/dir_cmd.cc
sebsura: hmm... technically yes but director commands are always incoming (even in the normal other process_directory_commands function).  This was supposed to distinguish between commands coming from a director that opened the connection vs a director where the fd initiated the connection.

sebsura: Im not 100% happy with the name either but i could not think of a better name. Maybe one could call it
```
process_director_commands_after_initiating_connection
```
but that also does not fit 100%.

core/src/win32/filed/vss_generic.cc
sebsura: See for example here for reference: https://stackoverflow.com/questions/2687569/c-do-while0

core/src/filed/dir_cmd.cc
pstorz: > hmm... technically yes but director commands are always incoming (even in the normal other process_directory_commands function). This was supposed to distinguish between commands coming from a director that opened the connection vs a director where the fd initiated the connection.

How about `process_fd_initiated_director_commands` ?

pstorz (commented):
Please see updated comments

core/src/filed/dir_cmd.cc
sebsura: sounds good to me!

pstorz (approved):
Good work!

PR #1689, opened 2024-01-29
Spring Cleanup 2024

arogge:
This PR contains mostly some refactoring.

* enable `-Wshadow` and refactor problematic code
* refactor BareosConfigureFile
* generate files for core/plaforms in output directory

#### Please check

##### General
- [ ] Check backport line

##### Source code quality

##### Tests


core/src/tests/test_bsnprintf.cc
```diff
@@ -170,8 +170,8 @@ TEST(bsnprintf, integers)
   EXPECT_EQ(Bsnprintf(dest, 100, "%zi", ss_int), 4);
   EXPECT_STREQ(dest, "-123");
 
-  unsigned int uint = 123;
-  EXPECT_EQ(Bsnprintf(dest, 100, "%u", uint), 3);
+  unsigned int uint2 = 123;
+  EXPECT_EQ(Bsnprintf(dest, 100, "%u", uint2), 3);
```
sebsura, line 174:
```suggestion
  unsigned int uns_int = 123;
  EXPECT_EQ(Bsnprintf(dest, 100, "%u", uns_int), 3);
```

core/src/tests/multiplied_device_test.cc
```diff
@@ -149,28 +149,28 @@ TEST(sd, MultipliedDeviceTest_CheckNames_1)
...
 
-  auto count = CheckNamesOfConfiguredDeviceResources_1(*my_config);
+  auto count = CheckNamesOfConfiguredDeviceResources_1(*test_config);
 
   EXPECT_EQ(count, 3);
 }
 
 static uint32_t CheckNamesOfConfiguredDeviceResources_2(
-    ConfigurationParser& my_config)
+    ConfigurationParser& t_config)
```
sebsura, line 163:
```suggestion
    ConfigurationParser& config)
```

core/src/tests/multiplied_device_test.cc
```diff
@@ -87,37 +87,37 @@ TEST(sd, MultipliedDeviceTest_CountAllAutomaticallyCreatedResources)
...
                                         const char* name)
 {
-  BareosResource* p = my_config.GetResWithName(R_DEVICE, name, false);
+  BareosResource* p = t_config.GetResWithName(R_DEVICE, name, false);
   return dynamic_cast<DeviceResource*>(p);
 }
 
 static uint32_t CheckNamesOfConfiguredDeviceResources_1(
-    ConfigurationParser& my_config)
+    ConfigurationParser& t_config)
```
sebsura, line 109:
```suggestion
    ConfigurationParser& config)
```

core/src/tests/multiplied_device_test.cc
```diff
@@ -87,37 +87,37 @@ TEST(sd, MultipliedDeviceTest_CountAllAutomaticallyCreatedResources)
...
+  ASSERT_TRUE(test_config->ParseConfig());
+  auto count = CountAllDeviceResources(*test_config);
 
   /* configurable using Device config for "AnotherMultipliedDeviceResource" */
   int amount_to_check = 103;
   EXPECT_EQ(count, amount_to_check);
 }
 
-DeviceResource* GetDeviceResourceByName(ConfigurationParser& my_config,
+DeviceResource* GetDeviceResourceByName(ConfigurationParser& t_config,
```
sebsura, line 101:
```suggestion
DeviceResource* GetDeviceResourceByName(ConfigurationParser& config,
```

core/src/tests/multiplied_device_test.cc
```diff
@@ -60,21 +60,21 @@ TEST(sd, MultipliedDeviceTest_ConfigParameter)
...
-  auto d = GetMultipliedDeviceResource(*my_config);
+  ASSERT_TRUE(test_config->ParseConfig());
+  auto d = GetMultipliedDeviceResource(*test_config);
   ASSERT_TRUE(d);
 
   EXPECT_EQ(d->count, 3);
 }
 
-static uint32_t CountAllDeviceResources(ConfigurationParser& my_config)
+static uint32_t CountAllDeviceResources(ConfigurationParser& t_config)
```
sebsura, line 73:
```suggestion
static uint32_t CountAllDeviceResources(ConfigurationParser& config)
```

core/src/tests/multiplied_device_test.cc
```diff
@@ -43,11 +43,11 @@ static void InitGlobals()
 }
 
 static DeviceResource* GetMultipliedDeviceResource(
-    ConfigurationParser& my_config)
+    ConfigurationParser& t_config)
```
sebsura, line 46:
```suggestion
    ConfigurationParser& config)
```

core/src/stored/stored_conf.cc
```diff
@@ -634,9 +632,10 @@ bool PrintConfigSchemaJson(PoolMem& buffer)
   json_t* bareos_sd = json_object();
   json_object_set_new(resource, "bareos-sd", bareos_sd);
 
-  for (int r = 0; resources[r].name; r++) {
-    ResourceTable resource = my_config->resource_definitions_[r];
-    json_object_set_new(bareos_sd, resource.name, json_items(resource.items));
+  for (int r = 0; my_config->resource_definitions_[r].name; r++) {
+    ResourceTable resource_table = my_config->resource_definitions_[r];
```
sebsura, line 636:
```suggestion
    ResourceTable& resource_table = my_config->resource_definitions_[r];
```

core/src/stored/stored_conf.cc
```diff
@@ -543,15 +543,15 @@ static void GuessMissingDeviceTypes(ConfigurationParser& my_config)
   }
 }
 
-static void CheckAndLoadDeviceBackends(ConfigurationParser& my_config)
+static void CheckAndLoadDeviceBackends(ConfigurationParser& t_config)
```
sebsura, line 546:
```suggestion
static void CheckAndLoadDeviceBackends(ConfigurationParser& config)
```

core/src/stored/stored_conf.cc
```diff
@@ -460,16 +460,16 @@ static void MultiplyDevice(DeviceResource& multiplied_device_resource)
   }
 }
 
-static void MultiplyConfiguredDevices(ConfigurationParser& my_config)
+static void MultiplyConfiguredDevices(ConfigurationParser& t_config)
```
sebsura, line 463:
```suggestion
static void MultiplyConfiguredDevices(ConfigurationParser& config)
```

core/src/stored/stored_conf.cc
```diff
@@ -480,14 +480,14 @@ static void ConfigBeforeCallback(ConfigurationParser& my_config)
       {R_DEVICE, "R_DEVICE"},
       {R_AUTOCHANGER, "R_AUTOCHANGER"},
       {R_CLIENT, "R_CLIENT"}}; /* needed for network dump */
-  my_config.InitializeQualifiedResourceNameTypeConverter(map);
+  t_config.InitializeQualifiedResourceNameTypeConverter(map);
 }
 
-static void CheckDropletDevices(ConfigurationParser& my_config)
+static void CheckDropletDevices(ConfigurationParser& t_config)
```
sebsura, line 486:
```suggestion
static void CheckDropletDevices(ConfigurationParser& config)
```

core/src/stored/stored_conf.cc
```diff
@@ -508,11 +508,11 @@ static void CheckDropletDevices(ConfigurationParser& my_config)
   }
 }
 
-static void GuessMissingDeviceTypes(ConfigurationParser& my_config)
+static void GuessMissingDeviceTypes(ConfigurationParser& t_config)
```
sebsura, line 511:
```suggestion
static void GuessMissingDeviceTypes(ConfigurationParser& config)
```

core/src/stored/sd_stats.cc
```diff
@@ -513,7 +513,7 @@ bool StatsCmd(JobControlRecord* jcr)
 
   if (job_statistics) {
     bool found;
-    JobControlRecord* jcr;
+    JobControlRecord* my_jcr;
```
sebsura, line 516: The jcr is only used for the loop below.  We should probably move the definition there as well (and maybe wrap it in a new scope).

core/src/stored/sd_stats.cc
```diff
@@ -513,7 +513,7 @@ bool StatsCmd(JobControlRecord* jcr)
 
   if (job_statistics) {
```
sebsura, line 514: This ~~function~~ file does not make sense to me.

Why do we lock job_statistics when we remove something but not when we iterate over it ? This cannot be right.
Looking at it, this is also the case for the job_stats stuff.

This probably shouldnt get fixed in this pr though.

core/src/stored/reserve.cc
```diff
@@ -244,9 +244,9 @@ static bool UseDeviceCmd(JobControlRecord* jcr)
   if (rctx.append) { jcr->sd_impl->dcr->SetWillWrite(); }
 
   if (!jcr->sd_impl->dcr) {
-    BareosSocket* dir = jcr->dir_bsock;
-    dir->fsend(T_("3939 Could not get dcr\n"));
-    Dmsg1(debuglevel, ">dird: %s", dir->msg);
+    BareosSocket* dir2 = jcr->dir_bsock;
```
sebsura, line 247: This looks like a c&p error to me.   `dir` already is `jcr->dir_bsock` and i dont see where `jcr->dir_bsock` would get changed in this function that we would need to reread it.

core/src/stored/backends/win32_tape_device.cc
```diff
@@ -105,20 +105,20 @@ static inline ULONG Read32BitUnsigned(const unsigned char* pValue)
...
-  return (LONGLONG)(
-      ((ULONGLONG)pValue[0] << 56) | ((ULONGLONG)pValue[1] << 48)
-      | ((ULONGLONG)pValue[2] << 40) | ((ULONGLONG)pValue[3] << 32)
-      | ((ULONGLONG)pValue[4] << 24) | ((ULONGLONG)pValue[5] << 16)
-      | ((ULONGLONG)pValue[6] << 8) | (ULONGLONG)pValue[7]);
+  return (
+      LONGLONG)(((ULONGLONG)pValue[0] << 56) | ((ULONGLONG)pValue[1] << 48)
+                | ((ULONGLONG)pValue[2] << 40) | ((ULONGLONG)pValue[3] << 32)
+                | ((ULONGLONG)pValue[4] << 24) | ((ULONGLONG)pValue[5] << 16)
+                | ((ULONGLONG)pValue[6] << 8) | (ULONGLONG)pValue[7]);
```
sebsura, line 112: Splitting up the cast looks weird, but i guess this was done by bareos-check-sources so its fine.

core/src/stored/backends/win32_tape_device.cc
```diff
@@ -265,20 +265,20 @@ int win32_tape_device::d_open(const char* pathname, int, int)
   return (int)idxFile + 3;
 }
 
-ssize_t win32_tape_device::d_read(int fd, void* buffer, size_t count)
+ssize_t win32_tape_device::d_read(int t_fd, void* buffer, size_t count)
```
sebsura, line 268:
```suggestion
ssize_t win32_tape_device::d_read(int table_index, void* buffer, size_t count)
```

I think this is a more fitting name. Do you know why the `+3` / `-3` stuff is done ? Is it "just" for better error detection ( so that 0/1 arent valid ids ) ?

core/src/stored/bcopy.cc
```diff
@@ -360,10 +360,10 @@ static bool RecordCb(DeviceControlRecord* in_dcr, DeviceRecord* rec)
 
 static void GetSessionRecord(Device* dev,
                              DeviceRecord* rec,
-                             Session_Label* sessrec)
+                             Session_Label* t_sessrec)
```
sebsura, line 363: Its probably better to rename the global one to `g_sessrec`

core/src/stored/bextract.cc
```diff
@@ -479,17 +479,17 @@ static void DoExtract(char* devname,
   return;
 }
 
-static bool StoreData(BareosFilePacket* bfd, char* data, const int32_t length)
+static bool StoreData(BareosFilePacket* t_bfd, char* data, const int32_t length)
```
sebsura, line 482: Its probably better to rename the global one to `g_bfd`

sebsura: Same goes for the other renames.

core/src/stored/btape.cc
```diff
@@ -208,9 +208,9 @@ int main(int margc, char* margv[])
 
   uint32_t x32 = 123456789;
```
sebsura, line 209: I feel like these tests should be part of our test suite and not part of btape.

Why does btape care for scanf/printf anyways ?

core/src/stored/btape.cc
```diff
@@ -2064,7 +2069,7 @@ static void statcmd()
 {
   char* status;
 
-  status = dev->StatusDev();
+  status = g_dev->StatusDev();
```
sebsura, line 2072:
```suggestion
  char* status = g_dev->StatusDev();
```

sebsura (changes requested):
I like the new warnings, but in general i feel like we should not rename local variables of functions but instead make it clearer that a global variable is global by renaming the global variable.

core/src/lib/tls_openssl.cc
```diff
@@ -139,7 +139,6 @@ bool TlsOpenSsl::TlsPostconnectVerifyCn(
     char data[256]; /* nullterminated by X509_NAME_get_text_by_NID */
     if (X509_NAME_get_text_by_NID(subject, NID_commonName, data, sizeof(data))
         > 0) {
-      std::string cn;
       for (const std::string& cn : verify_list) {
         std::string d(data);
```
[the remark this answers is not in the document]
arogge: yes. probably doesn't make sense to re-initialize `d` every time. I'd make it a `const std::string` though.

core/src/stored/backends/win32_tape_device.cc
arogge: that is probably to avoid the first three filehandles (i.e. stdin, stdout, stderr)

core/src/stored/btape.cc
arogge: bscan is the only user of `bsscanf()` and that seems to have had a bug some time in the past...

core/src/lib/bregex.cc
```diff
@@ -64,7 +64,7 @@
  * always true.  Original instances of "(short)x" were replaced by
  * SHORT(x), where SHORT is #defined below.  */
 
-#define SHORT(x) ((x)&0x8000 ? (x)-0x10000 : (x))
+#define SHORT(x) ((x) & 0x8000 ? (x)-0x10000 : (x))
```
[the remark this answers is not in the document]
arogge: or it should just declare `a` and `b` as int16_t. But then overflow would be undefined.
However, the original code is so old that it pre-dates fixed-width integer types and as the macro behaves different from a cast to int16_t and I'm not really interested in understanding that ancient regexp machine, I'd just leave it untouched tbh.

core/src/lib/bregex.cc
```diff
@@ -911,16 +911,20 @@ static int ReOptimize(regex_t* bufp)
...
 
-#define INSERT_JUMP(pos, type, addr)                                       \
-  {                                                                        \
-    int a, p = (pos), t = (type), ad = (addr);                             \
-    for (a = pattern_offset - 1; a >= p; a--) pattern[a + 3] = pattern[a]; \
-    pattern[p] = t;                                                        \
-    PUT_ADDR(p + 1, ad);                                                   \
-    pattern_offset += 3;                                                   \
+#define INSERT_JUMP(pos, type, addr)                                     \
+  {                                                                      \
```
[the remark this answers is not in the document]
arogge: while I totally agree that `do { ... } while(0);` is the correct pattern here, we either need to apply it to all macros in bregex.cc or to none.
Right now I'd go with none.

core/src/stored/backends/win32_tape_device.cc
arogge: and also I disagree. That should *not* be renamed as the `int` doesn't carry the meaning that this is supposed to be a file-descriptor by itself, so it should at least contain `fd`.

core/src/stored/btape.cc
arogge: ok... I just realized we have `#define sscanf bsscanf` in baconfig.h. We also have `#define fgets bfgets` so we have the "safer" home-grown functions for that...
I'd say we leave it like that for now and remove `bsscanf()` and `bfgets()` as soon as we have proper tests for those.

core/src/stored/sd_stats.cc
arogge: extracted into function

arogge:
Please take another look. I have addressed most of your requested changes and left comments on the others.

core/src/stored/btape.cc
sebsura: I already tried once to switch to sscanf instead of bscanf.  Sadly sscanf handles some things differently than bsscanf.  Iirc the biggest problem was that sscanf does not match empty strings, e.g.
```
sscanf("volume=%127s other=%d\n", "volume= other=5\n", &vol, &other) == 0
```
but
```
bsscanf("volume=%127s other=%d\n", "volume= other=5\n", &vol, &other) == 2
```
i.e. our message parsing will break with that change.  Im not sure how much effort it would be to fix this, especially if we want forwards/backwards compatibilty.

core/src/stored/backends/win32_tape_device.cc
sebsura: Maybe we are talking beside each other, but fd is not a file descriptor.  Its an index into a table of HANDLEs, this is why i proposed `table_index` as an alternative name.

While its true that behind the curtain filedescriptors are basically just an index into an array,  I would suspect that calling our own indices file descriptors is more confusing than it is helpful.

core/src/lib/tls_openssl.cc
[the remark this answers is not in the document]
sebsura: Shouldnt make a big difference.  I proposed string_view to prevent an allocation but in practice this should not matter.  Take care about the dmsg right after this as well.  We probably want to change it to data.c_str() there if you use std::string (with std::string_view this also works, but you have to change the format specifier accordingly).

core/src/dird/dbcheck.cc
```diff
@@ -715,7 +716,7 @@ static void repair_bad_paths()
       db->EscapeString(nullptr, esc_name, name, len);
       Bsnprintf(buf, sizeof(buf), "UPDATE Path SET Path='%s' WHERE PathId=%s",
                 esc_name, edit_int64(id_list.Id[i], ed1));
-      if (verbose > 1) { printf("%s\n", buf); }
+      if (g_verbose > 1) { printf("%s\n", buf); }
```
sebsura, line 719: I think verbose can currently only be set to 0 or 1, right ?
We should probably create a task that allows us to specify verbosity levels somehow, assuming that it was previously possible to do so.

core/src/lib/bareos_resource.h
```diff
@@ -3,7 +3,7 @@
 
    Copyright (C) 2000-2010 Free Software Foundation Europe e.V.
    Copyright (C) 2011-2012 Planets Communications B.V.
-   Copyright (C) 2013-2023 Bareos GmbH & Co. KG
+   Copyright (C) 2013-2024 Bareos GmbH & Co. KG
```
sebsura, line 6: This needs to be reverted.

core/src/lib/bsock_tcp.cc
```diff
@@ -2,7 +2,7 @@
    BAREOS® - Backup Archiving REcovery Open Sourced
 
    Copyright (C) 2007-2011 Free Software Foundation Europe e.V.
-   Copyright (C) 2013-2023 Bareos GmbH & Co. KG
+   Copyright (C) 2013-2024 Bareos GmbH & Co. KG
```
sebsura, line 5: revert

core/src/lib/messages_resource.h
```diff
@@ -3,7 +3,7 @@
 
    Copyright (C) 2000-2010 Free Software Foundation Europe e.V.
    Copyright (C) 2011-2012 Planets Communications B.V.
-   Copyright (C) 2013-2021 Bareos GmbH & Co. KG
+   Copyright (C) 2013-2024 Bareos GmbH & Co. KG
```
sebsura, line 6: revert

sebsura (changes requested):
Thanks for the changes.  I added some comments as well as adding some comments to the previous (open) conversations as well.

core/src/stored/backends/win32_tape_device.cc
arogge: I understand your point.
This is the implementation for the Windows Tape backend. Windows specifically does *not* use file descriptors to access tapes, but Bareos wants to pass a file descriptor around.
So while this is "just" and index to a table internally, the API that we implement mandates that it is a file descriptor.
And also, I don't see why this specific implementation of `d_read()` should have a signature that differs from the other 8 implementations.

core/src/lib/messages_resource.h
arogge: that's going to happen when the fixups are squashed and bareos-check-sources is re-applied

core/src/lib/bsock_tcp.cc
arogge: that's going to happen when the fixups are squashed and bareos-check-sources is re-applied

core/src/lib/bareos_resource.h
arogge: that's going to happen when the fixups are squashed and bareos-check-sources is re-applied

core/src/cats/cats.h
```diff
@@ -3,7 +3,7 @@
 
    Copyright (C) 2000-2012 Free Software Foundation Europe e.V.
    Copyright (C) 2011-2016 Planets Communications B.V.
-   Copyright (C) 2013-2023 Bareos GmbH & Co. KG
+   Copyright (C) 2013-2024 Bareos GmbH & Co. KG
```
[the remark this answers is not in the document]
arogge: that's going to happen when the fixups are squashed and bareos-check-sources is re-applied

core/src/dird/ua_output.cc
```diff
@@ -625,28 +625,27 @@ static bool DoListCmd(UaContext* ua, const char* cmd, e_list_type llist)
...
-  int hours = 0;
-  if (argument > 0) {
-    hours = str_to_int64(ua->argv[argument]);
-    schedtime = now - secs_in_hour * hours; /* Hours in the past */
+  {
+    int hours = 0;
+    if (int argument = FindArgWithValue(ua, NT_("hours")); argument > 0) {
+      hours = str_to_int64(ua->argv[argument]);
+      schedtime = now - secs_in_hour * hours; /* Hours in the past */
+    }
```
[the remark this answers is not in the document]
arogge: Every use of `FindArgWithValue()` should check the return code for `>= 0` or `!= -1`.
We could also have `const char* GetArgValue(UaContext*, const char*)` and have that return a pointer to the value or `nullptr`. Maybe that would be harder to use wrong.

Then we could do it like this:
```
if (const char* argument = GetArgValue(ua, NT_("hours"))) {
  hours = str_to_int64(argument);
  schedtime = now - secs_in_hour * hours; /* Hours in the past */
}
```

core/src/dird/dbcheck.cc
arogge: It has been like this since we forked. So we can surely fix this, but I'd just make `g_verbose` a bool and remove all the code that is guarded with `g_verbose > 1`.

sebsura (approved):
Thanks for the work! Looks good to me!

PR #1721, opened 2024-02-28
Fix warning about missing delcandidates table in director

smarsching:
Pruning jobs would cause the following warning in the output of the director (e.g. visible through `systemctl status bareos-director.service`):

    table "delcandidates" does not exist, skipping

This warning is harmless but clutters the output. With this PR, this problem is fixed by disabling warnings for the PostgreSQL connection that is established by the director.

See https://bugs.bareos.org/view.php?id=1561 and https://bugs.bareos.org/view.php?id=1606.

#### Please check

##### General
- [x] Check backport line

##### Source code quality

### DIFF
--- a/core/src/cats/postgresql.cc
+++ b/core/src/cats/postgresql.cc
@@ -238,6 +238,7 @@ bool BareosDbPostgresql::OpenDatabase(JobControlRecord* jcr)
 
   SqlQueryWithoutHandler("SET datestyle TO 'ISO, YMD'");
   SqlQueryWithoutHandler("SET cursor_tuple_fraction=1");
+  SqlQueryWithoutHandler("SET client_min_messages TO WARNING");
 
   /* Tell PostgreSQL we are using standard conforming strings
    * and avoid warnings such as:
--- a/core/src/cats/postgresql.cc
+++ b/core/src/cats/postgresql.cc
@@ -3,7 +3,7 @@
 
    Copyright (C) 2003-2011 Free Software Foundation Europe e.V.
    Copyright (C) 2011-2016 Planets Communications B.V.
-   Copyright (C) 2013-2023 Bareos GmbH & Co. KG
+   Copyright (C) 2013-2024 Bareos GmbH & Co. KG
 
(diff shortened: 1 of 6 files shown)


core/src/cats/postgresql.cc
```diff
@@ -205,9 +205,9 @@ bool BareosDbPostgresql::OpenDatabase(JobControlRecord* jcr)
 
   // If connection fails, try at 5 sec intervals for 30 seconds.
   for (int retry = 0; retry < 6; retry++) {
-    db_handle_ = PQsetdbLogin(db_address_,   /* default = localhost */
-                              port,          /* default port */
-                              NULL,          /* pg options */
+    db_handle_ = PQsetdbLogin(db_address_, /* default = localhost */
+                              port,        /* default port */
+                              "-c client_min_messages=warning", /* pg options */
```
arogge, line 210: I'm not really happy with doing this right here.

We already have code that sets several options during connect (see lines 239-245). I'd prefer doing a `SET client_min_messages TO WARNING;` there.

arogge (changes requested):
Thank you very much for taking the time working on this.
While the solution is probably correct (i.e. setting client_min_messages to WARNING), I would want you to implement it differently.

core/src/cats/postgresql.cc
smarsching: Sure, I just force-pushed a new commit that implements it the other way.

arogge:
I really don't know why we have three instances of that initialization code, but I'll try to replicate your change to the others.

smarsching:
Sorry, I force pushed the change a few weeks ago but forgot to write comment. Please have a look whether it now is now done at the correct location.

arogge (approved):
Great work. Thanks for taking the time to do this!

I removed an unused function (that would have require the same change) and added your change to the reconnect initialization, too.
If you're still happy with it, I'll accept the changes and backport them to Bareos 23.

smarsching:
Yes, this looks good to me.

PR #1724, opened 2024-03-01
[percona-xtrabackup] prevent High memory usage for no reason (IO_CLOSE)

banuchka:
In the case of IO_CLOSE inside the plugin, we don't need to wait for the subprocess to finish, because it trying to get the data from **xtrabackup**, but the destination for such data to backup is unavailable. 
It causes high memory usage and OOM in some cases.

#### Please check

##### General
- [ ] Check backport line

##### Source code quality

##### Tests


sduehr:
Hi, why did you close this PR?
It looks good so far and seems to make sense. I just need to find time to review it properly.
Your contribution is welcome. I will reopen it if you don't mind. If you still want to close it,
it would be nice if you drop a sentence why.

banuchka:
Hi @sduehr,
I've found an issue with the restore operation. I want to add more logic and checks before killing the subprocess.
Could you please put the review on hold until additional information or changes from my side?
Does it work for you? 
Thanks

sduehr:
Sure, I've set the on-hold label.

We have a systemtest for the plugin, see
https://github.com/bareos/bareos/tree/master/systemtests/tests/py3plug-fd-percona-xtrabackup
which is running automatically, it failed on restore with

```
xtrabackup: recognized client arguments: --prepare=1 --target-dir=./tmp/bareos-restores/_percona/4/00000000000000000000_00000000000018129147_0000000001 
/usr/bin/xtrabackup version 8.0.26-18 based on MySQL server 8.0.26 Linux (x86_64) (revision id: 4aecf82)
xtrabackup: cd to /var/lib/jenkins/workspace/bareos_PR-1724/build/systemtests/tests/py3plug-fd-percona-xtrabackup/tmp/bareos-restores/_percona/4/00000000000000000000_00000000000018129147_0000000001/
xtrabackup: Error: cannot open ./xtrabackup_checkpoints
xtrabackup: Error: failed to read metadata from './xtrabackup_checkpoints'
exit(1) is called. Set test to failure and end test.
```

The systemtest is probably too specific to be used outside of the build environment, but
we are planning to allow contributors access to the test results in the future.

Thanks again, waiting for your notification start review.

Regards,
Stephan

sduehr:
Thanks, build and tests are automatically triggered when new commits are pushed.
The systemtest for the percona-xtrabackup plugin did run successful now.
I'll probably have to suggest some changes during review, as it's required to comply with our guidelines to pass checks for PRs to get merged. For example the first line of a commit message must not be longer than 50 characters, see
https://docs.bareos.org/DeveloperGuide/gitworkflow.html

banuchka:
I rewrote the commit message and added details and explanations about the importance of the changes.

sduehr (changes requested):
Did you consider trying `self.stream.terminate()` first, and only if that fails use `self.stream.kill()`?

banuchka:
> Did you consider trying `self.stream.terminate()` first, and only if that fails use `self.stream.kill()`?

Good idea. Honestly, I wrote a separate function for the dump process termination with timeouts and TERM signal instead of killing it straightaway.
Can I do that as an additional commit to the current PR or force push(again), which is the preferred way?
I'm asking because you've already started reviewing it, and I don't want to waste your time.

core/src/plugins/filed/python/percona-xtrabackup/bareos-fd-percona-xtrabackup.py
```diff
@@ -398,6 +403,20 @@ def plugin_io(self, IOP):
             self.subprocess_returnCode = self.stream.poll()
             if self.subprocess_returnCode is None:
                 # Subprocess is open, we wait until it finishes and get results
+                if self.jobType == 'B':
+                    # In case of Backup Job terminate subprocess to prevent high memory usage, because destination is already unavailable
+                    DebugMessage(100, "terminame dump process with PID %s\n" % self.stream.pid)
+                    JobMessage(
+                        M_FATAL,
+                        "plugin_io[IO_CLOSE]: Network destination is unavailable, terminate dump process.\n",
```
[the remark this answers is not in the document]
banuchka: xtrabackup runs on the host you're right here, but it sends the output to bareos-sd via a pipe.
The first time when I encountered the issue, the unreachable bareos-sd caused it during the backup process with the log messages like the one below:
`Fatal error: filed/backup.cc:938 Network sends error to SD. ERR=Connection reset by peer.`
That is why my first idea was to drop a message to the job log about the unavailability of the destination.
Later, I found cases where the **IO_CLOSE** event wasn't caused by the remote unavailability. I completely removed those messages from the code.

sduehr:
Thanks, normally an additional commit is ok. But in this case it's only one file, so force push is ok, too.
I'll review again as soon as possible.

sduehr:
Thanks, looks good so far.
I pushed a small change to comply with our coding guidelines.

core/src/plugins/filed/python/percona-xtrabackup/bareos-fd-percona-xtrabackup.py
```diff
@@ -522,5 +564,56 @@ def restore_object_data(self, ROP):
...
+                if not sent_sigterm:
+                    DebugMessage(
+                        100,
+                        "sending SIGTERM to bareos_xtrabackup_dumper PID %s\n"
+                        % (self.stream.pid),
+                    )
+                    os.kill(self.stream.pid, signal.SIGTERM)
+                    sent_sigterm = True
+                    # Wait timeout after sending TERM to the process
+                    timeout = 30
```
sduehr, line 592:
```suggestion
```
This line can be deleted, because the timeout variable is never altered anywhere.

core/src/plugins/filed/python/percona-xtrabackup/bareos-fd-percona-xtrabackup.py
```diff
@@ -522,5 +564,56 @@ def restore_object_data(self, ROP):
...
+                    start_time = int(time.time())
+                    continue
+                else:
+                    DebugMessage(
+                        100,
+                        "Giving up to wait for bareos_xtrabackup_dumper PID %s to terminate\n"
+                        % (self.stream.pid),
+                    )
+                    break
```
sduehr, line 601:
```suggestion
                    # Set return code to 9 because we've tried to terminate subprocess earlier
                    self.stream.returncode = 9
                    break
```
I think this is missing here, otherwise it could happen that end_dumper() would return None,
and then the line after calling `self.end_dumper()` would fail because it compares it using `!= 0`

core/src/plugins/filed/python/percona-xtrabackup/bareos-fd-percona-xtrabackup.py
```diff
@@ -394,27 +400,57 @@ def plugin_io(self, IOP):
...
+                        (
+                            "plugin_io[IO_CLOSE]: bareos_xtrabackup_dumper returncode:"
+                            " %s\n"
+                        )
+                        % (bareos_xtrabackup_dumper_returncode),
+                    )
+
+                    # Cleanup tmpdir
+                    shutil.rmtree(self.tempdir)
+                    # Set return code to 9 because we've terminated subprocess earlier
```
sduehr, line 424:
```suggestion
```
Seems to be obsolete here, should be done in `end_dumper()`, see suggestion below.

core/src/plugins/filed/python/percona-xtrabackup/bareos-fd-percona-xtrabackup.py
```diff
@@ -394,27 +400,57 @@ def plugin_io(self, IOP):
...
+                    self.subprocess_returnCode = bareos_xtrabackup_dumper_returncode
+                    self.stream.returncode = bareos_xtrabackup_dumper_returncode
```
sduehr, line 426:
```suggestion
```
Should be done in `end_dumper()`, see suggestion below.

sduehr (changes requested):
Please check my suggestions. Are you able to reproduce the case where `end_dumper()` would go in the second timeout, where the process didn't terminate even after sending SIGTERM?

core/src/plugins/filed/python/percona-xtrabackup/bareos-fd-percona-xtrabackup.py
banuchka: That is +30s timeout after sending SIGTERM to the subprocess. Used in line 576.

banuchka:
> Please check my suggestions. Are you able to reproduce the case where `end_dumper()` would go in the second timeout, where the process didn't terminate even after sending SIGTERM?

I'm not sure that is 100% reproducible on my side. I saw an empty return code straight after sending SIGTERM which is why I put an additional 30 seconds waiting for the _correct_ termination.

I agree that that is possible not to wait 2nd 30 sec and jump to the *else* block with setting *9* to the stream's return code, but prefer to wait a bit more seconds.

core/src/plugins/filed/python/percona-xtrabackup/bareos-fd-percona-xtrabackup.py
sduehr: > That is +30s timeout after sending SIGTERM to the subprocess. Used in line 576.

Sure, but that's done by `start_time = int(time.time())`, the value of the timeout variable is always 30, as that is assigned before the while loop. Assigning the same value again is unnecessary and can be confusing when reading the code. It would only make sense if you would assign a different value than 30, for example if you want a longer or shorter timeout than 30s to wait after sending SIGTERM.

banuchka: Is that ok if I put the value of 10s there?

sduehr (approved):

sduehr:
OK, thanks. Yes, setting 9 to the stream's return isn't perfect, but ok for now.
I pushed some changes to make the commit messages compliant with our rules.

PR #1753, opened 2024-03-28
Generate LICENSE.txt from LICENSE.template

joergsteffens:
Moved away from generating `debian/copyright` and create `LICENSE.txt` instead.
Fix adding `webui/copyright`, remove unimportant sections, update to 2024.

#### Please check

- [x] Short description and the purpose of this PR is present _above this paragraph_
- [x] Your name is present in the AUTHORS file (optional)

If you have any questions or problems, please give a comment in the PR.

### Helpful documentation and best practices

- [Git Workflow](https://docs.bareos.org/DeveloperGuide/gitworkflow.html)
- [Automatic Sourcecode Formatting](https://docs.bareos.org/DeveloperGuide/generaldevel.html#automatic-sourcecode-formatting)
- [Check your commit messages](https://docs.bareos.org/DeveloperGuide/gitworkflow.html#commits)
- [Boy Scout Rule](https://docs.bareos.org/DeveloperGuide/generaldevel.html#boy-scout-rule)

### Checklist for the _reviewer_ of the PR (will be processed by the Bareos team)
Make sure you check/merge the PR using `devtools/pr-tool` to have some simple automated checks run and a proper changelog record added.

##### General
- [x] Is the PR title usable as CHANGELOG entry?
- [x] Purpose of the PR is understood
- [x] Commit descriptions are understandable and well formatted
- [x] Check backport line
- [x] Required backport PRs have been created

##### Source code quality
- [x] Source code changes are understandable
- [x] Variable and function names are meaningful
- [x] Code comments are correct (logically and spelling)
- [x] Required documentation changes are present and part of the PR

PR #1755, opened 2024-03-28
infra : GitHub community addition

bruno-at-bareos:
this PR aims to improve github community scoring by adding a root license.md file, also a contributing.md guide and a code of conduct.

There's no needs for building, nor for backport. 
We may want to test the commit on bareos-testing before merging.

OP#5754 

#### Please check

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,129 @@
+# Contributor Covenant Code of Conduct
+
+## Our Pledge
+
+We as members, contributors, and leaders pledge to make participation in our
+community a harassment-free experience for everyone, regardless of age, body
+size, visible or invisible disability, ethnicity, sex characteristics, gender
+identity and expression, level of experience, education, socio-economic status,
+nationality, personal appearance, color, religion, or sexual identity
+and orientation.
+
+We pledge to act and interact in ways that contribute to an open, welcoming,
+diverse, inclusive, and healthy community.
+
+## Our Standards
+
+Examples of behavior that contributes to a positive environment for our
+community include:
+
+* Demonstrating empathy and kindness toward other people
+* Being respectful of differing opinions, viewpoints, and experiences
+* Giving and gracefully accepting constructive feedback
+* Accepting responsibility and apologizing to those affected by our mistakes,
+  and learning from the experience
+* Focusing on what is best not just for us as individuals, but for the
+  overall community
+
+Examples of unacceptable behavior include:
+
+* The use of sexualized language or imagery, and sexual attention or
+  advances of any kind
+* Trolling, insulting or derogatory comments, and personal or political attacks
+* Public or private harassment
+* Publishing others' private information, such as a physical or email
+  address, without their explicit permission
+* Other conduct which could reasonably be considered inappropriate in a
+  professional setting
+
+## Enforcement Responsibilities
+
+Community leaders are responsible for clarifying and enforcing our standards of
+acceptable behavior and will take appropriate and fair corrective action in
+response to any behavior that they deem inappropriate, threatening, offensive,
+or harmful.
+
+Community leaders have the right and responsibility to remove, edit, or reject
+comments, commits, code, wiki edits, issues, and other contributions that are
+not aligned to this Code of Conduct, and will communicate reasons for moderation
+decisions when appropriate.
+
+## Scope
+
+This Code of Conduct applies within all community spaces, and also applies when
+an individual is officially representing the community in public spaces.
+Examples of representing our community include using an official e-mail address,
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,5 @@
+# Contribute
+
+If you want to contribute something to this repo please read our contribution policy at
+
+https://www.bareos.org/en/howto-contribute.html
```
arogge, line 5: that does not (yet) link to something reasonable.
Maybe we want to migrate the contributor section from the docs into MD format and place it here. 
We could still include it in the docs with the same mechanism that use for the CHANGELOG.md

LICENSE.md
```diff
@@ -0,0 +1 @@
+debian/copyright
```
arogge, line 1: same as above, we could move debian/copyright to LICENSE.md and then regenerate debian/copyright when building for Debian (CMake infrastructure for that is mostly in place already, so this should be pretty easy to do).

CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,129 @@
+# Contributor Covenant Code of Conduct
+
+## Our Pledge
+
+We as members, contributors, and leaders pledge to make participation in our
+community a harassment-free experience for everyone, regardless of age, body
+size, visible or invisible disability, ethnicity, sex characteristics, gender
+identity and expression, level of experience, education, socio-economic status,
+nationality, personal appearance, race, religion, or sexual identity
```
arogge, line 9: While I understand the idea I would definitely refrain from using _race_ here. The term itself is inappropriate for humans and everything that this could refer to is - in my opinion - already covered (i.e. ethnicity, nationality, personal appearance).
The 2.1 version of the contributor covenant has two interesting additions to the list: _caste_ and _color_. While I believe _caste_ is even worse than _race_, _color_ might express the original intent best.

[Wikipedia](https://en.wikipedia.org/wiki/Race_(human_categorization)) says:
> The concept of race is foundational to [racism](https://en.wikipedia.org/wiki/Racism), the belief that humans can be divided based on the superiority of one race over another.

CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,129 @@
...
+
+We pledge to act and interact in ways that contribute to an open, welcoming,
+diverse, inclusive, and healthy community.
+
+## Our Standards
+
+Examples of behavior that contributes to a positive environment for our
+community include:
+
+* Demonstrating empathy and kindness toward other people
```
arogge, line 20: That's probably a great goal to achieve universal peace. 
However, I'm not sure that we as a community are able and willing to enforce this.

arogge (commented):
I have a few comments on the CoC and suggestions for relocating contributor and license information.

LICENSE.md
bruno-at-bareos: as discussed this morning during our developer meeting we will keep the symlink mechanism for the moment.

CODE_OF_CONDUCT.md
bruno-at-bareos: well it is cited as example of behavior, it may also remember skinned elephant that maybe new comer will appreciate this kindness ;-)

CONTRIBUTING.md
bruno-at-bareos: Ok then we need input from owner how they want to handle this.

LICENSE.md
bruno-at-bareos: > same as above, we could move debian/copyright to LICENSE.md and then regenerate debian/copyright when building for Debian (CMake infrastructure for that is mostly in place already, so this should be pretty easy to do).

Then we should have /LICENSE.md build like debian/copyright is created actually (so adding a tier lib etc will not got forgotten) and then copied around when debian packages are created.

CODE_OF_CONDUCT.md
bruno-at-bareos: Totally agreed here. I've made the changes.

bruno-at-bareos:
About license.txt we may want to build it with #1753

CONTRIBUTING.md
bruno-at-bareos: Decision was made during DevMeeting to keep it like it is actually and web page will soon updated

joergsteffens: Still you should update the link to https://www.bareos.com/community/ ,which is the redirection target of the above link.

bruno-at-bareos: > Still you should update the link to https://www.bareos.com/community/ ,which is the redirection target of the above link.

Thanks I missed that redirection :-)

arogge (approved):
Great work! Sorry for taking that long to review this.

PR #1787, opened 2024-04-18
cats: add missing database locks

sebsura:

By default, bareos db connections get shared by multiple jobs/threads.  It is of utmost importance to ensure that those shared connections only get written to/read from if the calling thread has acquired the lock.

This PR by no means ensures that this is done correctly everywhere, but the start of an ongoing process to weed out every race condition in the database code.

This PR adds some missing locks as well as assertions that locks were taken correctly at the relevant places.
Additionally it renames `bac_cursor_` to `bar_cursor_`.

Some unused database code was also removed as well as making the `numeric` type count as a numeric type.

#### Please check

##### General
- ~Required backport PRs have been created~

##### Source code quality


core/src/cats/postgresql.cc
```diff
@@ -828,6 +844,7 @@ void BareosDbPostgresql::SqlUpdateField(int i)
 
 SQL_FIELD* BareosDbPostgresql::SqlFetchField(void)
 {
+  AssertOwnership();
```
arogge, line 847: do we have any idea how this will influence runtime?
`SqlFetchField()` is called a lot in pretty tight loops.
Its happy-path boils down to:
1. integer comparison
2. nullptr check
3. boolean check
4. add two pointers (for array access)
5. integer increment

Calling `AssertOwnership()` while assuming we're on Linux, it is completly inlined, on the happy path, and having `is_private_ == false`:
1. boolean check
2. pointer dereference
3. integer equality comparison
4. integer comparison
5. library call to `pthread_self()` (cannot be inlined)
7. long integer equality comparison (always inlined from `pthread_equal()`)

I don't say this will be a performance issue. But it definitely looks  like could be. Especially with the redundant check in `SqlUpdateField()` above we would call `AssertOwnership()` twice for each field.

core/src/cats/postgresql.cc
```diff
@@ -800,6 +815,7 @@ uint64_t BareosDbPostgresql::SqlInsertAutokeyRecord(const char* query,
 
 void BareosDbPostgresql::SqlUpdateField(int i)
 {
+  AssertOwnership();
```
arogge, line 818: `SqlUpdateField()` is only called by `SqlFetchField()`, so this check is redundant.

arogge (changes requested):
I'm not happy with the amount of calls to `AssertOwnership()`, especially as it looks pretty expensive. However, I don't have a great idea how to improve it right now.
I'll see if I can come up with an idea over the weekend...

core/src/cats/postgresql.cc
sebsura: What do you think of this:  We leave the calls where they are, but use an empty definition during release builds so that we only do the checking on debug builds ?

sebsura:
I had to rebase onto master to make it build again. Sorry!

core/src/cats/cats.cc
```diff
@@ -3,7 +3,7 @@
 
    Copyright (C) 2011-2011 Free Software Foundation Europe e.V.
    Copyright (C) 2011-2016 Planets Communications B.V.
-   Copyright (C) 2013-2023 Bareos GmbH & Co. KG
+   Copyright (C) 2013-2024 Bareos GmbH & Co. KG
```
arogge, line 6: If we don't have any change, this should also go away (after merging the fixups).

arogge (approved):
Besides the year in cats.cc's header, I think this is fine now.

core/src/cats/cats.cc
sebsura: Merged the fixups now.  Lets see what jenkins has to say.

PR #1835, opened 2024-05-28
fix #1775 plugin: fd mariabackup add support mariadb 11+

### LINKED ISSUE #1775: plugin:mariabackup will failed with mariadb 11+
### Bareos component version

1. bareos > 22.0
2. mariadb >= 11.0 

### Steps to reproduce

1. install bareos version 23
2. install mariadb version 11+
3. setup backup with mariabackup plugin
4. see error raised during backups

### Expected results

Terminated jobs

### Actual results

The message below appears when you try to make an incremental mariadb backup. I think the backup work fails because of the warning.

```
Fatal error: python3-fd-mod: Could not get LSN with command "echo 'SHOW ENGINE INNODB STATUS' | mysql --defaults-extra-file=/root/.my.cnf -r | grep 'Log sequence number' | awk '{ print $4 }'", Error: b"mysql: Deprecated program name. It will be removed in a future release, use '/usr/bin/mariadb' instead\n"
```

### Environment

```markdown
- OS: `Linux Ubuntu 22.04`
- component: mariadb version: 11.0.3-MariaDB-1:11.0.3+maria~ubu2204 mariadb.org binary distribution
```

### Relevant log output

_No response_

### Relevant traces output

_No response_

### Anything else?

Mariadb is getting ride of old names, so the long deprecated mysql* and also the symlink mariabackup should be replaced by the real binary name
mariabackup -> mariadb-backup
mysql -> mariadb 
....

Initial report https://bugs.bareos.org/view.php?id=1568

An experimental branch exist with some changes may serve as a basis
https://github.com/bruno-at-bareos/bareos/tree/dev/bruno/master/fix-mariadb-11%2B

bruno-at-bareos:
### plugin: fd mariabackup add support for MariaDB 11+

- **tests: remove generated config file**
- **systemtests: scripts mysql.sh replace inexistant call**
- **cmake: fix copyright year**
- **plugin: fd mariabackup add support for mariadb 11+**
- fix issue #1775

*note the code may still be improved by applying pylint recommandation and fixing todo*

#### Please check

##### General

##### Source code quality

##### Tests


sduehr (changes requested):
Only a trivial change requested and a question about a comment.
Is the updated plugin compatible with the old one, so restore of backups taken with the old version are still restorable or is a new full required after update? If yes, then we should document that.
Should we document that the changed plugin needs MariaDB >= 10.5? Should still work with 10.3 (it's EOL anyway) but would need to set some options, as it doesn't yet have the commands mariadb-backup and mariadb.

core/src/plugins/filed/python/mariabackup/bareos-fd-mariabackup.py
```diff
@@ -211,6 +293,8 @@ def start_backup_job(self):
                     100,
                     "Import of module MySQLdb failed. Using command pipe instead\n",
                 )
+            # TODO Fix as this doesn't use the mysqldefault passed file and always try to connect
```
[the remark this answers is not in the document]
bruno-at-bareos: This is still the case: it is an old problem since long time, the python mysql code path will try to connect to the default socket instead of the one described in .mysqldefault file.
This is not usually detected by end-user because the FD run as root on the DB host and usually this one is configured to backup a main default mysql/mariadb host.
It become visible if you try to run the systemtest as build user and run a "user" instance mysql.
It hasn't been detected before by our systemtest due to two things:
- we had a default mysql instance running before when we used several db and the code was trying to check the incremental from that instance.
- we often don't have the python mysql module installed, and then the systemtest used the external command

bruno-at-bareos:
> Is the updated plugin compatible with the old one, so restore of backups taken with the old version are still restorable or is a new full required after update? If yes, then we should document that. Should we document that the changed plugin needs MariaDB >= 10.5? 
Should still work with 10.3 (it's EOL anyway) but would need to set some options, as it doesn't yet have the commands mariadb-backup and mariadb.

For the case of Database engine upgrade from 10.3 to 10.5, it was clear to me that user need to run a new full just after the upgrade.  And previous backups are dooms as the would imply new db upgrade process.

I was not expecting to be compatible with eol mariadb 10.3 ( this plugin will be in use with bareos 24) but this can be discussed of course.

bruno-at-bareos:
Manually tested with `mariadb from 11.4.2-MariaDB, client 15.2 for debian-linux-gnu (x86_64) using  EditLine wrapper` work well.

bruno-at-bareos:
@sduehr  kindly ping : could finish the review for a definitive merge?

core/src/plugins/filed/python/mariabackup/bareos-fd-mariabackup.py
```diff
@@ -137,10 +145,81 @@ def check_plugin_options(self, mandatory_options=None):
...
+        )
+        DebugMessage(100, f'get_lsn_by_command: "{get_lsn_cmd}"\n')
+        try:
+            ret = check_output(get_lsn_cmd).decode()
+            return ret
+        except CalledProcessError as run_err:
+            DebugMessage(
+                100,
+                f"Error while looking for LSN: {run_err}\n",
+            )
```
sduehr, line 201:
```suggestion
            JobMessage(
                M_FATAL,
                f"Error while looking for LSN: {run_err}\n",
            )
```
I may be wrong, but I think the job should fail if this happens.

core/src/plugins/filed/python/mariabackup/bareos-fd-mariabackup.py
```diff
@@ -137,10 +145,81 @@ def check_plugin_options(self, mandatory_options=None):
...
+        last_lsn = None
+        try:
+            for line in res.split("\n"):
+                if line.startswith("Log sequence number"):
+                    last_lsn = int(line.split(" ")[-1])
+        except ValueError as val_err:
+            DebugMessage(
+                100,
+                f"Returned LSN seems wrong: {val_err}\n",
+            )
```
sduehr, line 214:
```suggestion
            JobMessage(
                M_FATAL,
                f"Returned LSN seems wrong: {val_err}\n",
            )
```
Same here.

core/src/plugins/filed/python/mariabackup/bareos-fd-mariabackup.py
```diff
@@ -137,10 +145,81 @@ def check_plugin_options(self, mandatory_options=None):
...
+            return bRC_Error
+        return last_lsn
```
sduehr, line 216:
```suggestion
        if last_lsn is None:
            JobMessage(
                M_FATAL,
                f"Could not get LSN from {res}\n",
            )
            return bRC_Error
        return last_lsn
```
I think this should be handled as error.

core/src/plugins/filed/python/mariabackup/bareos-fd-mariabackup.py
```diff
@@ -237,32 +322,9 @@ def start_backup_job(self):
...
-                        last_lsn = int(mysqlStdOut)
-                    except:
-                        JobMessage(
-                            M_FATAL,
-                            'Error reading LSN: "%s" not an integer' % mysqlStdOut,
-                        )
-                        return bRC_Error
+                innodb_status = self.get_lsn_by_command()
+
+            last_lsn = self.parse_innodb_status(innodb_status)
```
sduehr, line 327:
```suggestion
            last_lsn = self.parse_innodb_status(innodb_status)
            if last_lsn == bRC_Error:
                return bRC_Error
```
If an error occured, it should be returned here.

sduehr (changes requested):
When getting the LSN, error handling should be improved as I suggested.

core/cmake/BareosFindPrograms.cmake
```diff
@@ -1,6 +1,6 @@
 #   BAREOS® - Backup Archiving REcovery Open Sourced
 #
-#   Copyright (C) 2017-2021 Bareos GmbH & Co. KG
+#   Copyright (C) 2017-2023 Bareos GmbH & Co. KG
```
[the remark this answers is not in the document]
bruno-at-bareos: Beware this commit was done last year, so bareos-check-source will fix it back to 2023.
I will maybe edit it with a change this year, so the date will align with the rest of the Pr.

sduehr (approved):
Thanks, looks good now.

PR #1872, opened 2024-06-28
fix runtime status

arogge:
On a reload, resources (Job, Client and Storage) will get new runtime status objects. As those are used to track concurrent jobs and related things, this leads to all kinds of strange behaviours.
This PR registers the status objects by resource name, so on a reload a resource with the same name will get the same status object.

#### Please check

##### General

##### Source code quality

##### Tests


pstorz (approved):
Looks good!

core/src/dird/dird_conf.cc
```diff
@@ -914,6 +917,22 @@ bool PrintConfigSchemaJson(PoolMem& buffer)
...
+{
+  static std::unordered_map<std::string, std::weak_ptr<T>> map{};
+
+  std::shared_ptr<T> sptr = map[name].lock();
+  if (!sptr) {
+    sptr = std::make_shared<T>();
+    map[name] = sptr;
+  }
+  Dmsg0(50, "Returning RuntimeStatus with use_count=%ld\n", sptr.use_count());
+  return sptr;
```
sebsura, line 932: Maybe we want to actually save them in a vector, and every time this is asked, we run through the whole vector removing any `expired()` weak_ptrs and then return the match (or create a new one if no match was found).

Otherwise the memory for T & the reference counter will never be released for old configs (which to be fair should not be a big deal).

sebsura: > The memory for T will be released when the last shared_ptr<T> object is destroyed. The reason I implemented it like this is that I need a quick solution that works.

This is true for normal shared pointers, but not those created with `make_shared`, as the point of that function is to only use one allocation for both the reference count as well as the storage for T instead of having two allocations (one each for the reference count and the storage for T).
See here: https://godbolt.org/z/q9WGPY4fo

core/src/dird/ndmp_dma_storage.cc
```diff
@@ -997,19 +997,17 @@ bool NdmpSendLabelRequest(UaContext* ua,
...
-  unlock_mutex(store->runtime_storage_status->changer_lock);
 
-  lock_mutex(store->runtime_storage_status->changer_lock);
-  if (!NdmpUpdateStorageMappings(jcr, store)) { retval = false; }
-  unlock_mutex(store->runtime_storage_status->changer_lock);
+  const std::lock_guard lock_(store->runtime_storage_status->changer_lock);
+  if (!NdmpUpdateStorageMappings(jcr, store)) { return false; }
 
-  return retval;
+  return true;
```
sebsura, line 1010: Hmm.... are we sure that we want to change the behaviour here ?  Previously `NdmpUpdateStorageMappings` was called regardless of whether the first function succeeded or not.

core/src/dird/dird_conf.cc
arogge: I just read up on what the C++ standard defines for `std::make_shared`. That's again a great example where the standard complicates things for optimization and provides surprising semantics: different allocation/deallocation is just weird, but "as if calling `::new`" instead of "calling `new`" is evil. They could have just disabled `make_shared<T>` if `T` implements `operator new()`.

After finishing my rant: do you think it is worth spending the time to re-implement this properly with locking and everything?
If you're really concerned with the memory for `T` we could simply use
```
  sptr = new T();
```
instead of
```
  sptr = std::make_shared<T>();
```
However, `sizeof(T)` is 4 bytes for `RuntimeClientStatus` and `RuntimeJobStatus`. So this will save 4 bytes for the common objects and ~192 bytes for the much less common `RuntimeStorageStatus`.

sebsura: As i said above: its not a big deal, its just something we should be aware of.  Regarding locks: either this function can be called from multiple different threads concurrently and we need locks or it isnt and we dont, regardless of whether we remove elements or not during the call (as far as i can see we do not need them currently).

From my point of view we can merge this since, as you said, the memory waste is very minor.

core/src/dird/ndmp_dma_storage.cc
arogge: Definitely not. Good catch, I changed it back.

systemtests/tests/bareos-concurrency/testrunner-client-concurrency
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+set -e
+set -o pipefail
+set -u
+#
+# Run a simple backup
+#   then restore it.
+#
```
sebsura, line 8: Could you add a description to the tests that matches what they are doing ?  Its not clear to me what exactly  we are trying to test here.

sebsura (changes requested):

sebsura (approved):
Great work!

PR #1875, opened 2024-07-01
Fix multiple ACL handling bugs

joergsteffens:
ACL handling for commands have been wrong.
The UA detect a command, if all typed chars matches the start of a
command string. E.g. "w" => "whoami", ".cons" => ".consoles", ...
The ACL check also only used the typed part of the command,
so if your ACL denies the command ".consoles",
you could execute it with ".cons" (or ".co").
Now we check with the full command string against the ACL.

Fix handling ACL profiles

When using negative ACLs (ACLs starting with '!')
profiles could have overwritten this entries,
as the code did not distringuish between negative ACL and not found.

Now the ACL checking can have three states:
positive match found,
negative match found,
no match found.

Also unifies the ACL checking
and adding the ACL directive `!*all*`
to deny all (before a regex must be used (`!.*`).

#### Please check

##### General

##### Source code quality

##### Tests


core/src/dird/ua_acl.cc
```diff
@@ -47,132 +47,101 @@ bool UaContext::AclAccessOk(int acl, const char* item, bool audit_event)
...
+            pmatch[0].rm_so, pmatch[0].rm_eo);
+      if ((pmatch[0].rm_eo - pmatch[0].rm_so) >= match_length) {
+        Dmsg3(1400, "ACL found %s in %d using regex %s\n", item, acl,
+              acl_list_value);
+        regfree(&preg);
+        return true;
+      }
+    }
+    regfree(&preg);
+  }
```
sebsura, line 105: If we allow regexes anyways, why do we even have `*all*` ? Wouldnt `.*` have sufficed ?

I couldnt actually produce an error, but we should probably enumerate all matches and not just take the first one.  I have not found any guarantee that the returned matches are ordered by length, so technically we could have a case where a partial match is returned first even though a full match exists.
But im not sure if this is actually worth the effort.

sebsura: Implicitly doing a regex check is also a bit weird, as e.g. allowing `.list` also always means that we allow `llist`, but this should be fine in most cases.

core/src/dird/ua_acl.cc
```diff
@@ -205,25 +174,27 @@ bool UaContext::AclAccessOk(int acl,
...
-    foreach_alist (profile, user_acl->profiles) {
-      retval = FindInAclList(profile->ACL_lists[acl], acl, item, len);
+      foreach_alist (profile, user_acl->profiles) {
+        retval = FindInAclList(profile->ACL_lists[acl], acl, item, len);
 
-      // If we found a match break the loop.
-      if (retval) { break; }
+        // If we found a match break the loop.
+        if (retval.has_value()) { break; }
+      }
```
sebsura, line 188: This looks a bit nondeterministic.  E.g. if im connected to two profiles and one of them disallows `restore` while the other one allows it, then its up to chance if i get to restore or not.

Do you think we should fix this ? For example we could only allow a command if either *any* profile allows it or if *all* profiles allow it.
I would prefer the *any* case, but since we know use three boolean states, *all* would also make some sense.

core/src/dird/ua_cmds.cc
```diff
@@ -587,16 +587,16 @@ bool Do_a_command(UaContext* ua)
   len = strlen(ua->argk[0]);
   for (i = 0; i < comsize; i++) { /* search for command */
     if (bstrncasecmp(ua->argk[0], commands[i].key, len)) {
+      const char* command = commands[i].key;
       // Check if command permitted, but "quit" and "whoami" is always OK
-      if (!bstrcmp(ua->argk[0], NT_("quit"))
-          && !bstrcmp(ua->argk[0], NT_("whoami"))
-          && !ua->AclAccessOk(Command_ACL, ua->argk[0], true)) {
+      if (!bstrcmp(command, NT_("quit")) && !bstrcmp(command, NT_("whoami"))
+          && !ua->AclAccessOk(Command_ACL, command, true)) {
```
sebsura, line 593: What about the "exit" command ? We should probably also allow that one.

Im not sure why the quit/exit commands are even sent to the director in the first place.  They should be handled locally by the console but i digress.

systemtests/tests/python-bareos/test_acl.py
```diff
@@ -402,3 +398,30 @@ def test_status_subscription_admin(self):
...
+        )
+
+        with self.assertRaises(bareos.exceptions.JsonRpcErrorReceivedException):
+            result = console.call(".consoles")
+            # logger.debug(result)
+            # consoles = [item["name"] for item in result["consoles"]]
+            # self.assertIn(username, consoles)
+
+        with self.assertRaises(bareos.exceptions.JsonRpcErrorReceivedException):
+            result = console.call(".consol")
```
sebsura, line 427: i feel like we are missing something here.  What is this trying to test ?  I would assume that we would test that these calls return nothing as the acl disallows these commands but im not seeing this here.

We should probably also test the opposite: Have a positive acl and check that it actually allows us to use the commands.

core/src/dird/ua_acl.cc
```diff
@@ -47,132 +47,101 @@ bool UaContext::AclAccessOk(int acl, const char* item, bool audit_event)
  * A regexp uses the following chars:
  * ., (, ), [, ], |, ^, $, +, ?, *
  */
-static bool is_regex(std::string string_to_check)
+static inline bool is_regex(std::string string_to_check)
 {
   return std::string::npos != string_to_check.find_first_of(".()[]|^$+?*");
```
sebsura, line 52:
```suggestion
  return string_to_check.npos != string_to_check.find_first_of(".()[]|^$+?*");
```
This makes changing the type of `string_to_check` easier.

sebsura (changes requested):
Nice changes!  I found some issues that i would like to see fixed.

systemtests/tests/python-bareos/test_acl.py
joergsteffens: DirectorConsoleJson will raise an exception, if it detects an error. I added a comment for clarification to the code.

core/src/dird/ua_cmds.cc
joergsteffens: After an internal discussion, we decided to leave it as it is for the moment, at least don't change it with this PR:

core/src/dird/ua_acl.cc
```diff
@@ -47,132 +47,101 @@ bool UaContext::AclAccessOk(int acl, const char* item, bool audit_event)
...
-    /*
-     * Empty list for Where => empty where accept anything.
-     * For any other list, reject everything.
-     */
+    /* Empty list for Where => empty where accept anything.
+     * For any other list, reject everything. */
     if (len == 0 && acl == Where_ACL) {
       Dmsg0(1400, "Empty Where_ACL allowing restore anywhere\n");
-      retval = true;
+      return true;
```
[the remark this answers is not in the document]
joergsteffens: You're absolutely right.
Currently an empty WhereACL have not been handled, as the code for this has been wrong.
So I decided to remove this special treatment at all, as, even if implemented correctly would cause problems if used together with profiles (as emptiness would be check for each resource separately).

joergsteffens:
The test `config-syntax-crash` will now fail, as the Director will now produce a proper config error on invalid ACLs, instead of just ignoring them as before. The test have been introduced because of https://bugs.bareos.org/view.php?id=1175, leading to #410. @arogge do you remember, if there is anything more to this test? Otherwise I replaced the test with a small extra python-bareos test.

core/src/dird/ua_cmds.cc
sebsura: @joergsteffens but i think we wanted to add exit here as well.

core/src/dird/ua_run.cc
```diff
@@ -723,18 +723,24 @@ int ModifyJobParameters(UaContext* ua, JobControlRecord* jcr, RunContext& rc)
...
-              free(jcr->RegexWhere);
-              jcr->RegexWhere = NULL;
-            }
-            if (jcr->where) {
-              free(jcr->where);
-              jcr->where = NULL;
-            }
-            if (IsPathSeparator(ua->cmd[0]) && ua->cmd[1] == '\0') {
-              ua->cmd[0] = 0;
+            if (!ua->AclAccessOk(Where_ACL, ua->cmd, true)) {
```
sebsura, line 726: This should probably also happen in case 10 (File Relocation) where regexwhere is set instead.

sebsura (changes requested):
I have not checked everything yet but these are things i noticed

core/src/dird/ua_run.cc
joergsteffens: sure.

core/src/dird/ua_cmds.cc
joergsteffens: okay

sebsura (approved):
Thanks for your work!

PR #1796, opened 2024-04-29
bareos-fd-vmware: Fix Restore as a new VM 

robisetiapermadi:
I have modify the vmware bareos plugin source code, so when restoring  as a new VM, the VMware plugin will ignore if the backed up disk is not the same as the new VM. 

The check_vm_disks_match() function will check whether the recovery job is restored as a new vm or not.  If restored as a new VM, check_vm_disks_match() will return true even if the backed up disk is not the same as the new VM.

This PR should fix bug when VMware Plugin Restore as a new VM that i have mention on this report https://groups.google.com/g/bareos-users/c/RgHCwTfmQ5o/m/d-Q_DpdhAQAJ?utm_medium=email&utm_source=footer

#### Please check

- [ ] Short description and the purpose of this PR is present _above this paragraph_
- [ ] Your name is present in the AUTHORS file (optional)

If you have any questions or problems, please give a comment in the PR.

### Helpful documentation and best practices

- [Git Workflow](https://docs.bareos.org/DeveloperGuide/gitworkflow.html)
- [Automatic Sourcecode Formatting](https://docs.bareos.org/DeveloperGuide/generaldevel.html#automatic-sourcecode-formatting)
- [Check your commit messages](https://docs.bareos.org/DeveloperGuide/gitworkflow.html#commits)
- [Boy Scout Rule](https://docs.bareos.org/DeveloperGuide/generaldevel.html#boy-scout-rule)

### Checklist for the _reviewer_ of the PR (will be processed by the Bareos team)
Make sure you check/merge the PR using `devtools/pr-tool` to have some simple automated checks run and a proper changelog record added.

##### General
- [ ] Is the PR title usable as CHANGELOG entry?
- [ ] Purpose of the PR is understood
- [ ] Commit descriptions are understandable and well formatted
- [ ] Required backport PRs have been created

##### Source code quality
- [ ] Source code changes are understandable
- [ ] Variable and function names are meaningful
- [ ] Code comments are correct (logically and spelling)
- [ ] Required documentation changes are present and part of the PR

##### Tests
- [ ] Decision taken that a test is required (if not, then remove this paragraph)
- [ ] The choice of the type of test (unit test or systemtest) is reasonable
- [ ] Testname matches exactly what is being tested
- [ ] On a fail, output of the test leads quickly to the origin of the fault

PR #1876, opened 2024-07-01
VMware Plugin: Add option restore_allow_disks_mismatch

sduehr:
Adds the option `restore_ignore_disks_mismatch` to the VMware Plugin.
This can be helpful when using VSAN and the restore fails due to disk check mismatch
when the VM is recreated by the plugin.

##### General

##### Source code quality

### DIFF
--- a/core/src/plugins/filed/python/vmware/bareos-fd-vmware.py
+++ b/core/src/plugins/filed/python/vmware/bareos-fd-vmware.py
@@ -124,7 +124,7 @@ class BareosFdPluginVMware(BareosFdPluginBaseclass.BareosFdPluginBaseclass):
             "vadp_dumper_sectors_per_call",
             "vadp_dumper_query_allocated_blocks_chunk_size",
             "fallback_to_full_cbt",
-            "restore_ignore_disks_mismatch",
+            "restore_allow_disks_mismatch",
         ]
         self.allowed_options = (
             self.mandatory_options_default
@@ -393,8 +393,8 @@ class BareosFdPluginVMware(BareosFdPluginBaseclass.BareosFdPluginBaseclass):
         if self.options.get("fallback_to_full_cbt") == "no":
             self.vadp.fallback_to_full_cbt = False
 
-        if self.options.get("restore_ignore_disks_mismatch") == "yes":
-            self.vadp.restore_ignore_disks_mismatch = True
+        if self.options.get("restore_allow_disks_mismatch") == "yes":
+            self.vadp.restore_allow_disks_mismatch = True
 
         for option in self.options:
             bareosfd.DebugMessage(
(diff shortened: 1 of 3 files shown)


arogge:
This is an alternative to PR #1796

core/src/plugins/filed/python/vmware/bareos-fd-vmware.py
```diff
@@ -3162,10 +3176,21 @@ def add_disk_devices_to_vm(self, device_changes):
...
+                )
                 try:
                     pyVim.task.WaitForTask(reconfig_task)
                     device_created = True
                 except vim.fault.FileAlreadyExists:
+                    bareosfd.DebugMessage(
+                        100,
+                        "add_disk_devices_to_vm(): Disk %s already exists, retrying with datastore name only\n"
+                        % (device_spec.device.backing.fileName),
+                    )
```
arogge, line 3192: maybe this should be a JobMessage instead?

core/src/plugins/filed/python/vmware/bareos-fd-vmware.py
```diff
@@ -124,6 +124,7 @@ def __init__(self, plugindef):
             "vadp_dumper_sectors_per_call",
             "vadp_dumper_query_allocated_blocks_chunk_size",
             "fallback_to_full_cbt",
+            "restore_ignore_disks_mismatch",
```
arogge, line 127: As there is still a M_WARNING when this happens, I'd argue we do not ignore it, but just allow it.

```suggestion
            "restore_allow_disks_mismatch",
```

arogge (changes requested):
The code itself looks pretty good. I'd rename the option and make one of the debug messages a job message.

core/src/plugins/filed/python/vmware/bareos-fd-vmware.py
arogge: So by default we currently try to recreate the disks (that's basically VMDK-Files on the VMFS store, right) with the name they had when they were backed up. As you said, we could just always recreate the disks with new names.
With the current implementation, the user may expect to get the same names, because that is what usually happens. Thus I argued there could be a job message (M_INFO is totally sufficient) when we don't meet that expectation.
But feel free to ignore that, if it doesn't make sense - it is just a suggestion :)

sduehr:
In the new commit add_disk_devices_to_vm() changed so that it only passes the datastore name as backing path. Like this it should always succeed and register the new disk file name. This is only called when recreating the VM, and if the disk path/filename differs from the backed up name, it will generate a job message. I think like this it should also work with VSAN, without the need to allow mismatch of the disk check.

arogge (approved):
While I sadly cannot test this, when reading the code it looks good to me!

PR #1898, opened 2024-07-19
build: enable building for Fedora40 and Ubuntu 24.04

arogge:
**Backport of PR #1815 to bareos-23**

The following changes were done to the original PR:
* remove build of Debian 10, as it is now EOL
* limit BuildRequires: fmt-devel on SUSE to version < 1500

#### Backport quality
- [x] Original PR #1815 is merged


sebsura:
Why is the check still valid for bareos 23 ? It does not look different to me. If
* `calling thread == tcp server thread`, then calling `pthread_join()` is UB and will (probably) hang, 
* otherwise `BnetStopAndWaitForThreadServerTcp()` will already call `pthread_join(tcp_server_tid)`, so the second join is not necessary and even ub:
  > Joining with a thread that has previously been joined results in undefined behavior.
  
So to me it looks like that check is still nonsensical and should be removed.  I guess we could technically still call
```
delete sock_fds;
```
but this is unnecessary as well since this function is only called on shutdown anyways.

pstorz (approved):

sebsura (dismissed):
Just a reminder that my comment was not yet addressed.

Issue #1851, opened 2024-06-07
webui becomes unresponsive when trying to restore when a deleted client exists


sebsura:
### Bareos component version

24.0.0 (master)

### Steps to reproduce

1. Run a backup job on client 1
2. change name of client 1 to client 2 in the configuration
3. reload the configuration
4. in webui restore select client 2

### Expected results

I can select both client 1 and client 2 as source client, and selecting them does what one expects (i.e. you can select  jobs)

### Actual results

You can only select client 2 and doing so causes the webui to become unresponsive.
![Screenshots_59](https://github.com/bareos/bareos/assets/124262655/1672f961-f302-4e9a-a3f3-2d26fd183b8b)
[trace.zip](https://github.com/user-attachments/files/15738411/trace.zip)

### Relevant log output

```shell
Firefox console:

NS_ERROR_NOT_IMPLEMENTED: Component returned failure code: 0x80004001 (NS_ERROR_NOT_IMPLEMENTED) [nsIAppStartup.secondsSinceLastOSRestart]
    _collectStartupConditionsTelemetry resource:///modules/BrowserGlue.sys.mjs:1772
    BG__onFirstWindowLoaded resource:///modules/BrowserGlue.sys.mjs:1880
    BG_observe resource:///modules/BrowserGlue.sys.mjs:1102
    _delayedStartup chrome://browser/content/browser.js:2116
BrowserGlue.sys.mjs:1781:15
    _collectStartupConditionsTelemetry resource:///modules/BrowserGlue.sys.mjs:1781
    BG__onFirstWindowLoaded resource:///modules/BrowserGlue.sys.mjs:1880
    BG_observe resource:///modules/BrowserGlue.sys.mjs:1102
    _delayedStartup chrome://browser/content/browser.js:2116
TypeError: linkMap is undefined
    Links_onLinkChanged resource://gre/modules/NewTabUtils.sys.mjs:2003
    PlacesProvider__callObservers resource://gre/modules/NewTabUtils.sys.mjs:697
    handlePlacesEvents resource://gre/modules/NewTabUtils.sys.mjs:679
NewTabUtils.sys.mjs:699:19
    PlacesProvider__callObservers resource://gre/modules/NewTabUtils.sys.mjs:699
    handlePlacesEvents resource://gre/modules/NewTabUtils.sys.mjs:679
update.locale  file doesn't exist in either the application or GRE directories UpdateUtils.sys.mjs:143:13
[...]

PR #1933, opened 2024-08-14
webui: fix handling deleted clients in restore browser

joergsteffens:

See #1851

This change allows to restore files from clients no longer part of Bareos to other clients.
This did require changes to core as well as webui.
Also some systemtests has been added to test the parameter handling of the bconsole commands.

#### Please check

##### General

##### Source code quality

##### Tests


core/src/dird/ua_output.cc
```diff
@@ -619,146 +674,89 @@ static bool DoListCmd(UaContext* ua, const char* cmd, e_list_type llist)
...
+      || Bstrcasecmp(ua->argk[1], NT_("jobname"))
+      || Bstrcasecmp(ua->argk[1], NT_("jobid"))
+      || Bstrcasecmp(ua->argk[1], NT_("ujobid"))) {
+    return ListJobs(ua, llist, optionslist);
+  }
 
-    SetQueryRange(query_range, ua, &jr);
+  JobDbRecord jr;
+  std::string query_range;
+  SetQueryRange(query_range, ua, &jr);
```
sebsura, line 755: I find it weird that we call this here but then almost every other if branch also calls it an additional time.

But looking at the implementation it looks like this function is intended to be called multiple times so maybe this is intended ?

core/src/dird/ua_output.cc
```diff
@@ -619,146 +674,89 @@ static bool DoListCmd(UaContext* ua, const char* cmd, e_list_type llist)
...
-        if (optionslist.disabled) { SetDisabledFilter(ua, 1, R_JOB); }
-        break;
-    }
+  if (Bstrcasecmp(ua->argk[1], NT_("jobs"))
+      || Bstrcasecmp(ua->argk[1], NT_("job"))
+      || Bstrcasecmp(ua->argk[1], NT_("jobname"))
+      || Bstrcasecmp(ua->argk[1], NT_("jobid"))
+      || Bstrcasecmp(ua->argk[1], NT_("ujobid"))) {
+    return ListJobs(ua, llist, optionslist);
+  }
```
sebsura, line 751: Im not sure why we moved this from the if-else chain down below.

sebsura (changes requested):
There are some very good changes in here.  I only looked at the changes in core for now but i already have some comments.

core/src/dird/ua_output.cc
```diff
@@ -593,21 +616,53 @@ static bool ListMedia(UaContext* ua,
...
+    ua->ErrorMsg(T_("'list jobs' does not allow an jobs parameter\n"));
+    return false;
+  }
 
-  if (ua->argc <= 1) {
-    ua->ErrorMsg(T_("%s command requires a keyword\n"), NPRT(ua->argk[0]));
+  if (const char* value; (value = GetArgValue(ua, NT_("jobname")))
+                         || (value = GetArgValue(ua, NT_("job")))) {
+    jr.JobId = 0;
+    bstrncpy(jr.Name, value, MAX_NAME_LENGTH);
```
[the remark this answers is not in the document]
joergsteffens: Why? The name is invalid, so there also no jobs with that name, and the list command will also not return any result.

core/src/dird/ua_output.cc
joergsteffens: I tried to make the code better readable by this. I now applied the early return also to the rest of the function.

core/src/dird/ua_output.cc
joergsteffens: Not sure, why it is called that often. That should not hurt, but should normally also not help. I reduced the calls now.

core/src/dird/ua_select.cc
```diff
@@ -147,6 +147,36 @@ const char* GetArgValue(UaContext* ua, const char* keyword)
...
+{
+  ClientDbRecord cr;
+  int i = FindArgWithValue(ua, NT_("client"));
+  assert(i != 0);
+  if (i >= 0) {
+    if (!IsNameValid(ua->argv[i], ua->errmsg)) {
+      ua->ErrorMsg("%s argument: %s", ua->argk[i], ua->errmsg.c_str());
+      return -1;
+    }
+    bstrncpy(cr.Name, ua->argv[i], sizeof(cr.Name));
```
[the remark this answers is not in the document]
joergsteffens: As you already pointed out, IsNameValid does check the argument. If it complains, an error message will be written back for the user. What additional ASSERT would you like to see here?

core/src/dird/ua_select.cc
```diff
@@ -1988,6 +2018,12 @@ bool GetUserJobLevelSelection(UaContext* ua, std::vector<char>& joblevel_list)
     for (const auto& level : joblevelinput_list) {
       if (level.size() == 1 && level[0] >= 'A' && level[0] <= 'z') {
```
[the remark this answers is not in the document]
joergsteffens: That would indeed be better, however, I don't see an easy way to archive this. Anyhow, the worst thing that can happens here is that the user specifies an character, not related to a joblevel and therefore the resulting SQL Statement will not return matching jobs.

webui/vendor/Bareos/library/Bareos/BSock/BareosBSock.php
```diff
@@ -897,7 +897,7 @@ private function cram_md5_challenge($password)
      * @param $api
      * @return string
      */
-    public function send_command($cmd, $api=0)
+    public function send_command($cmd, $api = 0)
     {
         $result = "";
         $debug = "";
```
sebsura, line 903: Could the default case no be
```
self::send(".api " . $api);
$debug = self::receive_message();
```
so that we do not need `case 1` or is there an edge case that im missing ?

sebsura: Do we still want to keep code that supports bareos `15.2.2` ? Otherwise i would delete the `preg_match`.

Also why do we only catch exceptions when `api = 2`  ?

webui/vendor/Bareos/library/Bareos/BSock/BareosBSock.php
```diff
@@ -940,6 +940,73 @@ public function send_command($cmd, $api=0)
...
+        //         return
+        //     if "result" in json:
+        //         result = json["result"]
+        //     elif "error" in json:
+        //         raise bareos.exceptions.JsonRpcErrorReceivedException(json)
+        //     else:
+        //         raise bareos.exceptions.JsonRpcInvalidJsonReceivedException(json)
+        //     return result
+        $resultstring = self::send_command($cmd, 2);
+        $json = \Zend\Json\Json::decode($resultstring, \Zend\Json\Json::TYPE_ARRAY);
```
sebsura, line 998: Why are we not reusing the function we just defined ? Python does the same!

```suggestion
        $json = self::call_fullresult_json($cmd);
```

webui/vendor/Bareos/library/Bareos/BSock/BareosBSock.php
```diff
@@ -940,6 +940,73 @@ public function send_command($cmd, $api=0)
...
+        if (empty($json)) {
+            return null;
+        } elseif (array_key_exists("result", $json)) {
+            return $json["result"];
+        } elseif (array_key_exists("error", $json)) {
+            throw new \Exception("JsonRpcErrorReceived: " . $resultstring);
+        } else {
+            throw new \Exception("JsonRpcInvalidJsonReceived: " . $resultstring);
```
sebsura, line 1006: The exception name is a bit weird.  If we receive invalid json, we are actually in case 1 `empty($json)`, as `$json === NULL`.

I guess this happens when we get unexpected (but valid) json.  As python uses the same nomenclature i would not change it though.

webui/module/Client/src/Client/Model/ClientModel.php
```diff
@@ -73,16 +86,15 @@ public function getDotClients(&$bsock = null)
...
-    public function getClient(&$bsock = null, $client = null)
+    public function getClient(&$bsock = null, $client)
     {
-        if (isset($bsock, $client)) {
-            $cmd = 'llist client="' . $client . '"';
-            $result = $bsock->send_command($cmd, 2);
-            $client = \Zend\Json\Json::decode($result, \Zend\Json\Json::TYPE_ARRAY);
-            return $client['result']['clients'];
-        } else {
+        if (!isset($bsock)) {
```
sebsura, line 91: I dont quite get this change.  I feel like calling this function with `$client == null` should still throw an exception as the command is wrong.

systemtests/tests/restore/functions
```diff
@@ -5,6 +5,7 @@
 runner_name="$(basename "$0")"
 log_home="${tmp}/${runner_name}"
 mkdir -p "${log_home}"
+rm -rf "${log_home}"/*
```
sebsura, line 8: Good catch!

systemtests/tests/python-bareos/test_list_command.py
```diff
@@ -245,9 +247,9 @@ def test_list_jobs(self):
             self.assertTrue(job["jobstatus"], "T")
 
         # running a job a canceling
-        director.call("run job=backup-bareos-fd yes")
-        director.call("cancel job=backup-bareos-fd all yes")
-        director.call("wait")
+        jobid = self.run_job(director, "backup-bareos-fd")
+        director.call(f"cancel jobid={jobid} all yes")
```
sebsura, line 251: Does the `all` still make sense here as we specify exactly one jobid ?

systemtests/tests/python-bareos/test_list_command.py
```diff
@@ -58,10 +58,12 @@ def test_list_jobs(self):
...
-        director.call("run job=backup-bareos-fd yes")
-        director.call("wait")
-        director.call("restore client=bareos-fd fileset=SelfTest select all done yes")
-        director.call("wait")
+        backup_jobname = "backup-bareos-fd"
+        backup_jobid = self.get_backup_jobid(director, backup_jobname, level="Full")
+        self.run_restore(director, client="bareos-fd", jobid=backup_jobid, wait=True)
+        backup_jobid_incremental = self.get_backup_jobid(
+            director, backup_jobname, level="Incremental"
+        )
```
sebsura, line 66: previously we didnt run an incremental here.  The incremental jobid is also not used.  What changed that we now want to run this ?

systemtests/python-modules/bareos_unittest/json.py
```diff
@@ -127,31 +133,37 @@ def run_restore(
             run_parameter.append("jobid={}".format(jobid))
         if fileset:
             run_parameter.append("fileset={}".format(fileset))
+            run_parameter.append("select")
```
sebsura, line 136: Why does this only need to happen when a fileset is specified ?

At least on the bconsole
```
*restore jobid=1 where=/tmp/bareos fileset=SelfTest all done yes
```
works.

systemtests/python-modules/bareos_unittest/json.py
```diff
@@ -127,31 +133,37 @@ def run_restore(
...
+        run_parameter += ["all", "done", "yes"]
+        restore_cmd = "restore {}".format(" ".join(run_parameter))
+        result = director.call(restore_cmd)
         jobId = result["run"]["jobid"]
         if wait:
             self.wait_job(director, jobId)
         return jobId
 
-    def get_backup_jobid(self, director, jobname, level="F", extra=None):
+    def get_backup_jobid(self, director, jobname, level=None, fileset=None, extra=None):
```
sebsura, line 147: Some tests do not set `level` but it seems they do require a full.  Could you double check this ?

E.g. `test_list_media` seems to require a full but does not enforce it now.

systemtests/python-modules/bareos_unittest/json.py
```diff
@@ -127,31 +133,37 @@ def run_restore(
...
-        try:
-            jobid = result["jobs"][0]["jobid"]
-            return jobid
-        except IndexError:
-            # there is no valid backup for these parameters.
-            # run a backup job.
-            return self.run_job(director, jobname, level, extra, wait=True)
+        if len(result["jobs"]) >= 1:
+            job = result["jobs"][0]
+            # TODO: it would be more efficient, if the list command could by by fileset.
```
sebsura, line 161:
```suggestion
            # TODO: it would be more efficient, if the list command could filter by fileset.
```

systemtests/tests/python-bareos/test_list_command.py
```diff
@@ -335,8 +384,7 @@ def test_list_media(self):
             **self.director_extra_options,
         )
 
-        director.call("run job=backup-bareos-fd yes")
-        director.call("wait")
+        self.get_backup_jobid(director, "backup-bareos-fd")
```
sebsura, line 387: I feel like this should require `level=Full` here when looking at the following lines.

webui/module/Client/src/Client/Model/ClientModel.php
sebsura: I guess the idea is that the caller checks this and if he doesnt then `call_json` will throw the exception ?
Im fine with that

sebsura (changes requested):
I still need to look at some php files but as it might take some time i wanted to already give you some feedback.

core/src/dird/ua_run.cc
```diff
@@ -2193,10 +2193,21 @@ static bool ScanCommandLineArguments(UaContext* ua, RunContext& rc)
...
+            || (!ua->AclAccessOk(Client_ACL, cr.Name, true))) {
+          ua->ErrorMsg(T_("Client \"%s\" not found.\n"), rc.client_name);
+          return false;
+        }
+      } else {
+        if (*rc.client_name != 0) {
+          ua->WarningMsg(T_("Client \"%s\" not found.\n"), rc.client_name);
+        }
+        rc.client = select_client_resource(ua);
+        if (!rc.client) { return false; }
```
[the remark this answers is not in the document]
joergsteffens: Added a comment.

core/src/dird/ua_select.cc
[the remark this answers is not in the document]
joergsteffens: added.

systemtests/tests/python-bareos/test_list_command.py
joergsteffens: We need the incremental for some later tests in this function. Running it has just been moved to an earlier point in time.

webui/vendor/Bareos/library/Bareos/BSock/BareosBSock.php
sebsura: Hmm... should we adapt `call_fullresult_json` so that it detects those conversion errors then ?  As currently we just silently ignore json decode failures (see comment below).
From what i have gathered we can use `json_last_error()` to do so but im not sure what version of php we are targeting.

webui/module/Restore/view/restore/restore/versions.phtml
```diff
@@ -502,6 +506,41 @@ function updateRestoreParams(k, v) {
...
 }
 
+   function hasValue(array) {
+      for (let i = 0; i < array.length; i++) {
+         let id = array[i];
+         if(!document.getElementById(id).value) {
+            //console.log(id + ": missing");
+            return false;
+         }
+         //console.log(id + ": okay");
```
sebsura, line 516: Remove comments if not needed anymore.

webui/module/Restore/view/restore/restore/index.phtml
```diff
@@ -57,36 +58,37 @@ $this->headTitle($title);
...
 <?php echo $this->result; ?>
          </code></pre>
       </div>
       <div class="modal-footer">
-         <?php $jobid = substr($this->result, stripos($this->result, 'JobId=') + 6 ); ?>
-         <a class="btn btn-primary" href="<?php echo $this->basePath() . '/job/details/' . $jobid; ?>"><?php echo $this->translate("Job report"); ?></a>
+         <?php if (isset($this->restore_jobid)): ?>
+            <a class="btn btn-primary" href="<?php echo $this->basePath() . '/job/details/' . $this->restore_jobid; ?>"><?php echo $this->translate("Job report"); ?></a>.
+         <?php endif; ?>
          <button type="button" class="btn btn-primary" data-dismiss="modal"><?php echo $this->translate("Close"); ?></button>
```
sebsura, line 78: I feel like this should be a default button so as to not have 2 primary buttons on the same form. What do you think ?

webui/module/Restore/src/Restore/Form/RestoreForm.php
```diff
@@ -32,225 +32,91 @@ class RestoreForm extends Form
...
+            'options' => array(
+                'label' => _('Client'),
+                'empty_option' => _('Please choose a client'),
+                'value_options' => $this->restore_source_clients
+            ),
+            'attributes' => array(
+                'class' => 'form-control selectpicker show-tick',
+                'data-live-search' => 'true',
+                'id' => 'client',
+                'value' => $restore_params['client']
```
sebsura, line 63: Why do we use `$restore_params` here but `$this->restore_params` below ?

webui/module/Restore/src/Restore/Form/RestoreForm.php
```diff
@@ -32,225 +32,91 @@ class RestoreForm extends Form
...
+                'label' => _('Restore to client'),
+                'empty_option' => _('Please choose a client'),
+                'value_options' => $this->restore_target_clients
+            ),
+            'attributes' => array(
+                'class' => 'form-control selectpicker show-tick',
+                'data-live-search' => 'true',
+                'id' => 'restoreclient',
+                'value' => $this->restore_params['restoreclient'] ?? $this->restore_params['client'],
+                'disabled' => (!isset($this->restore_params['client']))
```
sebsura, line 81: Do we not already assume that this is set when setting the client above ?

sebsura: (Also applies to all the checks found below)

webui/module/Restore/src/Restore/Form/RestoreForm.php
```diff
@@ -599,137 +435,95 @@ public function __construct($restore_params = null, $clients = null, $filesets =
...
+            ),
+            'attributes' => array(
+                //'value' => _('Restore'),
+                'id' => 'btn-form-submit',
+                'class' => 'btn btn-primary',
+                // will be enabled by JS
+                'disabled' => true
+            )
+        );
+        $this->add($form_submit);
```
sebsura, line 452: This can probably be done with `$this->add(array(...))` like the others above.

webui/module/Restore/src/Restore/Form/RestoreForm.php
```diff
@@ -599,137 +435,95 @@ public function __construct($restore_params = null, $clients = null, $filesets =
...
      *
      */
-    private function getFilesetList()
+    private function getBackupList()
     {
         $selectData = array();
-        if (!empty($this->filesets)) {
-            foreach ($this->filesets as $fileset) {
-                $selectData[$fileset['name']] = $fileset['name'];
+        foreach ($this->backups as $backup) {
```
sebsura, line 492: The constructor sets `this->backups` to `null`  by default, so we probably need the `!empty($this->backups)` check here.

webui/module/Restore/src/Restore/Form/RestoreForm.php
```diff
@@ -599,137 +435,95 @@ public function __construct($restore_params = null, $clients = null, $filesets =
...
      */
-    private function getRestoreJobWhereDirectives()
+    private function getRestoreJobList()
     {
         $selectData = array();
-        if (!empty($this->restorejobresources)) {
-            foreach ($this->restorejobresources as $restorejob) {
-                $selectData[$restorejob['where']] = $restorejob['where'];
-            }
+        foreach ($this->restorejobresources as $key => $value) {
```
sebsura, line 523: For the same reason we also need an empty check here i believe.

sebsura (changes requested):

webui/module/Restore/src/Restore/Form/RestoreForm.php
joergsteffens: Not sure what you mean. If you refer to "value": if `$restore_params['client']` is not set, `"value"` will be `Null`, which is fine.

webui/module/Restore/src/Restore/Form/RestoreForm.php
joergsteffens: done

webui/module/Restore/src/Restore/Form/RestoreForm.php
joergsteffens: done

webui/module/Restore/src/Restore/Form/RestoreForm.php
joergsteffens: done

webui/module/Restore/view/restore/restore/versions.phtml
```diff
@@ -29,96 +29,97 @@ $this->headTitle($title);
...
-         </div>
+  <div class="modal-dialog modal-lg">
+    <div class="modal-content">
+      <div class="modal-header">
+        <button type="button" class="close" data-dismiss="modal" aria-label="Close"><span aria-hidden="true">&times;</span></button>
+        <h4 class="modal-title" id="myModalLabel"></h4>
       </div>
-   </div>
+      <div class="modal-body">
+         <p><?php echo $this->translate("No file selected to restore!"); ?></p>
```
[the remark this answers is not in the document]
joergsteffens: I changed the English text to use your suggestion. I can only adapt the German text after the language files have been updated on https://poeditor.com/projects/view?id=180893 (which can happen after the merge).

webui/module/Restore/view/restore/restore/versions.phtml
joergsteffens: done

webui/vendor/Bareos/library/Bareos/BSock/BareosBSock.php
joergsteffens: In case` \Zend\Json\Json::decode` fails because of an invalid json string, it itself raises an `Zend\\Json\\Exception\\RuntimeException` exception. So this is probably fine (except that we do not handle that exception properly, but that is another story).

webui/module/Restore/src/Restore/Form/RestoreForm.php
sebsura: I think i made the wrong assumption that accessing a non existing key in an array throws an exception, but testing it in the interactive shell shows that it "only" prints a warning, so i guess there is nothing to do here!

webui/vendor/Bareos/library/Bareos/BSock/BareosBSock.php
sebsura: Im a bit confused.  In an earlier comment you said
> When Python fails to convert to JSON, it generates an exception on its own. In PHP this is not the case

So im not quite sure whether it generates an exception or it doesnt.

Im not quite sure how the check the behaviour of ZendJson but from what i can tell it acts like a polyfill for json_decode() which does not throw an exception but instead just returns null.

https://docs.zendframework.com/zend-json/basics/
> By default, the above two calls will proxy to the json_decode() and json_encode() functions of ext/json, which is bundled in default installations of PHP [...]

sebsura: So looking into this it looks like Zend does throw an exception on invalid json. Weird!

sebsura (approved):
Great Changes!

PR #1952, opened 2024-09-11
filed: detect integer overflow during backup

pstorz:
The filedaemon now makes sure that the fileindex does not overflow during the backup.  The upper limit for file indices is currently 2^(31) - 1, but we now also build a test file daemon that has this limit set to 20, so that we can actually test the overflow detection.

#### Please check

##### General
- [x] Required backport PRs have been created : #2024 

##### Source code quality

##### Tests


core/src/filed/backup.cc
```diff
@@ -1604,7 +1604,12 @@ bool EncodeAndSendAttributes(JobControlRecord* jcr,
 
   {
     std::unique_lock l(jcr->mutex_guard());
-    jcr->JobFiles++;                   /* increment number of files sent */
+    int IncreasedJobFiles;
+    if (__builtin_sadd_overflow(jcr->JobFiles, 1, &IncreasedJobFiles)) {
+      Jmsg0(jcr, M_FATAL, 0, T_("jcr->JobFiles overflow prevented!\n"));
+      return false;
+    }
+    jcr->JobFiles = IncreasedJobFiles; /* increment number of files sent */
```
sebsura, line 1612:
```suggestion
    if (jcr->JobFiles == std::numeric_limits<decltype(jcr->JobFiles)>::max()) {
      Jmsg0(jcr, M_FATAL, 0, T_("You are trying to back up too many files at once.  Please split up your fileset.\n"));
      return false;
    }
    jcr->JobFiles += 1; /* increment number of files sent */
```

sebsura: `__builtin_sadd_overflow` is a gcc builtin and as such not available on windows.   `jcr->JobFiles` is also unsigned, why are we using a simple int here ? If we want to ensure that there is no overflow even with a signed value we should use something like
```
static_cast<std::make_signed_t<decltype(jcr->JobFiles)>>(jcr->JobFiles) >= 0
```
after the increment.

sebsura (changes requested):

core/src/filed/backup.cc
arogge: I'd even go so far to use a macro like this
```
#if not defined JOB_FILES_MAX
#define JOB_FILES_MAX std::numeric_limits<decltype(JobControlRecord::JobFiles)::max())
#endif
```
Because then we could build with `-DJOB_FILES_MAX=20` and run a proper test.

sebsura (changes requested):
We should probably add checks to `::SendBaseFileList` and `::SendDeletedList`; either they should check for the return value of `EncodeAndSendAttributes` or they should check if the job was canceled and stop.

Maybe even both.

core/src/filed/backup.cc
```diff
@@ -1605,6 +1605,13 @@ bool EncodeAndSendAttributes(JobControlRecord* jcr,
   {
     std::unique_lock l(jcr->mutex_guard());
     jcr->JobFiles++;                   /* increment number of files sent */
+    if (jcr->JobFiles > MAXIMUM_ALLOWED_FILES_PER_JOB) {
+      Jmsg1(jcr, M_FATAL, 0,
+            "FileIndex overflow detected. Please split the fileset to backup "
+            "less files.\n");
+      return false;
+    }
+
```
sebsura, line 1614: This does not work.  You cannot be bigger than the max.

The correct way to check for this is
```
if (jcr->JobFiles == MAXIMUM_ALLOWED_FILES_PER_JOB) { ... }
jcr->JobFiles += 1;
```
Currently `jcr->JobFiles` will overflow causing the check to always be false.

core/src/filed/backup.h
```diff
@@ -72,6 +72,10 @@ void StripPath(FindFilesPacket* ff_pkt);
 void UnstripPath(FindFilesPacket* ff_pkt);
 int SaveFile(JobControlRecord* jcr, FindFilesPacket* ff_pkt, bool top_level);
 
+// allow to set the allowed maximum to something lower for testing purposes
+#ifndef MAXIMUM_ALLOWED_FILES_PER_JOB
+#define MAXIMUM_ALLOWED_FILES_PER_JOB std::numeric_limits<decltype(JobControlRecord::JobFiles)>::max()
+#endif
```
sebsura, line 78: This is fine for now, but we should think of a way to do this in which it is not only testable but also _tested_.

Maybe for 25 we can change this to be a run time value that you can set when starting the FD.  This way we can actually set up an automated test for it!

sebsura (changes requested):

core/src/filed/backup.cc
```diff
@@ -1602,11 +1613,15 @@ bool EncodeAndSendAttributes(JobControlRecord* jcr,
...
-    std::unique_lock l(jcr->mutex_guard());
-    jcr->JobFiles++;                   /* increment number of files sent */
-    ff_pkt->FileIndex = jcr->JobFiles; /* return FileIndex */
+  if (std::optional findex = get_next_findex(jcr)) {
+    ff_pkt->FileIndex = *findex;
     PmStrcpy(jcr->fd_impl->last_fname, ff_pkt->fname);
+  } else {
+    Jmsg1(jcr, M_FATAL, 0,
+          "FileIndex overflow detected. Please split the fileset to backup "
+          "less files.\n");
```
sebsura, line 1622:
```suggestion
          "fewer files.\n");
```

core/src/filed/backup.cc
```diff
@@ -1602,11 +1613,15 @@ bool EncodeAndSendAttributes(JobControlRecord* jcr,
...
 
-  {
-    std::unique_lock l(jcr->mutex_guard());
-    jcr->JobFiles++;                   /* increment number of files sent */
-    ff_pkt->FileIndex = jcr->JobFiles; /* return FileIndex */
+  if (std::optional findex = get_next_findex(jcr)) {
+    ff_pkt->FileIndex = *findex;
     PmStrcpy(jcr->fd_impl->last_fname, ff_pkt->fname);
+  } else {
+    Jmsg1(jcr, M_FATAL, 0,
```
sebsura, line 1620: Currently this causes the backup to end in state `*** Backup Error ***`.  Should we maybe set the state to `JS_Incomplete` instead ?  Then the end state will show as `Backup failed -- incomplete`

Im not sure which one is better, what do you think @pstorz  ?

sebsura (changes requested):

sebsura (approved):

PR #2016, opened 2024-11-15
alist: fix race condition while iterating

sebsura:
**Backport of PR #1859 to bareos-23**

Also removed `foreach_alist` in `bnet_server_tcp` as it did not exist there in 24 anymore.

##### General
- [x] Backport to 22 created

#### Backport quality
- [x] Original PR #1859 is merged

### DIFF
--- a/core/src/dird/fd_cmds.cc
+++ b/core/src/dird/fd_cmds.cc
@@ -580,6 +580,8 @@ static bool HaveClientRunscripts(alist<RunScript*>* RunScripts)
 {
   bool retval = false;
 
+  if (!RunScripts) { return false; }
+
   if (RunScripts->empty()) { return false; }
 
   for (auto* cmd : *RunScripts) {
--- a/core/src/dird/fd_cmds.cc
+++ b/core/src/dird/fd_cmds.cc
@@ -582,7 +582,7 @@ static bool HaveClientRunscripts(alist<RunScript*>* RunScripts)
 
   if (RunScripts->empty()) { return false; }
 
-  foreach_alist (cmd, RunScripts) {
+  for (auto* cmd : *RunScripts) {
     if (!cmd->IsLocal()) { retval = true; }
   }
 
@@ -609,7 +609,7 @@ int SendRunscriptsCommands(JobControlRecord* jcr)
 
   msg = GetPoolMemory(PM_FNAME);
   ehost = GetPoolMemory(PM_FNAME);
-  foreach_alist (cmd, jcr->dir_impl->res.job->RunScripts) {
+  for (auto* cmd : *jcr->dir_impl->res.job->RunScripts) {
     if (!cmd->target.empty()) {
       ehost = edit_job_codes(jcr, ehost, cmd->target.c_str(), "");
       Dmsg2(200, "dird: runscript %s -> %s\n", cmd->target.c_str(), ehost);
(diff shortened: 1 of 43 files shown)


core/src/dird/fd_cmds.cc
```diff
@@ -578,12 +578,13 @@ bool SendIncludeExcludeLists(JobControlRecord* jcr)
 // This checks to see if there are any non local runscripts for this job.
 static bool HaveClientRunscripts(alist<RunScript*>* RunScripts)
 {
-  RunScript* cmd = nullptr;
   bool retval = false;
 
+  if (!RunScripts) { return false; }
```
florian-at-bareos, line 583: What do you think about taking `RunScripts` as a reference instead of a pointer? Usually I would always prefer this over taking the arg as a pointer and then check for NULL.

if you want to keep the check for null, maybe pass a `JobResource` reference and then check
```cpp
if (!job->RunScripts) { return false; }
```

florian-at-bareos (commented):
Thanks for your work! The changes look very good!

core/src/dird/storage.cc
```diff
@@ -245,21 +237,23 @@ void SetPairedStorage(JobControlRecord* jcr)
...
             = new alist<StorageResource*>(10, not_owned_by_alist);
-        foreach_alist (store,
-                       jcr->dir_impl->res.paired_read_write_storage_list) {
-          if (store->paired_storage) {
-            Dmsg1(100, "write_storage_list=%s\n",
-                  store->paired_storage->resource_name_);
-            jcr->dir_impl->res.write_storage_list->append(
-                store->paired_storage);
+        if (jcr->dir_impl->res.paired_read_write_storage_list) {
+          for (auto* store :
```
florian-at-bareos, line 241: Maybe it's just me, but I was kinda confused at first when I saw this
```
if (jcr->dir_impl->res.paired_read_write_storage_list) {
    for (auto* store : jcr->dir_impl->res.paired_read_write_storage_list) {
       ...
    }
}
```
for me, it was not clear that `paired_read_write_storage_list` is a pointer due to the fact that we iterate over it with a ranged-for.

So i thought 'why do we change the list whether it's empty or not if only plan to iterate over it?'.

Maybe we can make this more explicit by dereferencing the pointer in the ranged-for
```
for (auto* store : *jcr->dir_impl->res.paired_read_write_storage_list) {
       ...
}
```
and then remove the `begin(alist<T>* list)`  altogether?

florian-at-bareos:
but maybe it's also just too much for just a backport actually!

core/src/dird/storage.cc
arogge: That was basically the original implementation, but that requires a nullptr-check before each of the loops.
To be more "backwards compatible" to what `foreach_alist()` did (i.e. require less code changes for the refactoring), we decided to use a free function that will just skip the loop when the provided list is a nullptr.

core/src/dird/fd_cmds.cc
sebsura: This function doesnt need anything from the job resource (apart from the RunScripts), so im not sure if this would really be a good idea.  As this function is only called once i would much rather just remove it entirely (and use std::find_if or something like that), but as this is a backport i would not change this anymore.

florian-at-bareos (approved):

PR #2159, opened 2025-02-07
config: fix issues with config directive aliases

florian-at-bareos:
Fixes internal issue#167

Currently, aliases are defined as new directives with the flag CFG_ITEM_ALIAS.
The problem with this approach is that nowhere is specified which directive this is an alias to.
This information is important, e.g. if the original directive is marked as CFG_ITEM_REQUIRED, because specifying one of its aliases should suffice the requirement.
Also, outputting the parsed config wasn't working with aliases, because they were simply skipped.

To solve this i added a new member ResourceItem::aliases (std::set\<std::string\>) to track the aliases of a directive.

This PR
- marks 'Address' (Console -> Director) as required (currently, not specifying the 'Address' passes the config parsing successfully but then always crashes when trying to connect to the director since address is NULL)
- introduces a new alias 'DirAddress' (Console -> Director) for it
- makes the config parser output a warning if aliases are being used (the bconsole does not output config warnings currently, changed that)
- replaces all config aliases used in (system)tests

#### Please check

##### General

##### Source code quality


florian-at-bareos:
I still need to fix the documentation for the aliases.

core/src/console/console.cc
```diff
@@ -979,6 +979,14 @@ int main(int argc, char* argv[])
           my_config->get_base_config_path().c_str());
   }
 
+  if (my_config->HasWarnings()) {
+    // messaging not initialized, so Jmsg with  M_WARNING doesn't work
+    fprintf(stderr, T_("There are configuration warnings:\n"));
+    for (auto& warning : my_config->GetWarnings()) {
+      fprintf(stderr, " * %s\n", warning.c_str());
```
sebsura, line 986: We should probably switch this to use `Emsg` like the errors above.  If you use `M_WARNING` then the function should not exit the daemon, but you should test that out.

core/src/lib/parse_conf.cc
```diff
@@ -316,20 +317,39 @@ ResourceTable* ConfigurationParser::GetResourceTable(
...
 }
 
+static bool EqualsIgnoreCase(std::string_view v1, std::string_view v2)
+{
+  if (v1.length() != v2.length()) { return false; }
+  for (size_t i = 0; i < v1.length(); ++i) {
+    if (std::tolower(v1.at(i)) != std::tolower(v2.at(i))) { return false; }
+  }
+  return true;
+}
```
sebsura, line 327: remove (see below)

core/src/lib/parse_conf.cc
```diff
@@ -316,20 +317,39 @@ ResourceTable* ConfigurationParser::GetResourceTable(
...
-  for (i = 0; resource_items_[i].name; i++) {
-    if (Bstrcasecmp(resource_items_[i].name, item)) {
-      result = i;
+  for (int i = 0; resource_items_[i].name; i++) {
+    if (EqualsIgnoreCase(resource_items_[i].name, item)) {
+      return i;
       break;
+    } else {
+      for (const auto& alias : resource_items_[i].aliases) {
+        if (EqualsIgnoreCase(alias, item)) {
```
sebsura, line 338: As `alias` is an `const std::string&`, we should just do
```
Bstrcasecmp(alias.c_str(), item)
```

sebsura: I dont think duplicating this logic (i.e. Bstrcasecmp and EqualsIgnoreCase) here is good as it is easily preventable.

If we ever want to do this properly we would have to find all these implementations and fix them one by one.

core/src/lib/resource_item.h
```diff
@@ -50,6 +53,8 @@ struct ResourceItem {
    * Every new directive should have a description. */
   const char* description;
 
+  std::set<std::string> aliases = {};
```
sebsura, line 56: As we search for the aliases ourselves anyways, we could just make this a vector.

sebsura: With minimal changes we can even simply use
```
std::initializer_list<const char*>
```
here, which does not use any allocations.  This would have the upside of not preventing us from making ResourceItem `constexpr` in the future.

core/src/lib/res.cc
```diff
@@ -2113,8 +2107,16 @@ json_t* json_item(ResourceItem* item)
...
-  if (item->description) {
-    json_object_set_new(json, "description", json_string(item->description));
+  if (!is_alias) {
+    if (item->description) {
+      json_object_set_new(json, "description", json_string(item->description));
+    }
+  } else {
+    std::string alias_description
+        = std::string("This is an alias, consider using \"") + item->name
+          + "\" instead.";
```
sebsura, line 2117: This should probably still contain the "real" description as well.

docs/manuals/source/include/autogenerated/bareos-dir-config-schema.json
```diff
@@ -1837,33 +1845,42 @@
...
           "code": 0,
           "equals": true
         },
+        "Password": {
+          "datatype": "AUTOPASSWORD",
+          "code": 0,
+          "alias": true,
+          "equals": true,
+          "description": "This is an alias, consider using \"DbPassword\" instead."
+        },
```
sebsura, line 1883: My understanding was that we did not want new aliases for the db stuff.

core/src/lib/res.cc
```diff
@@ -2072,24 +2069,21 @@ bool BareosResource::PrintConfig(OutputFormatterResource& send,
  *     "code": int
  *     [ "defaultvalue": "xyz", ]
  *     [ "required": true, ]
- *     [ "alias": true, ]
```
sebsura, line 2075: Why was this removed ? Currently it still adds this field if needed.

sebsura (changes requested):
This looks very good. Nice work!

core/src/console/console.cc
sebsura: Also for the other daemons!

core/src/lib/resource_item.h
florian-at-bareos: You're right! I would probably change it to vector because initializer_list is in my opinion not that self explanatory and also uses const char* which we want to avoid.

docs/manuals/source/include/autogenerated/bareos-dir-config-schema.json
florian-at-bareos: Password already existed before. But i added "Socket" for "DbSocket" for the sake of consistency because all other database properties had an alias to the name without the prefix already. I didn't understand it that way that we shouldn't introduce new aliases for the db stuff.

core/src/lib/res.cc
florian-at-bareos: yes you're right!

core/src/console/console.cc
florian-at-bareos: This works only for bconsole, for the other ones the message is not displayed like stated in the comment.

florian-at-bareos: I would suggest to leave it like that because the formatting is a lot different if we use fprintf in comparison to Emsg and i think this  should be somewhat consistent across bareos applications

florian-at-bareos: for Emsg:
14-Feb 09:47 bconsole: Warning: Found alias usage "DirAddress" in configuration which is discouraged, consider using "Address" instead.

core/src/lib/res.cc
florian-at-bareos: if we want to do this i would say that we need a new line operation for this. But I'm unsure how to do since this depends on where the description is displayed (e.g. if the description is pasted to html one cannot just use \n it think)

sebsura (approved):
Great Work!

PR #2176, opened 2025-02-14
various cmake improvements

arogge:
This PR tries to clean up cmake.
The main goal is to leverage target-dependencies instead of plain library-links. As we linked plain libraries instead of targets by mistake, the PR should also make this impossible.

For debugging our project now supports a flag `DEBUG_TARGET_DEPENDENCIES` that will dump (hopefully helpful) information about target dependencies into the binary dir.

* find-modules for more libraries
* support single-output directory, so vcpkg dll-magic will work (see #2211)
* remove support for multi-build generators

#### Please check

##### General

##### Source code quality


core/src/stored/CMakeLists.txt
```diff
@@ -144,77 +64,136 @@ if(${HAVE_DYNAMIC_SD_BACKENDS})
     list(APPEND BACKEND_OBJECTS libbareossd-chunked)
   endif()
 
-  set(LIBBAREOSSD_LIBRARIES bareos ${THREADS_THREADS} ${BACKEND_OBJECTS})
```
sebsura, line 147: why do we not define bareossd/Bareos::LibSD before this and simply use `target_link_libraries()` here ?

This would also save us the same if later on!

sebsura:
```cmake
add_library(bareossd-backends INTERFACE)

if(HAVE_DYNAMIC_SD_BACKENDS)
  target_sources(bareossd-backends INTERFACE sd_backends_dynamic.cc)
else()
  target_link_libraries(bareossd-backends INTERFACE Bareos::Lib)
  if(TARGET droplet)
    target_link_libraries(bareossd-backends INTERFACE droplet)
    target_compile_definitions(bareossd-backends INTERFACE HAVE_BAREOSSD_DROPLET_DEVICE)
  endif()
  if(HAVE_GFAPI)
    target_link_libraries(bareossd-backends INTERFACE ${GFAPI_LIBRARIES})
  endif()
endif()
```

and then later on we can simply do
```diff
-if(HAVE_DYNAMIC_SD_BACKENDS)
-  target_sources(bareossd PRIVATE sd_backends_dynamic.cc)
-else()
-  target_link_libraries(bareossd PRIVATE Bareos::Lib ${LIBBAREOSSD_LIBRARIES})
-endif()
-if(TARGET droplet)
-  target_compile_definitions(bareossd PRIVATE HAVE_BAREOSSD_DROPLET_DEVICE)
-endif()
+ target_link_libraries(bareossd PRIVATE bareossd-backends)
```

NOTE: I only tested the dynamic option, but jenkins does test both!

core/src/stored/CMakeLists.txt
```diff
@@ -144,77 +64,136 @@ if(${HAVE_DYNAMIC_SD_BACKENDS})
...
   bareossd PROPERTIES VERSION "${BAREOS_NUMERIC_VERSION}"
                       SOVERSION "${BAREOS_VERSION_MAJOR}"
 )
 
-if(${HAVE_DYNAMIC_SD_BACKENDS})
+if(HAVE_DYNAMIC_SD_BACKENDS)
   target_sources(bareossd PRIVATE sd_backends_dynamic.cc)
 else()
-  target_link_libraries(bareossd PRIVATE ${LIBBAREOSSD_LIBRARIES})
+  target_link_libraries(bareossd PRIVATE Bareos::Lib ${LIBBAREOSSD_LIBRARIES})
```
sebsura, line 122: why do we only link with bareos::lib when we have static backends ?

core/src/stored/CMakeLists.txt
```diff
@@ -144,77 +64,136 @@ if(${HAVE_DYNAMIC_SD_BACKENDS})
...
+endif()
+if(TARGET droplet)
+  target_compile_definitions(bareossd PRIVATE HAVE_BAREOSSD_DROPLET_DEVICE)
```
sebsura, line 125: This appears unused => delete

core/src/stored/CMakeLists.txt
```diff
@@ -144,77 +64,136 @@ if(${HAVE_DYNAMIC_SD_BACKENDS})
...
 
-  set(LIBBAREOSSD_LIBRARIES bareos ${THREADS_THREADS} ${BACKEND_OBJECTS})
   message(STATUS "BACKEND_OBJECTS ARE ${BACKEND_OBJECTS}")
 else()
   if(TARGET droplet)
     list(APPEND LIBBAREOSSD_LIBRARIES droplet)
   endif()
-  if(${HAVE_GFAPI})
+  if(HAVE_GFAPI)
     list(APPEND LIBBAREOSSD_LIBRARIES ${GFAPI_LIBRARIES})
```
sebsura, line 73: shouldnt we use `PkgConfig::gfapi` here ?

core/src/stored/backends/CMakeLists.txt
```diff
@@ -36,8 +36,7 @@ endmacro()
 if(${HAVE_GLUSTERFS})
```
sebsura, line 36:
```suggestion
if(HAVE_GLUSTERFS)
```

core/src/stored/CMakeLists.txt
```diff
@@ -144,77 +64,136 @@ if(${HAVE_DYNAMIC_SD_BACKENDS})
...
   endif()
 
-  set(LIBBAREOSSD_LIBRARIES bareos ${THREADS_THREADS} ${BACKEND_OBJECTS})
   message(STATUS "BACKEND_OBJECTS ARE ${BACKEND_OBJECTS}")
 else()
   if(TARGET droplet)
     list(APPEND LIBBAREOSSD_LIBRARIES droplet)
   endif()
-  if(${HAVE_GFAPI})
+  if(HAVE_GFAPI)
```
sebsura, line 72: in this file we check for `HAVE_GFAPI` and `GFAPI_LIBRARIES` whereas in `backends` we check for `HAVE_GLUSTERFS`.  Are they different somehow ? If not => delete/unify them.
If they are, then this doesnt seem correct and should check for the same as the actual target (unless we need the gfapi for something other than the glusterfs backend).

core/src/tests/thread_list.cc
```diff
@@ -90,6 +90,7 @@ static void* ShutdownCallback(void* data)
 
 static constexpr int maximum_thread_count = 10;
 
+#if !defined(WIN32) && !defined(_WIN32) && !defined(__WIN32)
```
sebsura, line 93: Do the tests not have access to `HAVE_WIN32` ?

core/src/tools/CMakeLists.txt
```diff
@@ -19,84 +19,71 @@
...
-
-  set(TOOLS_SBIN bscrypto bwild bregex)
-else()
-  set(TOOLS_SBIN bwild bregex)
-endif()
-set(TOOLS_BIN bsmtp bwild bregex)
-
-if(NOT HAVE_WIN32)
-  list(APPEND TOOLS_SBIN bpluginfo btestls)
+  target_link_libraries(bpluginfo Bareos::Lib ${CMAKE_DL_LIBS} CLI11::CLI11)
```
sebsura, line 65: if we want use `CMAKE_DL_LIBS` we should probably also add it to the other libraries that need it.  E.g. `sd_backends_dynamic.cc`, `Bareos::Lib`, `grpc/bridge_module`.

sebsura (approved):

sebsura (changes requested):
misclicked, and cannot change it now.

core/src/cats/CMakeLists.txt
```diff
@@ -18,9 +18,8 @@
 #   02110-1301, USA.
 message("Entering ${CMAKE_CURRENT_SOURCE_DIR}")
 
-include_directories(${PostgreSQL_INCLUDE_DIR})
-
 add_library(bareossql SHARED)
+add_library(Bareos::SQL ALIAS bareossql)
```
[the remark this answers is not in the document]
arogge: Because CMake forbids you to have library/executable targets with a colon in its name.

core/src/droplet/libdroplet/CMakeLists.txt
```diff
@@ -113,10 +113,9 @@ bareos_disable_warnings(
 
 target_include_directories(
   droplet PUBLIC include ${LIBXML2_INCLUDE_DIR} ${JSONC_INCLUDE_DIRS}
-                 ${OPENSSL_INCLUDE_DIR}
 )
 target_link_libraries(
-  droplet ${LIBXML2_LIBRARIES} ${JSONC_LIBRARIES} ${OPENSSL_LIBRARIES}
+  droplet ${LIBXML2_LIBRARIES} ${JSONC_LIBRARIES} OpenSSL::SSL
```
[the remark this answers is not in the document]
arogge: I'd rather just change the `include(FindOpenSSL)` there to the more correct `find_package(OpenSSL REQUIRED)`

core/src/tests/thread_list.cc
arogge: we'd have to include `config.h` for that

core/cmake/BareosFindAllLibraries.cmake
```diff
@@ -136,51 +131,36 @@ elseif(
...
 if(ENABLE_LZO)
-  bareosfindlibraryandheaders("lzo2" "lzo/lzoconf.h" "")
-  if(${LZO2_FOUND})
-    set(HAVE_LZO 1)
-    if(${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
-      set(LZO2_LIBRARIES "${HOMEBREW_PREFIX}/opt/lzo/lib/liblzo2.a")
-    endif()
-  endif()
+  find_package(LZO REQUIRED)
+  set(HAVE_LZO 1)
```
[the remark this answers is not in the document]
arogge: > `FindLZO` already sets this according to its doc comment.

I'll go and fix the docs then, because it doesn't seem to do that.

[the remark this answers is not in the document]
arogge: > Why dont we just use `#cmakedefine HAVE_LZO @LZO_FOUND@` ? I think this is the only use of these kinds of variables.

What `#cmakedefine <symbol> <...>` will evaluate to `#define <symbol> <...>` only if `symbol` evaluates to a non-false value in CMake, otherwise it will result in `/* #undef <symbol> */`.
So the only thing we could do is to directly use `#cmakedefine LZO_FOUND` and then change the places using `HAVE_LZO` to just check `defined(LZO_FOUND)`.
I don't know if it's worth the effort though.

All these `HAVE_xyz` defines are a remnant of autoconf. Basically `HAVE_xyz` is autoconf's version of CMake's `xyz_FOUND`.

core/src/lib/crypto_openssl.cc
```diff
@@ -141,17 +137,17 @@ IMPLEMENT_ASN1_FUNCTIONS(RecipientInfo)
 IMPLEMENT_ASN1_FUNCTIONS(SignatureData)
 IMPLEMENT_ASN1_FUNCTIONS(CryptoData)
 
-#    if OPENSSL_VERSION_NUMBER < 0x10100000L || defined(LIBRESSL_VERSION_NUMBER)
+#if OPENSSL_VERSION_NUMBER < 0x10100000L || defined(LIBRESSL_VERSION_NUMBER)
```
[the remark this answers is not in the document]
arogge: I guess that breaks the (currently untested) LibreSSL support that we might want to use on MacOS one day, but you're probably right.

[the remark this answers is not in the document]
sebsura: I assumed that `defined(LIBRESSL_VERSION_NUMBER)` is sufficient if we plan to support libressl.

core/src/stored/CMakeLists.txt
```diff
@@ -51,90 +52,9 @@ if(NOT build_client_only)
...
-if(HAVE_WIN32)
-  bareos_configure_file(FILES ../win32/stored/bscanres.rc.in)
-  list(APPEND BSCANSRCS ${CMAKE_CURRENT_BINARY_DIR}/../win32/stored/bscanres.rc)
-endif()
-
-set(BCOPYSRCS bcopy.cc)
-
-set(STORED_RESTYPES autochanger device director ndmp messages storage)
-
 set(BACKEND_OBJECTS "") # backend shared objects
```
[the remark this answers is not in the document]
sebsura: yes we write to it, but where is it read ? It looks like its only used for some build messages.

core/src/stored/CMakeLists.txt
arogge: Good catch! Something is seriously off here. I'll have to go and figure it out.

core/src/lib/crypto_openssl.cc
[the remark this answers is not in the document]
arogge: it is, but I just removed the whole ifdef now, we can still get it back if we ever need LibreSSL support.

core/src/stored/CMakeLists.txt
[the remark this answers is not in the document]
arogge: you're right, that's not used anymore.

core/src/stored/CMakeLists.txt
arogge: That was totally broken as we didn't have `HAVE_GFAPI` anymore due to a previous cleanup commit.
I merged `HAVE_GFAPI` and `HAVE_GLUSTERFS` now so things should be fine again.

core/src/stored/CMakeLists.txt
arogge: this was redundant, so I removed it

cmake/BareosWindowsResource.cmake
```diff
@@ -16,24 +16,18 @@
...
+      file(RELATIVE_PATH rel_file "${CMAKE_CURRENT_SOURCE_DIR}" "${in_file}")
+      set(out_file "${CMAKE_CURRENT_BINARY_DIR}/${rel_file}")
+    else()
+      set(out_file "${CMAKE_CURRENT_BINARY_DIR}/${in_file}")
+    endif()
+    string(REGEX REPLACE ".in\$" "" out_file "${out_file}")
+    message(STATUS "creating file ${out_file}")
+    configure_file(${in_file} ${out_file} @ONLY NEWLINE_STYLE UNIX)
+    target_sources(${target} PRIVATE ${out_file})
+  endif()
```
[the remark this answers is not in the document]
arogge: Yep, it is a lot better now without complicating bareos_configure_file() too much.

sebsura (approved):
There are a lot of great changes here.  Thanks for your work!

PR #2198, opened 2025-03-03
stored: implicitly create autochanger from device

florian-at-bareos:
automatically create autochanger and reserved device 0000 if the device resource has directive count != 1

#### Please check

##### General

##### Source code quality

##### Tests

### DIFF
--- a/core/src/stored/autochanger_resource.cc
+++ b/core/src/stored/autochanger_resource.cc
@@ -3,7 +3,7 @@
 
    Copyright (C) 2000-2011 Free Software Foundation Europe e.V.
    Copyright (C) 2011-2012 Planets Communications B.V.
-   Copyright (C) 2019-2024 Bareos GmbH & Co. KG
+   Copyright (C) 2019-2025 Bareos GmbH & Co. KG
 
    This program is Free Software; you can redistribute it and/or
    modify it under the terms of version three of the GNU Affero General Public
@@ -22,10 +22,10 @@
 */
 
 #include "stored/autochanger_resource.h"
+#include "stored_conf.h"
 #include "lib/alist.h"
 #include "stored/device_resource.h"
 #include "stored/stored_globals.h"
-
 namespace storagedaemon {
 
 AutochangerResource::AutochangerResource()
@@ -34,6 +34,8 @@ AutochangerResource::AutochangerResource()
     , changer_name(nullptr)
     , changer_command(nullptr)
 {
+  rcode_ = R_AUTOCHANGER;
+  rcode_str_ = "Autochanger";
   return;
 }
 
--- a/core/src/stored/stored_conf.cc
+++ b/core/src/stored/stored_conf.cc
@@ -3,7 +3,7 @@
 
    Copyright (C) 2000-2009 Free Software Foundation Europe e.V.
    Copyright (C) 2011-2012 Planets Communications B.V.
-   Copyright (C) 2013-2024 Bareos GmbH & Co. KG
+   Copyright (C) 2013-2025 Bareos GmbH & Co. KG
 
    This program is Free Software; you can redistribute it and/or
    modify it under the terms of version three of the GNU Affero General Public
@@ -432,38 +432,88 @@ static void ParseConfigCb(LEX* lc,
   }
 }
 
-static void MultiplyDevice(DeviceResource& multiplied_device_resource)
+static DeviceResource* CopyDevice(DeviceResource& original, int serial_number)
 {
-  /* append 0001 to the name of the existing resource */
-  multiplied_device_resource.CreateAndAssignSerialNumber(1);
+  DeviceResource* copy = new DeviceResource(original);
+  copy->CreateAndAssignSerialNumber(serial_number);
+  copy->multiplied_device_resource = std::addressof(original);
+  copy->count = 0;
+  copy->refcnt_ = 1;
+  if (copy->changer_res && copy->changer_res->device_resources) {
+    copy->changer_res->device_resources->append(copy);
+  }
 
-  multiplied_device_resource.multiplied_device_resource
-      = std::addressof(multiplied_device_resource);
+  return copy;
+}
+static void MultiplyDevice(DeviceResource& original,
+                           ConfigurationParser& config)
+{
+  // create device 0
+  std::string device0_name = original.resource_name_ + std::string("0000");
+  DeviceResource* device0 = dynamic_cast<DeviceResource*>(
+      config.GetResWithName(R_DEVICE, device0_name.c_str()));
+  if (!device0) {
+    device0 = CopyDevice(original, 0);
+    device0->autoselect = false;
+    config.AppendToResourcesChain(device0, device0->rcode_);
+  } else {
+    Dmsg0(0,
+          "Device resource \"%s\" will not implicitly create device \"%s\" "
+          "resource since there already exists a device resource with that "
+          "name.\n",
+          original.resource_name_, device0_name.c_str());
+  }
 
-  uint32_t count = multiplied_device_resource.count - 1;
+  // create autochanger
+  if (!original.changer_res) {
+    original.changer_res = new AutochangerResource();
+    original.changer_res->resource_name_
+        = strdup((original.resource_name_ + std::string("Changer")).c_str());
+    original.changer_res->changer_name = strdup("dev/null");
+    original.changer_res->changer_command = strdup("");
+    original.changer_res->device_resources = new alist<DeviceResource*>();
+    original.changer_res->device_resources->append(device0);
+    original.changer_res->device_resources->append(std::addressof(original));
+    original.changer_res->rcode_ = R_AUTOCHANGER;
+    original.changer_res->rcode_str_ = "Autochanger";
+    original.changer_res->refcnt_ = 1;
 
-  /* create the copied devices */
-  for (uint32_t i = 0; i < count; i++) {
-    DeviceResource* copied_device_resource
-        = new DeviceResource(multiplied_device_resource);
+    if (!config.GetResWithName(R_AUTOCHANGER,
+                               original.changer_res->resource_name_)) {
+      config.AppendToResourcesChain(original.changer_res, R_AUTOCHANGER);
+    } else {
+      Emsg0(M_WARNING, 0,
+            "Device resource \"%s\" will not implicitly create an autochanger "
+            "resource \"%s\" since an autochanger resource with that name "
+            "already exists.\n",
+            original.resource_name_, original.changer_res->resource_name_);
+    }
+  } else {
+    Dmsg0(0,
+          "Device resource \"%s\" will not implicitly create an autochanger "
+          "resource since it is already used by autochanger resource \"%s\".\n",
+          original.resource_name_, original.changer_res->resource_name_);
+  }
 
-    /* append 0002, 0003, ... */
-    copied_device_resource->CreateAndAssignSerialNumber(i + 2);
-
-    copied_device_resource->multiplied_device_resource
-        = std::addressof(multiplied_device_resource);
-    copied_device_resource->count = 0;
-
-    my_config->AppendToResourcesChain(copied_device_resource,
-                                      copied_device_resource->rcode_);
-
-    if (copied_device_resource->changer_res) {
-      if (copied_device_resource->changer_res->device_resources) {
-        copied_device_resource->changer_res->device_resources->append(
-            copied_device_resource);
-      }
+  // create devices 2 - count
+  for (uint32_t i = 2; i <= original.count; i++) {
+    std::string serial_number = std::string("0000") + std::to_string(i);
+    std::string copy_name = original.resource_name_
+                            + serial_number.substr(serial_number.size() - 4);
+    if (!config.GetResWithName(R_DEVICE, copy_name.c_str())) {
+      DeviceResource* copy = CopyDevice(original, i);
+      config.AppendToResourcesChain(copy, copy->rcode_);
+    } else {
+      Dmsg0(0,
+            "Device resource \"%s\" will not implicitly create device \"%s\" "
+            "resource since there already exists a device resource with that "
+            "name.\n",
+            original.resource_name_, copy_name.c_str());
     }
   }
+  // create device 1
+  original.CreateAndAssignSerialNumber(1);
+  original.multiplied_device_resource = std::addressof(original);
 }
 
 static void MultiplyConfiguredDevices(ConfigurationParser& config)
@@ -471,7 +521,7 @@ static void MultiplyConfiguredDevices(ConfigurationParser& config)
   BareosResource* p = nullptr;
   while ((p = config.GetNextRes(R_DEVICE, p))) {
     DeviceResource& d = dynamic_cast<DeviceResource&>(*p);
-    if (d.count > 1) { MultiplyDevice(d); }
+    if (d.count > 1) { MultiplyDevice(d, config); }
   }
 }
 
--- a/core/src/stored/autochanger_resource.h
+++ b/core/src/stored/autochanger_resource.h
@@ -3,7 +3,7 @@
 
    Copyright (C) 2000-2011 Free Software Foundation Europe e.V.
    Copyright (C) 2011-2012 Planets Communications B.V.
-   Copyright (C) 2019-2021 Bareos GmbH & Co. KG
+   Copyright (C) 2019-2025 Bareos GmbH & Co. KG
 
    This program is Free Software; you can redistribute it and/or
    modify it under the terms of version three of the GNU Affero General Public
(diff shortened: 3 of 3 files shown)


florian-at-bareos:
still requires documentation

docs/manuals/source/Configuration/StorageDaemon.rst
```diff
@@ -254,9 +254,15 @@ Multiplied Device
...
-When the configuration is loaded the |bareosSD| will then automatically multiply this device :config:option:`sd/device/Count` times. The number of multiplied devices includes the original Device.
+Note that this option only has an effect if the assigned value is greater than 1.
 
-A number "0001" will be appended to name of the initial Device. All other multiplied Devices have increasing numbers "0002", "0003", accordingly. In the example below the name of the multiplied devices will be "MultiFileStorage0001", "MultiFileStorage0002", and so on.
+When the configuration is loaded, the |bareosSD| will then automatically multiply this device :config:option:`sd/device/Count` + 1 times adding suffixes to the names starting from "0000". 
+
+The multiplied device with the suffix "0000" serves a special purpose, it is implicitly assigned :config:option:`sd/device/Autoselect` to "no".
+All other multiplied devices are an exact copy of the original device.
+
+Additionally, specifying :config:option:`sd/device/Count` also implicitly creates an autochanger with the same name as the original device with all devices listet in that autochanger.
```
arogge, line 264:
```suggestion
Additionally, specifying :config:option:`sd/device/Count` also implicitly creates an autochanger with the same name as the original device with all devices listed in that autochanger.
```

arogge (changes requested):
I had a quick first look. I found a typo in the docs and the systemtest does not seem to work correctly yet.

bruno-at-bareos (changes requested):
Generally good, work quite well and being helpful.
Please checks the remarks.

core/src/stored/stored_conf.cc
```diff
@@ -432,37 +434,74 @@ static void ParseConfigCb(LEX* lc,
...
+
+    if (!config.GetResWithName(R_AUTOCHANGER,
+                               original.changer_res->resource_name_)) {
+      config.AppendToResourcesChain(original.changer_res, R_AUTOCHANGER);
+    } else {
+      Emsg0(M_WARNING, 0,
+            "Device resource \"%s\" will not implicitly create an autochanger "
+            "resource \"%s\" since an autochanger resource with that name "
+            "already exists.\n",
+            original.resource_name_, original.changer_res->resource_name_);
```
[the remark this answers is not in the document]
florian-at-bareos: I see your point! When discussing this with Andreas and Philipp we decided to do it like this because we want people to get rid of their manual defined vchanger and use the implicitly created one.

core/src/stored/stored_conf.cc
```diff
@@ -432,37 +434,74 @@ static void ParseConfigCb(LEX* lc,
...
+        config.GetResWithName(R_DEVICE, device_name.c_str()));
+    if (!device) {
+      device = CopyDevice(original, i);
+      if (i == 0) { device->autoselect = false; }
+      config.AppendToResourcesChain(device, R_DEVICE);
+    } else {
+      Emsg0(M_WARNING, 0,
+            "Device resource \"%s\" will not implicitly create device \"%s\" "
+            "resource since there already exists a device resource with that "
+            "name.\n",
```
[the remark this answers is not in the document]
arogge: Should be `M_INFO` instead of `M_WARNING` for all of these.
Also the release notes should contain an information on how/why we changed this and how to migrate to the new schema.

florian-at-bareos:
> maybe it is time to split this testrunner to separate start clean default

You think that is necessary?

bruno-at-bareos (changes requested):
Still to be done: 
- In the documentation, we must also rephrase the :config:option:`sd/device/Count` option, some of its properties have changed.
- adapt the changelog

systemtests/tests/multiplied-device/testrunner
```diff
@@ -95,46 +101,57 @@ check_two_logs
...
 #make sure all volumes are purged
-grep "There are no more Jobs associated with Volume \"Full-0001\". Marking it purged." "${tmp}"/log3.out >/dev/null 2>&1 &&
-grep "There are no more Jobs associated with Volume \"Full-0002\". Marking it purged." "${tmp}"/log3.out >/dev/null 2>&1 &&
-grep "There are no more Jobs associated with Volume \"Full-0003\". Marking it purged." "${tmp}"/log3.out >/dev/null 2>&1
-if test $? -ne 0 ; then
-  echo "The Purge command failed for a MultiFileStorage."
-  estat=1;
-fi
-
+expect_grep "There are no more Jobs associated with Volume \"Full-0001\". Marking it purged." \
```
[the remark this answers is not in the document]
florian-at-bareos: Good catch! I will specify the volume the purge failed for in the error message

bruno-at-bareos (approved):
All's good for me, I'll give an approval.
@arogge would you mind to have a recheck for c++ code ?

core/src/stored/stored_conf.cc
```diff
@@ -427,37 +429,74 @@ static void ParseConfigCb(LEX* lc,
   }
 }
 
-static void MultiplyDevice(DeviceResource& multiplied_device_resource)
+static DeviceResource* CopyDevice(DeviceResource& original, int serial_number)
```
sebsura, line 432: serial_number is just used to create the name, that we already created before.

Currently now have two different locations where the serial_number -> name conversion occurs, that have to be kept in sync, but nowhere is this stated.

Why does CopyDevice not just take a `new_name` parameter ?  This way we only have to do the conversion once, while also making clearer what CopyDevice actually does!

core/src/stored/stored_conf.cc
```diff
@@ -207,7 +209,7 @@ static const ResourceItem dev_items[] = {
...
   { "EofOnErrorIsEot", CFG_TYPE_BOOL, ITEM(res_dev, eof_on_error_is_eot), {config::IntroducedIn{18, 2, 4}, config::Description{"If Yes, Bareos will treat any read error at an end-of-file mark as end-of-tape. You should only set this option if your tape-drive fails to detect end-of-tape while reading."}}},
-  { "Count", CFG_TYPE_PINT32, ITEM(res_dev, count), {config::DefaultValue{"1"}, config::Description{"If Count is set to (1 < Count < 10000), this resource will be multiplied Count times. The names of multiplied resources will have a serial number (0001, 0002, ...) attached. If set to 1 only this single resource will be used and its name will not be altered."}}},
+  { "Count", CFG_TYPE_PINT32, ITEM(res_dev, count), {config::DefaultValue{"1"}, config::Description{"If Count is set to (1 < Count < 10000), duplicated devices will be created post-fixed with serial numbers 0000 up to Count. The 0000 device is automatically assigned 'Autoselect=No'. Additionally, an autochanger resource is created with the name of the device the Count directive is specified for. The duplicated devices will be assigned to this autochanger unless they are used in another autochanger already."}}},
```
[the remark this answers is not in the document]
sebsura: is 1 < Count < 10000 actually checked anywhere ?

core/src/stored/stored_conf.cc
```diff
@@ -466,7 +505,7 @@ static void MultiplyConfiguredDevices(ConfigurationParser& config)
   BareosResource* p = nullptr;
   while ((p = config.GetNextRes(R_DEVICE, p))) {
     DeviceResource& d = dynamic_cast<DeviceResource&>(*p);
-    if (d.count > 1) { MultiplyDevice(d); }
+    if (d.count > 1) { MultiplyDevice(d, config); }
```
sebsura, line 508: Here would be a good place to check if count (if set at all) is > 1 and < 10000.

core/src/stored/stored_conf.cc
```diff
@@ -427,37 +429,74 @@ static void ParseConfigCb(LEX* lc,
...
-    my_config->AppendToResourcesChain(copied_device_resource,
-                                      copied_device_resource->rcode_);
+  DeviceResource* copy = new DeviceResource(original);
+  copy->CreateAndAssignSerialNumber(serial_number);
+  copy->multiplied_device_resource = std::addressof(original);
+  copy->count = 0;
+  copy->refcnt_ = 1;
+  return copy;
+}
+static void MultiplyDevice(DeviceResource& original,
```
sebsura, line 441: Previously the original device resource became part of the multiplied devices.  This seems to be not the case anymore.  So now i get an extra device that i didnt want.  Is this really what we want ?

```
* status storage
...
Device status:
Autochanger "FileStorage" with devices:
   "FileStorage0000" (storage)
   "FileStorage0001" (storage)
   "FileStorage0002" (storage)
   "FileStorage0003" (storage)

Device "FileStorage" (storage) is not open.
==

Device "FileStorage0000" (storage) is not open.
==

Device "FileStorage0001" (storage) is not open.
==

Device "FileStorage0002" (storage) is not open.
==

Device "FileStorage0003" (storage) is not open.
==
```

sebsura (changes requested):
This needs some more time in the oven.  I could very reliably create a crash:

Add an implicit autochanger with 4 devices.  Set MaxConcurrentJobs to 10 everywhere.  Start 10 concurrent jobs.  Then the sd crashes.

This is because the sd tries to use the original device resource, which is not setup correctly (it has changer_res set, but not changer_command; see OpenDevice).

core/src/stored/stored_conf.cc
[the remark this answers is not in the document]
florian-at-bareos: The Autochanger is always created.

core/src/stored/stored_conf.cc
florian-at-bareos: I think it makes sense to not override the "original" device.
- Copied devices reference the original device which makes it weird if the original device is being transformed to a copied device
- Overriding the original device makes it complicated to `PrintConfig` so we need to store the `DeviceResource::multiplied_device_resource_base_name`
- Is a little difficult to implement because my changes allow for copy "specializations", which means that i can for example explicitly define "device0002" if i want it to be special in some sense. This means that before creating a copy we check whether the device with that name is not already defined. Overriding the original device would only be possible if we actually need to create a copy, so it is not clear whether the original device is "device0000", "device0001"... 

What I would do instead is maybe hide it in the "status storage" view? It can't be accidentally used anyway since we create an autochanger of the same name.

core/src/stored/stored_conf.cc
```diff
@@ -427,37 +429,74 @@ static void ParseConfigCb(LEX* lc,
...
-        = std::addressof(multiplied_device_resource);
-    copied_device_resource->count = 0;
-
-    my_config->AppendToResourcesChain(copied_device_resource,
-                                      copied_device_resource->rcode_);
+  DeviceResource* copy = new DeviceResource(original);
+  copy->CreateAndAssignSerialNumber(serial_number);
+  copy->multiplied_device_resource = std::addressof(original);
+  copy->count = 0;
+  copy->refcnt_ = 1;
```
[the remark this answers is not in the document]
florian-at-bareos: When i developed this these devices where destroyed and this solved it. Now testing it it seems unnecessary

sebsura (changes requested):
We should definitely have a systemtest that actually uses this feature.  I imagine the test could check
1) that the maximum number of running jobs equals the count specified in the resource
2) that one device was created with autoselect = no
...

This test would also make it easier to reproduce bugs as we would not have to set up the environment from scratch every time :)



Im still getting messages like this when running multiple jobs:
>  2025-06-05 07:51:23 bareos-sd JobId 5: Using Device "FileStorage" (storage) to write.

So im guessing the issue with the extra device is still not fixed.  Could you double check that ?

Interestingly the device does not turn up in `status sd`.

sebsura:
```
*show storage
Storage {
  Name = "File"
  Address = "localhost"
  Port = 30363
  Password = "[md5]519001602a0e85b0fe7dbead5d95d710"
  Device = "FileStorage"
  MediaType = "File"
  AutoChanger = Yes
  MaximumConcurrentJobs = 20
}

$ cat etc/bareos/bareos-sd.d/device/FileStorage.conf 
Device {
  Name = FileStorage
  Media Type = File
  Device Type = File
  Archive Device = storage
  LabelMedia = yes;                   # lets Bareos label unlabeled media
  Random Access = yes;
  AutomaticMount = yes;               # when device opened, read it
  RemovableMedia = no;
  AlwaysOpen = yes;
  Description = "File device. A connecting Director must have the same Name and MediaType."
  Count = 4
}
```

if you want to try and reproduce this.

systemtests/tests/multiplied-device/testrunner-count
```diff
@@ -0,0 +1,184 @@
...
+. ./environment
+
+JobName=backup-bareos-fd
+#shellcheck source=../scripts/functions
+. "${rscripts}"/functions
+"${rscripts}"/cleanup
+"${rscripts}"/setup
+
+# Fill ${BackupDirectory} with data.
+setup_data
```
sebsura: this should happen in test-setup

systemtests/tests/multiplied-device/testrunner-count
```diff
@@ -0,0 +1,184 @@
...
+truncate volstatus=Purged drive=0 yes
+purge volume=Full-0002 yes
+truncate volstatus=Purged drive=1 yes
+purge volume=Full-0003 yes
+truncate volstatus=Purged drive=2 yes
+messages
+quit
+END_OF_DATA
+
+run_bareos "$@"
```
sebsura: run_bconsole; test-setup should start the daemons.

systemtests/tests/multiplied-device/testrunner-explicit-autochanger
```diff
@@ -0,0 +1,184 @@
...
+
+#
+# This systemtest validates that devices, multiplied using the 'Count' directive,
+# work and can be used as any other device.
+# To validate the devices behaviour, we perform the following actions:
+#  - backups
+#  - restores
+#  - purges
+#  - truncates
+#
```
sebsura, line 14: The description does not match the name.

systemtests/tests/multiplied-device/testrunner-implicit-autochanger
```diff
@@ -0,0 +1,67 @@
...
+set -u
+
+#
+# This systemtest validates that devices, multiplied using the 'Count' directive,
+# work and can be used as any other device.
+# To validate the devices behaviour, we perform the following actions:
+#  - backups
+#  - restores
+#  - purges
+#  - truncates
```
sebsura, line 33: i only see backups here.  Did you accidentally switch up the test names ?

systemtests/tests/multiplied-device/etc/bareos/bareos-sd.d/device/ImplicitAutochanger.conf
```diff
@@ -0,0 +1,25 @@
+# Autochanger {
+#   Name = ImplicitAutchanger
+
+#   Changer Device = /dev/null
+
+#   # list here only the basename of the multidevice
+#   Device = ImplicitAutchangerDevice
+
+#   Changer Command = ""
```
sebsura: this comment seems unnecessary

systemtests/tests/multiplied-device/etc/bareos/bareos-sd.d/device/ImplicitAutochanger.conf
```diff
@@ -0,0 +1,25 @@
+# Autochanger {
+#   Name = ImplicitAutchanger
+
+#   Changer Device = /dev/null
+
+#   # list here only the basename of the multidevice
```
sebsura: base name = resource name ?

core/src/stored/autochanger_resource.cc
```diff
@@ -53,17 +66,20 @@ bool AutochangerResource::PrintConfig(OutputFormatterResource& send,
                                       bool hide_sensitive_data,
                                       bool verbose)
 {
+  if (implicitly_created_) { return false; }
```
sebsura: ill have to look into this, but i would think this should return true.  I think false here means that an error occurred.

core/src/stored/autochanger_resource.cc
```diff
@@ -22,19 +22,32 @@
...
 
 AutochangerResource::AutochangerResource()
-    : BareosResource()
-    , device_resources(nullptr)
-    , changer_name(nullptr)
-    , changer_command(nullptr)
 {
-  return;
+  rcode_ = R_AUTOCHANGER;
+  rcode_str_ = "Autochanger";
```
sebsura, line 37: i think this can be moved above the {

sebsura: This sadly has to be done this way because those are part of the base class.

arogge (approved):
Looks good to me!

PR #2207, opened 2025-03-12
stored: add new volume status 'Unlabeled'

### LINKED ISSUE #1804: Director should not create volume record if storage daemon fails to label the volume
### Bareos component version

current / 23.0.2 and probably every older version

### Steps to reproduce

1. set up an SD with File backend pointing to a non-existent directory
2. autolabel a volume on that SD

### Expected results

Getting an error message from the Storage and no Volume record in the media table.

### Actual results

Error message from the Storage is there, but the Director already created a volume record in the catalog for a volume that could not be labeled.

### Environment

_No response_

### Relevant log output

_No response_

### Relevant traces output

_No response_

### Anything else?

https://bugs.bareos.org/view.php?id=1598

florian-at-bareos:
This status is the initial status when a new volume is created in the catalog and is changed if the storage daemon (successfully) wrote on the volume.

Fixes #1804

Currently:
1. the sd asks for a new volume to append to
2. the director creates a new record in the catalog with that new volume with "Append" status and sends the information to the sd
3. the sd can't label the volume, errors out
-> the volume record with "Append" status persists in the catalog even if no volume with that name was ever created by the sd

With this PR:
1. the sd asks for a new volume to append to
2. the director creates a new record in the catalog with that new volume with "Unlabeled" status and sends the information to the sd
3. the sd can't label the volume, errors out
-> the volume record with "Unlabeled" status persists in the catalog and is selected again if the sd asks for a volume to append to until labeling works

If the sd successfully wrote on the "Unlabeled" volume, its status is updated to "Append", "Full" or whatever the sd decides.

**Note:**
Manually updating an existing volume record in the catalog to the status "Unlabeled" is not dangerous, no backup data is ever overwritten!
When the sd wants to send volume updates to the director, and the status is "Unlabeled" but has written bytes, the status is set to "Append".

- [x] Update docs accordingly

#### Please check

##### General

##### Source code quality


florian-at-bareos:
Currently, when manually labeling a volume its status is set to unlabeled

core/src/cats/ddl/creates/postgresql.sql
```diff
@@ -171,7 +171,7 @@ CREATE TABLE Media
     VolWrites         BIGINT      DEFAULT 0,
     VolCapacityBytes  BIGINT      DEFAULT 0,
     VolStatus         TEXT        NOT NULL
-        CHECK (VolStatus IN ('Full', 'Archive', 'Append',
```
arogge, line 174: we need migrations to do that on upgrade and the schema version needs to be updated when schema changes.

core/src/cats/ddl/creates/postgresql.sql
```diff
@@ -304,7 +304,7 @@ CREATE TABLE LocationLog (
     MediaId           INTEGER DEFAULT 0,
     LocationId        INTEGER DEFAULT 0,
     NewVolStatus              TEXT NOT NULL
-        CHECK (NewVolStatus IN ('Full', 'Archive', 'Append',
```
arogge, line 307: see above

arogge (changes requested):
* You'll need to upgrade the database correctly. This means we need to introduce a new database version and create a migration script for that.
* As far as I can tell an unlabeled volume was definitely never written to (as the label would have been written first), so neither the purge nor the prune operation makes sense on such a volume.
* When selecting a volume, I think we should prefer appendable volumes over unlabeled ones

florian-at-bareos:
- I don't know how to write such a migration script and how much effort it is
- You're right, that is a little awkward but it was necessary to integrate with the existing behaviour because otherwise "purge" and "prune" commands would just fail for a newly created volume which it doesn't currently. I'll think about how one can make this more intuitive.
- that's the behaviour of the current changes

maybe we can discuss sometime how to pursue with this.

arogge:
> I don't know how to write such a migration script and how much effort it is

The migration scripts are in `core/src/cats/ddl/updates`.
We have a schema version (usually that's the Bareos target version prefixed by a `2` and suffixed with a `0`, i.e. `2250` for Bareos 25), this must be set correctly in the create-script, too.

For the upgrade script you create a new file (probably based on an existing one) named `postgresql.<old_db_version>_<new_db_version>.sql` that contains the commands required to transition the old database schema into the new one. Everything should be wrapped into a transaction (i.e. `begin` and `commit`) and the version update should be the last things before the transaction commit.

When changing the database version, you also have to change the `#define BDB_VERSION` in `core/src/cats/cats.h` so Bareos knows what database version to expect.

arogge:
> You're right, that is a little awkward but it was necessary to integrate with the existing behaviour because otherwise "purge" and "prune" commands would just fail for a newly created volume which it doesn't currently. I'll think about how one can make this more intuitive.

That's actually a good point. I think it doesn't make much sense to support purge/prune on these volumes, but you're right that it should probably not fail either.
Maybe it makes sense to just ignore volumes (maybe with a message) that are in status `Unlabeled` for prune/purge operations?

core/src/dird/next_vol.cc
```diff
@@ -385,8 +392,8 @@ bool GetScratchVolume(JobControlRecord* jcr,
       smr.StorageId = mr->StorageId; /* want only Scratch Volumes in changer */
     }
 
-    bstrncpy(smr.VolStatus, "Append",
-             sizeof(smr.VolStatus)); /* want only appendable volumes */
+    bstrncpy(smr.VolStatus, "Unlabeled",
+             sizeof(smr.VolStatus)); /* want only unlabeled volumes */
```
arogge, line 396: I don't think this is correct. When taking a volume from Scratch it should be in status `Append`, we should probably not try `Unlabeled` volumes from Scratch.
If we wanted this to work we'd still have to prefer appendable volumes here.

arogge (changes requested):

core/src/dird/next_vol.cc
florian-at-bareos: Agree!

core/src/dird/ua_label.cc
```diff
@@ -263,6 +267,7 @@ bool SendLabelRequest(UaContext* ua,
   }
 
   if (retval) {
+    bstrncpy(mr->VolStatus, "Append", sizeof(mr->VolStatus));
```
sebsura, line 270: Why is this done twice ? (its also done in update_database() but only sometimes)

core/src/dird/ua_purge.cc
```diff
@@ -3,7 +3,7 @@
 
    Copyright (C) 2002-2012 Free Software Foundation Europe e.V.
    Copyright (C) 2011-2016 Planets Communications B.V.
-   Copyright (C) 2013-2024 Bareos GmbH & Co. KG
+   Copyright (C) 2013-2025 Bareos GmbH & Co. KG
```
sebsura, line 6: why was this changed ?

sebsura (changes requested):

core/src/dird/next_vol.cc
```diff
@@ -88,11 +88,17 @@ int FindNextVolumeForAppend(JobControlRecord* jcr,
...
+    //  1. Look for volume with "Append" or "Unlabeled" status.
     SetStorageidInMr(store, mr);
 
     bstrncpy(mr->VolStatus, "Append", sizeof(mr->VolStatus));
     ok = jcr->db->FindNextVolume(jcr, index, InChanger, mr, unwanted_volumes);
+    if (!ok) {
+      bstrncpy(mr->VolStatus, "Unlabeled", sizeof(mr->VolStatus));
+      ok = jcr->db->FindNextVolume(jcr, index, InChanger, mr, unwanted_volumes);
+    }
+    if (!ok) { bstrncpy(mr->VolStatus, "Append", sizeof(mr->VolStatus)); }
```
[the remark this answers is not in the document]
florian-at-bareos: Do you think it makes more sense to prioritize unlabeled volumes over appendable volumes? I would say the over way round makes more sense. If a volume that is appendable and usable already exists, why should we label a new one?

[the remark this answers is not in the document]
florian-at-bareos: I am not 100% whether it is necessary to reset to "Append" from a functional point of view, but i would definitely do it because the method `FindNextVolumeForAppend` takes a media record as a parameter which is changed to the volume that was found and I think it is logical to not change the media record if no volume is found.

core/src/stored/askdir.cc
```diff
@@ -302,6 +302,11 @@ bool StorageDaemonDeviceControlRecord::DirUpdateVolumeInfo(
   if (label == is_labeloperation::True) {
     bstrncpy(vol->VolCatStatus, "Append", sizeof(vol->VolCatStatus));
   }
+  // Manually labeled volume was written on
+  else if (bstrncmp(vol->VolCatStatus, "Unlabeled", sizeof(vol->VolCatStatus))
+           && vol->VolCatBytes > 0) {
+    bstrncpy(vol->VolCatStatus, "Append", sizeof(vol->VolCatStatus));
+  }
```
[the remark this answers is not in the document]
florian-at-bareos: This should not happen.

You can manually update the volume status with "update volume" > Volume Status but I removed the "Unlabeled" option.

I'll add a comment.

core/src/dird/next_vol.cc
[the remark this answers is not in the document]
sebsura: You are right.  Prioritizing an already labeled one is correct, and then you definitely need to reset to "Append" again.

docs/manuals/source/DeveloperGuide/catalog.rst
```diff
@@ -406,7 +406,7 @@ created for each of the NumVols specified in the Pool resource record.
 | VolCapacityBytes | bigint    | Capacity estimate for this volume       |
 +------------------+-----------+-----------------------------------------+
 | VolStatus        | text      | | Status of media:                      |
-|                  |           | | Full, Archive, Append,                |
+|                  |           | | Unlabeled, Full, Archive, Append,     |
 |                  |           |   Recycle, Read-Only, Disabled, Error,  |
 |                  |           |   Busy                                  |
```
sebsura: This creates a line that is slightly too long.
```
| VolStatus        | text      | | Status of media:                      |
|                  |           | | Unlabeled, Full, Archive, Append,     |
|                  |           | | Recycle, Read-Only, Disabled, Error,  |
|                  |           |   Busy                                  |
```
this might fix it.

docs/manuals/source/DeveloperGuide/catalog.rst
```diff
@@ -645,26 +645,26 @@ LocationLog
...
++--------------+-----------+-----------------------------------------------+
+| Comment      | text      | Time/date log record created                  |
++--------------+-----------+-----------------------------------------------+
+| MediaId      | integer   | Points to Media record                        |
++--------------+-----------+-----------------------------------------------+
+| LocationId   | integer   | Points to Location record                     |
++--------------+-----------+-----------------------------------------------+
+| NewVolStatus | integer   | enum: Unlabeled, Full, Archive, Append,       |
+|              |           | Recycle, Purged, Read-only, Disabled, Error,  |
+|              |           | Busy, Used, Cleaning                          |
```
sebsura: The same applies here:  i get a scroll bar for the table.  You can just split it up like we did above.

core/src/stored/dir_cmd.cc
```diff
@@ -669,7 +669,7 @@ static bool DoLabel(JobControlRecord* jcr, bool relabel)
   FreeMemory(poolname);
   FreeMemory(mediatype);
   dir->signal(BNET_EOD);
-  return true;
+  return ok;
```
sebsura, line 680: Why was this changed ?  The return value of a command indicates if the _console connection should be killed_, not whether the command was successfull.
See `HandleDirectorConnection` (variable: quit).

sebsura: It probably makes sense to just update `LabelCmd`/`RelabelCmd` instead, so they always return true.

arogge:
@sebsura I squashed some of the commits. Do you want to take another look at this?

PR #2222, opened 2025-03-21
core: refactor config parser; fix ktls configuration; fix crashes/ub

sebsura:

This pr fixes a lot of bugs in our configuration & config parser.  It also changes how we define our configuration to make sure that these mistakes cannot happen anymore.  The PR also contains the following fixes:

- make the ktls configuration useful
  The ktls configuration of each daemon will now determine whether that particular daemon tries to use it.
- remove odr violations from the sd (the same file was getting compiled twice)
- fix out of bounds reads in the fd
- fix crashes related to reload
  Now each configuration gets kept alive by their predecessors, which means that jobs which incorrectly try to access out of date configurations, will not create use-after-free situations.

#### Please check

##### General

##### Source code quality


core/src/lib/parse_conf.h
```diff
@@ -419,58 +419,62 @@ class ConfigurationParser {
...
+      free_res(configuration_resources_[i], i);
     }
-    free(configuration_resources_);
-    configuration_resources_ = nullptr;
   }
+
   void SetTimestampToNow() { timestamp_ = std::chrono::system_clock::now(); }
   std::string TimeStampAsString() { return TPAsString(timestamp_); }
+
+  void SetNext(std::shared_ptr<ConfigResourcesContainer> next) { next_ = next; }
```
florian-at-bareos, line 469:
```c++
next_ = std::move(next);
```

core/src/lib/address_conf.h
```diff
@@ -58,16 +58,14 @@ class IPADDR {
...
-    struct sockaddr_in6 dontuse6;
-  } saddrbuf;
-  struct sockaddr* saddr = nullptr;
-  struct sockaddr_in* saddr4 = nullptr;
-  struct sockaddr_in6* saddr6 = nullptr;
+    sockaddr addr;
+    sockaddr_in addr_in;
+    sockaddr_in6 addr_in6;
+    sockaddr_storage addr_storage = {};
+  };
```
florian-at-bareos, line 67: This is already way better than it was before!

Maybe we can reduce this to
```
std::variant<sockaddr, sockaddr_in, sockaddr_in6> addr_;
```

core/src/lib/resource_item.h
```diff
@@ -26,41 +26,210 @@
...
+  int type{};
+  std::size_t offset{};
+  BareosResource** allocated_resource{};
+  int32_t code{}; /* Item code/additional info */
+  std::vector<std::string> aliases{};
+  bool required{};
+  bool deprecated{};
+  bool platform_specific{};
+  bool no_equal{};
+  const char* default_value{}; /* Default value */
```
florian-at-bareos, line 211: If it's not too much effort, I would love to see `default_value` to be of type `std::optional<std::string>`.

florian-at-bareos (changes requested):
Great work!
I think there are some really nice improvements to the config parser code.

core/src/lib/resource_item.h
sebsura: Cannot use std::strings (or anything that allocates) because the plan is to make these constexpr (so that you can use `static_assert` to verify some properties) and you cannot keep allocated memory from compile time execution to runtime execution.

florian-at-bareos (approved):
Great work! I especially like the `ResourceItem` config system you wrote!

If you want you can replace the `const char*` with `std::optional<std::string_view>` (see above discussion) but since this was a `const char* ` before we can also just leave it as is for now.

In case you haven't noticed:
Some of your commit titles are longer than 50 characters.

- [resource-item: add new way to specify resource items](https://github.com/bareos/bareos/pull/2222/commits/e38af28291d871df97118f021dd1518a96be2e04)
- [parse-conf: introduce configresourcecontainer chain](https://github.com/bareos/bareos/pull/2222/commits/85ad08c1afeb7ae7d35434edfc4f28388e281344)

core/src/lib/resource_item.h
sebsura: Yeah that would be a good alternative.  Currently the string needs to be zero terminated :(, but i can look into removing this requirement in the future, and changing this to be a string view.

sebsura:
> maybe rename to "refactor config parser" since there are more changes than in parse_conf?

good idea!

PR #2240, opened 2025-04-03
dplcompat: fix unsupported size-suffixes on chunksize

arogge:
Add support for size-suffixes to chunksize parameter to be more compatible with the droplet backend. 
This will now also detect trailing junk after numbers and changes debugging at crud_storage's trace level (130) will now log the output of the program the stat operation is run. This also slightly improves some other trace information.

#### Please check

##### General

##### Source code quality

### DIFF
--- a/core/src/stored/backends/crud_storage.cc
+++ b/core/src/stored/backends/crud_storage.cc
@@ -1,7 +1,7 @@
 /*
    BAREOS® - Backup Archiving REcovery Open Sourced
 
-   Copyright (C) 2024-2024 Bareos GmbH & Co. KG
+   Copyright (C) 2024-2025 Bareos GmbH & Co. KG
 
    This program is Free Software; you can redistribute it and/or
    modify it under the terms of version three of the GNU Affero General Public
@@ -30,6 +30,9 @@
 #if defined(HAVE_WIN32)
 #  include <shlwapi.h>  // for PathIsRelativeA()
 #endif
+
+// we want the real sscanf() and not bsscanf()
+#undef sscanf
 
 namespace {
 constexpr int debug_info = 110;
@@ -231,22 +234,30 @@ tl::expected<void, std::string> CrudStorage::test_connection()
 auto CrudStorage::stat(std::string_view obj_name, std::string_view obj_part)
     -> tl::expected<Stat, std::string>
 {
-  Dmsg1(debug_trace, "stat %s called\n", obj_name.data());
+  Dmsg1(debug_trace, "stat %s/%s called\n", obj_name.data(), obj_part.data());
   std::string cmdline
       = fmt::format("\"{}\" stat \"{}\" \"{}\"", m_program, obj_name, obj_part);
   auto bph{
       BPipeHandle::create(cmdline.c_str(), m_program_timeout, "r", m_env_vars)};
   if (!bph) { return tl::unexpected(bph.error()); }
-  auto rfh = bph->getReadFd();
-  Stat stat;
-  if (int n = fscanf(rfh, "%zu\n", &stat.size); n != 1) {
-    return tl::unexpected(
-        fmt::format("could not parse data returned by {}\n", cmdline));
-  }
-  if (auto ret = bph->close(); ret != 0) {
+  auto output = bph->getOutput();
+  auto ret = bph->close();
+  Dmsg1(debug_trace,
+        "stat returned %d\n"
+        "== Output ==\n"
+        "%s"
+        "============\n",
+        ret, output.c_str());
+  if (ret != 0) {
     Dmsg1(debug_info, "stat returned %d\n", ret);
     return tl::unexpected(
         fmt::format("Running \"{}\" returned {}\n", cmdline, ret));
+  }
+
+  Stat stat;
+  if (int n = sscanf(output.c_str(), "%zu\n", &stat.size); n != 1) {
+    return tl::unexpected(
+        fmt::format("could not parse data returned by {}\n", cmdline));
   }
   Dmsg1(debug_trace, "stat returns %zu\n", stat.size);
   return stat;
--- a/core/src/stored/backends/crud_storage.cc
+++ b/core/src/stored/backends/crud_storage.cc
@@ -18,6 +18,9 @@
    Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
    02110-1301, USA.
 */
+
+#define FMT_ENFORCE_COMPILE_STRING
+
 #include "include/bareos.h"
 #include <fmt/format.h>
 #include "crud_storage.h"
@@ -30,9 +33,12 @@
 #if defined(HAVE_WIN32)
 #  include <shlwapi.h>  // for PathIsRelativeA()
 #endif
+#include "util.h"
 
 // we want the real sscanf() and not bsscanf()
 #undef sscanf
+
+namespace utl = backends::util;
 
 namespace {
 constexpr int debug_info = 110;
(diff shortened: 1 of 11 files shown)


core/src/stored/backends/crud_storage.cc
```diff
@@ -231,23 +234,31 @@ tl::expected<void, std::string> CrudStorage::test_connection()
 auto CrudStorage::stat(std::string_view obj_name, std::string_view obj_part)
     -> tl::expected<Stat, std::string>
 {
-  Dmsg1(debug_trace, "stat %s called\n", obj_name.data());
+  Dmsg1(debug_trace, "stat %s/%s called\n", obj_name.data(), obj_part.data());
```
sebsura, line 237: Not a fan of this tbh. Its too hard to misuse this function now.
```suggestion
  Dmsg1(debug_trace, "stat %.*s/%.*s called\n", saturating_cast<int>(obj_name.size()), obj_name.data(), saturating_cast<int>(obj_part.size()), obj_part.data());
```
where `saturating_cast<S>(U u)` returns
```
* S::max(), when u > (U)S::max()
* u, otherwise
```

core/src/stored/backends/crud_storage.cc
```diff
@@ -231,23 +234,31 @@ tl::expected<void, std::string> CrudStorage::test_connection()
...
     Dmsg1(debug_info, "stat returned %d\n", ret);
     return tl::unexpected(
         fmt::format("Running \"{}\" returned {}\n", cmdline, ret));
   }
+
+  Stat stat;
+  if (int n = sscanf(output.c_str(), "%zu\n", &stat.size); n != 1) {
+    return tl::unexpected(
+        fmt::format("could not parse data returned by {}\n", cmdline));
+  }
```
sebsura, line 261: Why are we using sscanf here instead of just using `strtoul(l)` ?  If you want to make sure that there is a `\n` at the end, you can also do this with `strtoul(l)`.

core/src/stored/backends/dplcompat_device.cc
```diff
@@ -277,7 +277,7 @@ bool DropletCompatibleDevice::ReadRemoteChunk(chunk_io_request* request)
 {
   const std::string_view obj_name{request->volname};
   const std::string obj_chunk = get_chunk_name(request);
-  Dmsg1(debug_trace, "Reading chunk %s\n", obj_name.data());
+  Dmsg1(debug_trace, "Reading chunk %s/%s\n", obj_name.data(), obj_chunk.data());
```
sebsura, line 280:
```suggestion
  Dmsg1(debug_trace, "Reading chunk %s/%s\n", request->volname, obj_chunk.c_str());
```

sebsura (changes requested):

core/src/stored/backends/dplcompat_device.cc
arogge: Hmm... maybe we instead want a `Dfmt(lvl, fmt_format_string, ...)` that internally uses fmt instead of sprintf?
Then we could just do `Dfmt(debug_trace, "Reading chunk {}/{}", obj_name, obj_chunk)` and it would be safe.

sebsura: Yeah that makes sense and is also what im doing in my plugins :)

core/src/stored/backends/crud_storage.cc
sebsura: I guess you use sscanf for the other items anyways, so this can stay this way anyways.

core/src/stored/backends/dplcompat_device.cc
arogge: I implemented the `Dfmt()` thingy. It *should* work for the general case and I need it in `crud_storage.cc`, too.
Do you think it makes sense to put it into lib as it is? What would add fmt as a dependency to libbareos, which we currently don't have.
Otherwise I'd just put it in `dplcompat_device.h`, but then it isn't reusable.

core/src/stored/backends/crud_storage.cc
```diff
@@ -369,55 +392,59 @@ tl::expected<gsl::span<char>, std::string> CrudStorage::download(
     if (bytes_read < read_size) {
       if (feof(rfh)) {
         return tl::unexpected(
-            fmt::format("unexpected EOF after reading {} of {} bytes while "
-                        "downloading {}/{}",
+            fmt::format(FMT_STRING("unexpected EOF after reading {} of {} "
```
sebsura, line 395: Im suprised that you never just created a function `format_err(...) -> tl::unexpected<std::string>`.

core/src/stored/backends/dplcompat_device.cc
```diff
@@ -54,36 +55,56 @@ bool is_chunk_name(const std::string& name)
...
+  return val;
+}
+
+unsigned long stoul_notrailing(const std::string& str)
+{
+  size_t pos;
+  unsigned long val = std::stoul(str, &pos);
+  if (!std::all_of(str.begin() + pos, str.end(), b_isjunkchar)) {
+    throw std::invalid_argument{"unparseable trailing characters"};
+  }
```
sebsura, line 79: Maybe we can add information here on where this character is ?

```C++
if (auto iter = std::find_if_not(...); iter != str.end()) {
  std::string errmsg = std::format("unexpected character '{0}' ({0:X}) at pos {1}", *iter, iter - str.begin());
  throw std::invalid_argument{errmsg};
}
```

Im not sure if this is better.  Let me know what you think!

core/src/stored/backends/dplcompat_device.cc
```diff
@@ -54,36 +55,56 @@ bool is_chunk_name(const std::string& name)
...
                                       const std::string& from)
 {
-  to = std::stoul(from);
+  to = stoul_notrailing(from);
 }
 
 template <> void convert_value<>(uint8_t& to, const std::string& from)
 {
-  to = gsl::narrow<uint8_t>(std::stoul(from));
+  to = gsl::narrow<uint8_t>(stoul_notrailing(from));
```
sebsura, line 102: Cool find.  I often needed something like `gsl::narrow`!

sebsura (changes requested):

core/src/stored/backends/util.h
```diff
@@ -42,6 +44,31 @@ using options = std::map<std::string, std::string, key_comparator>;
...
+};
+
+/* emit a fmt-formatted debug message */
+template <typename Fmt, typename... Args>
+void Dfmt(LevelAndLocation lal, Fmt fmt, Args&&... args)
+{
+  const auto formatted
+      = fmt::format(std::forward<Fmt>(fmt), std::forward<Args>(args)...);
+  d_msg(lal.loc.file_name(), lal.loc.line(), lal.level, "%s\n",
+        formatted.c_str());
```
arogge, line 69: this should still check the current debuglevel

core/src/stored/backends/dplcompat_device.cc
arogge: I'm not really in favour of providing a character offset here. Basically the user was providing a string of `key=val,...` and the only thing we can provide is the offset inside of `val`. I'm not sure this is really useful.

arogge: Nevertheless, we could add `str` to provide a bit more detail. As a result we would have something like `"unexpected character '{0}' ({0:X} at offset {1} in '{2}'`.
What do you think?

arogge: So what I implemented will result in the following when you set `iothreads=2 *` in your configuration:
```
bareos-sd: Fatal Error at backends/dplcompat_device.cc:196 because:
invalid argument '2 *' for option 'iothreads': unparseable character '*' (0x2a) at pos 2
```

sebsura: yes that looks good.

core/src/stored/backends/dplcompat_device.cc
```diff
@@ -92,31 +118,42 @@ template <> void convert_value<>(std::string& to, const std::string& from)
...
-  } catch (std::out_of_range& e) {
-    return tl::unexpected(
-        fmt::format("value '{}' for option '{}' is out of range: {}\n", value,
-                    key, e.what()));
-  } catch (gsl::narrowing_error& e) {
     return tl::unexpected(
-        fmt::format("value '{}' for option '{}' would be truncated: {}\n",
+        fmt::format(FMT_STRING("invalid argument '{}' for option '{}': {}\n"),
                     value, key, e.what()));
+  } catch (std::out_of_range& e) {
```
sebsura: This doesnt really matter, but feel free to 'fix' it anyways:
```suggestion
  } catch (const std::out_of_range& e) {
```
(same for the other exceptions).

PR #2273, opened 2025-05-13
plugins: fix error_string construction

florian-at-bareos:
This PR fixes a bug introduced with commit ce9f5bc, see #225.

PyUnicode_AsUTF8 can return NULL and doing std::string(NULL) will crash.
With the changes of this PR we explicitly do a null-pointer check before constructing a std::string.

Fixes bareos/internal#225: ce9f5bc introduces creation of std::string from potential nullptr

#### Please check

##### General

##### Source code quality

### DIFF
--- a/core/src/plugins/include/python_plugins_common.inc
+++ b/core/src/plugins/include/python_plugins_common.inc
@@ -1,7 +1,7 @@
 /*
    BAREOS® - Backup Archiving REcovery Open Sourced
 
-   Copyright (C) 2020-2024 Bareos GmbH & Co. KG
+   Copyright (C) 2020-2025 Bareos GmbH & Co. KG
 
    This program is Free Software; you can redistribute it and/or
    modify it under the terms of version three of the GNU Affero General Public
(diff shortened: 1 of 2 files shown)


sebsura (changes requested):
Looks good.  Just added a small nit to think about.

core/src/plugins/include/python_plugins_common.inc
```diff
@@ -116,14 +116,19 @@ static std::string GetStringFromPyErrorHandler()
     strRetval
         = PyObject_CallMethod(emptyString, (char*)"join", (char*)"O", tbList);
 
-    error_string = std::string(PyUnicode_AsUTF8(strRetval));
+    if (const char* err = PyUnicode_AsUTF8(strRetval)) {
```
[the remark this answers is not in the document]
florian-at-bareos: I can do that, however, PyObject* are used at hundreds of places so it will probably won't do much because if the python interpreter is broken we would get NULL at some place before probably.

core/src/plugins/include/python_plugins_common.inc
```diff
@@ -105,25 +105,37 @@ static std::string GetStringFromPyErrorHandler()
...
+        error_string = "Unable to retrieve empty-string.";
+      }
+    }
+    else {
+      error_string = "Calling method 'format_exception' failed.";
+    }
     Py_DECREF(tbList);
     Py_DECREF(emptyString);
     Py_DECREF(strRetval);
     Py_DECREF(tracebackModule);
```
sebsura, line 136:
```suggestion
    Py_XDECREF(tbList);
    Py_XDECREF(emptyString);
    Py_XDECREF(strRetval);
    Py_XDECREF(tracebackModule);
```
Py_DECREF assumes its argument to be not null.  If we dont know if its null or not, we need to use `XDECREF`.

sebsura (changes requested):

core/src/plugins/include/python_plugins_common.inc
florian-at-bareos: Good catch!

sebsura (changes requested):

florian-at-bareos:
If I interpret this correctly `type == NULL` should implies that `value == NULL` and `traceback == NULL`, see https://docs.python.org/3/c-api/exceptions.html#c.PyErr_Fetch.

sebsura:
That is true, but as we do not check whether traceback or value are null before calling the function, we can still pass a null `type` into the python function. 

Looking into the cpython doc i get the following (for `Py_BuildValue`):
```
O (object) [PyObject *]

    Pass a Python object untouched but create a new [strong reference](https://docs.python.org/3/glossary.html#term-strong-reference) to it (i.e. its reference count is incremented by one).
    If the object passed in is a NULL pointer, it is assumed that this was caused because the call producing the argument found an error and set an exception.
    Therefore, [Py_BuildValue()](https://docs.python.org/3/c-api/arg.html#c.Py_BuildValue) will return NULL but won’t raise an exception.
    If no exception has been raised yet, [SystemError](https://docs.python.org/3/library/exceptions.html#SystemError) is set.
```

so this isnt fatal, but still not great.  It looks like we should also be using `Py_Restore(...)` once we are dont with the result of `Py_Fetch(...)`, so its probably best to just leave this as is and rework it completely at a later point.
After all almost all of these  functions are deprecated in 3.12, so this will need to get replaced soon anyways.

sebsura (approved):

PR #2279, opened 2025-05-19
fix next-pool overrides by job

### LINKED ISSUE #2253: Copy job fails, because "Next Pool" is not accepted as an override in the job config.
### Bareos component version

Director: 23.1.3

### Steps to reproduce

Try to call an copy job, where the "Next pool" is only set in the job config.

### Expected results

That the copy job is executed.

### Actual results

Error message:
`22-Apr 10:05 Bareos Director JobId 0: Fatal error: No Next Pool specification found in Pool "SourcePool".`
But the [documentation](https://docs.bareos.org/Configuration/Director.html#config-Dir_Job_NextPool) say's:

> A Next Pool override used for Migration/Copy and Virtual Backup Jobs.

Therefore, if you overwrite undefined with defined, the result should be the overwritten value and not an error.
This means that the check as to whether there is a ‘next pool’ can only be carried out after all parameters have been analysed. Presumably the system simply checks whether there is a ‘Next Pool’ in the source pool, regardless of whether it is overwritten or not.

### Environment

```markdown
- OS: RHEL-8/9
```

### Relevant log output

```shell

```

### Relevant traces output

```shell

```

### Anything else?

Sample config:
Job:
```
Job {
        Name = "Test copy"
        Type = Copy
         # Jobdef only contain settings for update statistics and messages
        Job Defs = "WebUI"
        Pool = "SourcePool"
        Next Pool = "TargetPool"
        Selection Type = SQLQuery
        Selection Pattern = "select jobid from job where jobid=365279;"
}
```
Source pool:
```
Pool {
  Name = "SourcePool"
  Action On Purge = Truncate
  Job Retention = 1w
  File Retention = 1w
  Volume Retention = 1w
  Label Format = "Foo-"
  Maximum Volume Bytes = 1G
  Maximum Volumes = 60
  Storage = "Foo storage"
}
```
Target pool:
```
Pool {
  Name = "TargetPool"
  Action On Purge = Truncate
  Job Retention = 1w
  File Retention = 1w
  Volume Retention = 1w
  Label Format = "Bar-"
  Maximum Volume Bytes = 5G 
  Maximum Volumes = 100
  Storage = "Bar storage"
}
```

florian-at-bareos:
The next pool directive can be defined in the Pool and Job resource, the next pool directive in the Job resource should override the directive in the Pool resource. Currently, The directive in the job resource is not taken into account when running a job via the "run" command in the bconsole. This PR fixes this and takes the directive from the Job resource if specified.

Also, this PR adds a section in the copy-migrate systemtest to test the override behaviour of this directive.

Fixes #2253: Copy job fails, because "Next Pool" is not accepted as an override in the job config.

#### Please check

##### General

##### Source code quality

##### Tests

### DIFF
--- a/core/src/dird/ua_run.cc
+++ b/core/src/dird/ua_run.cc
@@ -2150,7 +2150,14 @@ static bool ScanCommandLineArguments(UaContext* ua, RunContext& rc)
       rc.next_pool = select_pool_resource(ua);
     }
   } else if (!rc.next_pool) {
-    rc.next_pool = rc.pool->NextPool; /* use default */
+    // use next pool from job
+    if (rc.job->next_pool) {
+      rc.next_pool = rc.job->next_pool;
+    }
+    // use next pool from pool
+    else {
+      rc.next_pool = rc.pool->NextPool;
+    }
   }
   if (rc.next_pool) {
     Dmsg1(100, "Using next pool %s\n", rc.next_pool->resource_name_);
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,6 +55,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - VMware Plugin: Adapt to pyVmomi 9 [PR #2341]
 - VMWare Plugin: Fix VirtualSerialPort, NVRAM timeouts configurable [PR #2344]
 - doc: storage backend add note about static build [PR #2350]
+- fix next-pool overrides by job [PR #2279]
 
 ### Documentation
 - docs: fix grpc-fd plugin call [PR #2068]
@@ -187,6 +188,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 [PR #2273]: https://github.com/bareos/bareos/pull/2273
 [PR #2275]: https://github.com/bareos/bareos/pull/2275
 [PR #2278]: https://github.com/bareos/bareos/pull/2278
+[PR #2279]: https://github.com/bareos/bareos/pull/2279
 [PR #2283]: https://github.com/bareos/bareos/pull/2283
 [PR #2286]: https://github.com/bareos/bareos/pull/2286
 [PR #2287]: https://github.com/bareos/bareos/pull/2287
--- a/docs/manuals/source/TasksAndConcepts/BareosConsole.rst
+++ b/docs/manuals/source/TasksAndConcepts/BareosConsole.rst
@@ -1050,6 +1050,8 @@ run
 
    Any information that is needed but not specified will be listed for selection, and before starting the job, you will be prompted to accept, reject, or modify the parameters of the job to be run, unless you have specified yes, in which case the job will be immediately sent to the scheduler.
 
+   All specified options override their counterpart in the configuration, if specified.
+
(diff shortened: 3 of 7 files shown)


florian-at-bareos:
The code is still awkward because the ScanCommandLineArguments method in ua_run.cc also sets values that are not given in the command line, applying overrides.

florian-at-bareos:
Also, is the name of the method ResetRestoreContext in ua_run.cc correct? The method updates jcr->dir_impl->res values from a RunContext reference, no RestoreContext at all.

core/src/dird/ua_run.cc
```diff
@@ -379,15 +378,15 @@ int DoRunCmd(UaContext* ua, const char*)
...
-    SetJcrDefaults(jcr, rc.job);
-    jcr->dir_impl->unlink_bsr
-        = ua->jcr->dir_impl->unlink_bsr; /* copy unlink flag from caller */
-    ua->jcr->dir_impl->unlink_bsr = false;
-  }
+  /* Create JobControlRecord to run job. */
+  auto jcr = std::unique_ptr<JobControlRecord,
+                             std::function<void(JobControlRecord*)>>(
+      NewDirectorJcr(DirdFreeJcr),
+      [&](JobControlRecord* to_delete) { FreeJcr(to_delete); });
```
sebsura: Why does this lambda capture everything ?  It only accesses its only argument.

core/src/dird/ua_run.cc
```diff
@@ -518,11 +517,6 @@ int DoRunCmd(UaContext* ua, const char*)
 
     return JobId;
   }
-
-bail_out:
-  ua->SendMsg(T_("Job not run.\n"));
```
sebsura: What happened to this message ?

core/src/dird/ua_run.cc
sebsura: Im also not a fan of the std::function here.
If its possible maybe we can just do
```
auto* jcr = NewDirectorJcr(...);
auto res = FunctionWithAllTheLogic(jcr);
FreeJcr(jcr);
```
This may be easier to read.  A `unique_ptr` isnt just something that calls a destructor automatically for you: It signifies ownership.  Clearly this function is _not_ the single owner of this jcr, as we later pass it to `RunJob()`, which holds on to a reference.
This makes the code somewhat confusing to read for me, e.g. here:
```
    JobId = RunJob(jcr.get());
```
i was about to write a comment saying that you probably wanted to use `.release()`  here but that would be completely wrong!

jcr already has a builtin reference counting and is shared by default, so using a unique_ptr here is not the right call.
Maybe we can create our own JCR smart pointer for cases such as this.

florian-at-bareos: I see! Do you think that it makes sense for me to work on that?
"Maybe we can create our own JCR smart pointer for cases such as this." sounds like it's a lot of effort. Do you have another idea on how to cleanly get rid of this "bail_out" goto or should i just rollback this for now?

arogge: There's `finally` available in the GSL. That could replace `goto` at least in some cases. Not sure if it is worth the effort though.

https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Re-finally

sebsura: Yes either roll it back or try to use `gsl::finally`.  I would be fine with either of them.

sebsura (approved):
Looks good to me.  Thanks for the work!

PR #238, opened 2019-08-01
sending all related Plugin Options to FD

benibr:
This is a WIP fix for https://bugs.bareos.org/view.php?id=1106

PR #2283, opened 2025-05-23
fix admin job issues

florian-at-bareos:
Fixes #238: admin job only prints first line of list result
Fixes #342: Provide resultset-less DB query function

#### Please check

##### General

##### Source code quality

##### Tests

### DIFF
--- a/core/src/cats/cats.h
+++ b/core/src/cats/cats.h
@@ -486,6 +486,9 @@ class BareosSqlError : public std::runtime_error {
   BareosSqlError(const char* what) : std::runtime_error(what) {}
 };
 
+// used in SqlQuery method
+struct SqlDiscardResult {};
+
 class BareosDb : public BareosDbQueryEnum {
  protected:
   brwlock_t lock_; /**< Transaction lock */
@@ -901,7 +904,8 @@ class BareosDb : public BareosDbQueryEnum {
   void FillQuery(PoolMem& query, SQL_QUERY predefined_query, ...);
 
   bool SqlQuery(SQL_QUERY query, ...);
-  bool SqlQuery(const char* query, int flags = 0);
+  bool SqlQuery(const char* query);
+  bool SqlQuery(const char* query, SqlDiscardResult);
   bool SqlQuery(const char* query, DB_RESULT_HANDLER* ResultHandler, void* ctx);
 
   /* sql_update.cc */
@@ -978,7 +982,8 @@ class BareosDb : public BareosDbQueryEnum {
   virtual int SqlNumFields(void) = 0;
   virtual void SqlFreeResult(void) = 0;
   virtual SQL_ROW SqlFetchRow(void) = 0;
-  virtual bool SqlQueryWithoutHandler(const char* query, int flags = 0) = 0;
+  virtual bool SqlQueryWithoutHandler(const char* query) = 0;
+  virtual bool SqlQueryWithoutHandler(const char* query, SqlDiscardResult) = 0;
   virtual bool SqlQueryWithHandler(const char* query,
                                    DB_RESULT_HANDLER* ResultHandler,
                                    void* ctx)
@@ -1020,9 +1025,6 @@ BareosDb* db_init_database(JobControlRecord* jcr,
                            bool try_reconnect,
                            bool exit_on_fatal,
                            bool need_private = false);
-
-/* SqlQuery Query Flags */
-#define QF_STORE_RESULT 0x01
 
 /* flush the batch insert connection every x changes */
 #define BATCH_FLUSH 800000
--- a/core/src/cats/postgresql.cc
+++ b/core/src/cats/postgresql.cc
@@ -133,8 +133,7 @@ bool BareosDbPostgresql::CheckDatabaseEncoding(JobControlRecord* jcr)
   SQL_ROW row;
   bool retval = false;
 
-  if (!SqlQueryWithoutHandler("SELECT getdatabaseencoding()",
-                              QF_STORE_RESULT)) {
+  if (!SqlQueryWithoutHandler("SELECT getdatabaseencoding()")) {
     Jmsg(jcr, M_ERROR, 0, "%s", errmsg);
     return false;
   }
@@ -514,7 +513,7 @@ bool BareosDbPostgresql::SqlQueryWithHandler(const char* query,
   Dmsg1(500, "SqlQueryWithHandler starts with '%s'\n", query);
 
   DbLocker _{this};
-  if (!SqlQueryWithoutHandler(query, QF_STORE_RESULT)) {
+  if (!SqlQueryWithoutHandler(query)) {
     Mmsg(errmsg, T_("Query failed: %s: ERR=%s\n"), query, sql_strerror());
     Dmsg0(500, "SqlQueryWithHandler failed\n");
     return false;
@@ -544,7 +543,7 @@ bool BareosDbPostgresql::SqlQueryWithHandler(const char* query,
  * Returns:  true  on success
  *           false on failure
  */
-bool BareosDbPostgresql::SqlQueryWithoutHandler(const char* query, int)
+bool BareosDbPostgresql::SqlQueryWithoutHandler(const char* query)
 {
   int i;
   bool retry = true;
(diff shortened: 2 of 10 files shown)


core/src/tests/main.cc
```diff
@@ -0,0 +1,8 @@
+#include <gtest/gtest.h>
+
+int main(int argc, char** argv)
+{
+  testing::InitGoogleTest(&argc, argv);
+
+  return RUN_ALL_TESTS();
+}
```
sebsura: Im not sure i understand the reoson for this.  Why doesnt this just use the "default" gtest main class like all the other unit tests ?

core/src/cats/cats.h
```diff
@@ -978,7 +982,8 @@ class BareosDb : public BareosDbQueryEnum {
   virtual int SqlNumFields(void) = 0;
   virtual void SqlFreeResult(void) = 0;
   virtual SQL_ROW SqlFetchRow(void) = 0;
-  virtual bool SqlQueryWithoutHandler(const char* query, int flags = 0) = 0;
+  virtual bool SqlQueryWithoutHandler(const char* query) = 0;
+  virtual bool SqlQueryWithoutHandler(const char* query, SqlDiscardResult) = 0;
```
sebsura: Why dont we just use a different name for this ?  `SqlQueryWithoutResult` or something like that.

core/src/dird/dird.cc
```diff
@@ -121,7 +121,7 @@ static bool DirDbLogInsert(JobControlRecord* jcr,
   Mmsg(query, "INSERT INTO Log (JobId, Time, LogText) VALUES (%s,'%s','%s')",
        edit_int64(jcr->JobId, ed1), dt, esc_msg.c_str());
 
-  return jcr->db->SqlQuery(query.c_str());
+  return jcr->db->SqlQuery(query.c_str(), SqlDiscardResult{});
```
sebsura: Maybe something like

```suggestion
  return jcr->db->SqlQueryIgnoringResult(query.c_str());
```
would sound ok ?

core/src/cats/cats.h
```diff
@@ -901,7 +904,8 @@ class BareosDb : public BareosDbQueryEnum {
   void FillQuery(PoolMem& query, SQL_QUERY predefined_query, ...);
 
   bool SqlQuery(SQL_QUERY query, ...);
-  bool SqlQuery(const char* query, int flags = 0);
+  bool SqlQuery(const char* query);
+  bool SqlQuery(const char* query, SqlDiscardResult);
```
sebsura: See SqlQueryWithoutHandler

core/src/cats/postgresql.cc
```diff
@@ -133,8 +133,7 @@ bool BareosDbPostgresql::CheckDatabaseEncoding(JobControlRecord* jcr)
   SQL_ROW row;
   bool retval = false;
 
-  if (!SqlQueryWithoutHandler("SELECT getdatabaseencoding()",
-                              QF_STORE_RESULT)) {
+  if (!SqlQueryWithoutHandler("SELECT getdatabaseencoding()")) {
```
sebsura, line 192: Its interesting that we basically already had a flag that tells us whether to store the result or not, but the psql backend just ignored this.  I wonder if the old backends (mysql, etc.) did properly check that flag.

core/src/cats/postgresql.cc
```diff
@@ -638,6 +637,73 @@ bool BareosDbPostgresql::SqlQueryWithoutHandler(const char* query, int)
...
+                                                SqlDiscardResult)
+{
+  constexpr int kMaximumTries = 2;
+
+  AssertOwnership();
+  Dmsg1(500, "SqlQueryWithoutHandler (discard result) starts with '%s'\n",
+        query);
+
+  // try query until successfull or maxium number of tries
+  for (int current_try = 0; current_try < kMaximumTries; ++current_try) {
```
sebsura: just having `current_try < 2` is probably easier to read. If this was used more than once (e.g. also in the function above), then i would agree that giving it a descriptive name would be better.

core/src/cats/postgresql.cc
```diff
@@ -638,6 +637,73 @@ bool BareosDbPostgresql::SqlQueryWithoutHandler(const char* query, int)
...
+    std::unique_ptr<PGresult, std::function<void(PGresult*)>> result = nullptr;
```
sebsura: Im not a fan of this.  Not only is this 40 bytes (!) big, but everytime you assign to it you create a new std::function on the heap (with an allocation) that just calls PGclear();

I would suggest the following: you can set the c++ version to c++20 as we recently updated our compilers (but not the stdlib!) to allow that.

Then you can simply define
```C++
using pgres_ptr = std::unique_ptr<PGresult, decltype([] (PGresult* ptr) -> void { PQclear(ptr); })>;
``` 

This type is only `8` bytes big and does not allocate any memory on the heap on its own and also does not contain this footgun:

```C++
{
  std::unique_ptr<PGresult, std::function<void(PGresult*)>> result{};
  result.reset(PQexec(...));
}  // oops, we crashed :(
```

core/src/cats/postgresql.cc
```diff
@@ -638,6 +637,73 @@ bool BareosDbPostgresql::SqlQueryWithoutHandler(const char* query, int)
...
+    // try to exectute query in a 5 sec interval 10 times
+    for (int i = 0; i < 10; i++) {
+      if ((result = std::unique_ptr<PGresult, std::function<void(PGresult*)>>(
+               PQexec(db_handle_, query), PQclear))) {
+        break;
+      }
```
sebsura: This can simply become
```C++
result.reset(PQexec(...));
if (result) { break; }
```
which also gets rid of the ugly `(())`

core/src/cats/postgresql.cc
```diff
@@ -638,6 +637,73 @@ bool BareosDbPostgresql::SqlQueryWithoutHandler(const char* query, int)
...
+        if (PQstatus(db_handle_) == CONNECTION_OK) {
+          // Reset the connection settings.
+          // prevent leak
+          result = std::unique_ptr<PGresult, std::function<void(PGresult*)>>(
+              PQexec(db_handle_,
+                     "SET datestyle TO 'ISO, YMD';"
+                     "SET cursor_tuple_fraction=1;"
+                     "SET standard_conforming_strings=on;"
+                     "SET client_min_messages TO WARNING;"),
+              PQclear);
```
sebsura: See above

core/src/cats/postgresql.cc
```diff
@@ -638,6 +637,73 @@ bool BareosDbPostgresql::SqlQueryWithoutHandler(const char* query, int)
...
+
+          if (!result || PQresultStatus(result.get()) != PGRES_COMMAND_OK) {
```
sebsura: Do we need to check for nullptr here ? We dont do this at line 661 after all.

core/src/cats/postgresql.cc
```diff
@@ -638,6 +637,73 @@ bool BareosDbPostgresql::SqlQueryWithoutHandler(const char* query, int)
 ok_out:
   return retval;
 }
+bool BareosDbPostgresql::SqlQueryWithoutHandler(const char* query,
```
sebsura: Im not a fan of how much duplicated code we have here.

Would it be possible to just have one function that returns a new result object and then
the old SqlQuery function would just do `result_ = res.release();` where as the new one would simply ignore the return value.

Its probably not possible to deduplicate all of the logic of these functions, but it should definitely be possible to at least deduplicate some parts of them, e.g. the reconnect logic.

core/src/cats/postgresql.cc
```diff
@@ -638,6 +637,73 @@ bool BareosDbPostgresql::SqlQueryWithoutHandler(const char* query, int)
...
+          Emsg1(M_ERROR_TERM, 0, "Fatal database error: %s\n", sql_strerror());
+        }
+
+        /* Only try reconnecting when no transaction is pending.
+         * Reconnecting within a transaction will lead to an aborted
+         * transaction anyway so we better follow our old error path. */
+        if (!try_reconnect_ || transaction_
+            || current_try + 1 == kMaximumTries) {
+          return false;
+        }
```
sebsura: I didnt think this quite through yet, but the logic could be split like so ? 
```
void SqlQuery...(...) {
  auto res = TryQuery(...);
  if (res.recoverable_error() && try_reconnect_ && !transaction_) {
     res = TryQuery(...);
  }
  // handle the actual result now
}
```

where `res.recoverable_error()` indicates that `PGRES_FATAL_ERROR` was returned.

This might be easier to understand, but im not sure how this will look in the end.  Maybe you can try this out.

core/src/tests/main.cc
florian-at-bareos: I saw that this main method is in the catalog.cc unit-test. So I moved it to its own file. Isn't this main method for all unit tests?

core/src/cats/cats.h
florian-at-bareos: I didn't come up with a proper name for them. Especially for this one
`virtual bool SqlQueryWithoutHandler(const char* query, SqlDiscardResult) = 0;`

I don't like `SqlQueryWithoutHandleIgnoringResults` because it's so long.

Also I think that this shouldn't be regarded as a seperate function with different meaning, but more of a parameter you pass to the query to discard the result or not. This being 2 methods is more like "implementation" details in my opinion.

core/src/tests/main.cc
sebsura: No normally we use the default main method.  See e.g. here
```
  bareos_add_test(
    addresses_and_ports_config
    LINK_LIBRARIES Bareos::Lib Bareos::Dir Bareos::Findlib Bareos::SQL
                   testing_common GTest::gtest_main
  )
```
the `GTest::gtest_main` target contains the "default" gtest main (afaik).

florian-at-bareos: Ok, then I'll just remove this file

sebsura (approved):
Great Work!

Issue #2312, opened 2025-07-02
Update job command crashes director when jobid doesn't exist


SamuelBoerlin:
### Bareos component version

bareos-fd --version
24.0.4~pre30.dd9a00734

bareos-dir --version
24.0.4~pre30.dd9a00734

bareos-sd --version
24.0.4~pre30.dd9a00734

bconsole --version
24.0.4~pre30.dd9a00734

### Steps to reproduce

1. Run `update jobid=*None* jobtype=A`
2. Director crashes

### Expected results

Director shouldn't crash, regardless of invalid `jobid` argument.

### Actual results

The director crashes.

### Environment

```markdown
- OS: `20.04.6 LTS (Focal Fossa)`
```

### Relevant log output

```shell
After restarting, I can also see the following messages which happen right after the update command:

*m
02-Jul 10:28 bareos-dir: ERROR in lib/rwlock.cc:275 Failed assert called from strerror ../../../../core/src/cats/sql.cc:339
02-Jul 10:28 bareos-dir: ERROR in lib/rwlock.cc:281 Failed ASSERT: is_locked
```

### Relevant traces output

```shell
I have a .traceback, but I'm not sure if it could contain sensitive information.
Please let me know if you need it.
```

bruno-at-bareos:
Can also crash the dir during update volume=VOL0000 pool=NONEXISTING
or interactively when you type a non existing pool name

PR #2331, opened 2025-07-24
cats: fix missing locks

sebsura:

This pr fixes issue #2312.  I added the missing locks and also removed the builtin crash from the ownership check.  As the director going down is a pretty major issue, it should not happen if only some error messages get potentially misread.

This also fixes issue #2336 ...

We added also a fix for ``subscription`` view to better classified filer units, reported by customer with support.

#### Please check

##### General

##### Source code quality


core/src/cats/cats.h
```diff
@@ -1041,7 +1041,7 @@ class BareosDb : public BareosDbQueryEnum {
   void AssertOwnership(libbareos::source_location l
                        = libbareos::source_location::current())
   {
-    if (!is_private_) { RwlAssertWriterIsMe(&lock_, l); }
+    if (!is_private_) { RwlCheckWriterIsMe(&lock_, l); }
   }
 };
```
arogge: `AssertOwnership()` should either assert something or get another name.

arogge:
We should avoid the double-lock on batch-inserts, especially as the error retrieval we added a lock for doesn't work when batch-insert is used.

core/src/console/console.cc
```diff
@@ -80,9 +80,7 @@ static bool file_selection = false;
 static bool force_send_pam_credentials_unencrypted = false;
 static bool use_pam_credentials_file = false;
 static std::string pam_credentials_filename;
-static const std::string program_arguments{"D:lc:d:np:ostu:x:z:?"};
 #else
-static const std::string program_arguments{"D:lc:d:nstu:x:z:?"};
```
arogge, line 85: the empty `#else` should be removed

core/src/dird/vbackup.cc
```diff
@@ -541,12 +541,14 @@ void NativeVbackupCleanup(JobControlRecord* jcr, int TermCode, int JobLevel)
  */
 static int InsertBootstrapHandler(void* ctx, int, char** row)
 {
-  JobId_t JobId;
-  int FileIndex;
-  RestoreBootstrapRecord* bsr = (RestoreBootstrapRecord*)ctx;
+  auto* rx = reinterpret_cast<RestoreContext*>(ctx);
```
arogge:
```suggestion
  auto* rx = static_cast<RestoreContext*>(ctx);
```
from and to `void*` should use `static_cast<>`.

core/src/dird/vbackup.cc
```diff
@@ -564,12 +566,21 @@ static bool CreateBootstrapFile(JobControlRecord& jcr,
   rx.bsr = std::make_unique<RestoreBootstrapRecord>();
   rx.JobIds = jobids_pm.addr();
 
+  rx.found = false;
+
   if (DbLocker _{jcr.db_batch}; !jcr.db_batch->GetFileList(
           &jcr, rx.JobIds, false /* don't use md5 */, true /* use delta */,
-          InsertBootstrapHandler, (void*)rx.bsr.get())) {
+          InsertBootstrapHandler, (void*)&rx)) {
```
arogge: Do we really want to pass the address of the pointer here?
Also, the cast to `void*` is implicit.

core/src/lib/edit.cc
```diff
@@ -341,14 +341,14 @@ char* edit_utime(utime_t val, char* buf, int buf_len)
       = {60 * 60 * 24 * 365, 60 * 60 * 24 * 30, 60 * 60 * 24, 60 * 60, 60};
   static const char* mod[] = {"year", "month", "day", "hour", "min"};
   int i;
-  uint32_t times;
+  std::size_t times;
```
arogge: this assumes `std::size_t` is equal to `utime_t`.
You can probably just statically assert this.

arogge (changes requested):
I have a few minor comments.

core/src/dird/vbackup.cc
sebsura: ill remove the void cast

core/src/lib/edit.cc
sebsura: i just used utime_t instead.  Seems easier

core/src/cats/cats.h
sebsura: Renamed it to `CheckOwnership`

core/src/dird/vbackup.cc
```diff
@@ -564,12 +566,21 @@ static bool CreateBootstrapFile(JobControlRecord& jcr,
...
+          InsertBootstrapHandler, &rx)) {
     Jmsg(&jcr, M_ERROR, 0, "%s\n", jcr.db_batch->strerror());
   }
 
+  if (!rx.found) {
+    Jmsg(&jcr, M_ERROR, 0, "No files were found in the jobids %s\n", rx.JobIds);
+    return false;
+  }
+
+  rx.found = false;
```
arogge: This is either wrong or I can't (locally) reason about this, in which case a comment would help.

arogge (changes requested):
Sorry, I have one final thing...

core/src/dird/vbackup.cc
sebsura: `RestoreContext::found` is used to say "the last handler executed at least once", i.e. "at least one result was found".  This is only relevant for the code directly after the query.

I was just trying to be hygenic: Some of the code relied on `found` being false before their call (because if its true before the call then the bool does not say anything), so I did not want to change the value of rx.found for the functions that this function calls.   Obviously the flag is currently not getting used afterwards anymore, so this is probably unnecessary (until someone adds a query to e.g. WriteBsrFile for some reason).

I also added fixes for another function which did not properly reset this flag before using it.

sebsura: I checked that all uses of `rx->found` now actually properly reset this flag, so i removed this (now unnecessary) reset.

core/platforms/packaging/bareos.spec
```diff
@@ -922,11 +922,7 @@ cmake  .. \
 %endif
   -Dwebuiconfdir=%{_sysconfdir}/bareos-webui \
   -DVERSION_STRING=%version
-%if 0%{?make_build:1}
-%make_build
-%else
-%__make %{?_smp_mflags};
-%endif
+cmake --build "." %{?_smp_mflags}
```
arogge: while I appreciate the idea, this is not sufficient to make things ninja-capable, as we still call `make install/fast` in `%install`.
Basically we'd want to switch to `cmake --install`, but that's only available on 3.18 and up while we still support 3.17 for some old distro.

I guess it is reasonable to create a task to
* figure out what range of CMake versions we have to support
* update `cmake_minimum_required()`
* update all the build scripts to use `cmake --build` and `cmake --install`

arogge:
Are you actually going to backport this down into 23?

PR #2341, opened 2025-08-08
VMware Plugin: Adapt to pyVmomi 9

sduehr:
The import of the VmomiJSONEncoder was changed in pyVmomi9, this change adapts the Bareos VMware plugin so that it works with both the new and the old pyVmomi versions.

##### General

##### Source code quality

### DIFF
--- a/core/src/plugins/filed/python/vmware/bareos-fd-vmware.py
+++ b/core/src/plugins/filed/python/vmware/bareos-fd-vmware.py
@@ -52,7 +52,7 @@ from pyVmomi import vim
 from pyVmomi import vmodl
 import pyVim.task
 
-# adapt to change in pyVmomi 9.x
+# Try import pyVmomi 9.x first and fallback to old method
 try:
     from pyVmomi.VmomiJSONEncoder import VmomiJSONEncoder
 except ImportError:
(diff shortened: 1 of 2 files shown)


core/src/plugins/filed/python/vmware/bareos-fd-vmware.py
```diff
@@ -51,7 +51,12 @@
 from pyVmomi import vim
 from pyVmomi import vmodl
 import pyVim.task
-from pyVmomi.VmomiSupport import VmomiJSONEncoder
+
+# adapt to change in pyVmomi 9.x
```
bruno-at-bareos: To make clear on first read which is the new call and which was the old one I would suggest to change the comment to:
```suggestion
# Try import pyVmomi 9.x first and fallback to old method
```

bruno-at-bareos (changes requested):
Except an adjustment to comment, it's fine

core/src/plugins/filed/python/vmware/bareos-fd-vmware.py
sduehr: Thanks, I just replaced the comment as suggested.

bruno-at-bareos (approved):
Great job, thanks.

PR #2378, opened 2025-09-25
cats: fix missing locks

sebsura:
**Backport of PR #2331 to bareos-24**

Removed the fix to the db schema, as we have no good way to deploy these on a minor upgrade.

##### General

#### Backport quality
- [x] Original PR #2331 is merged

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 - systemtest: speed up always-incremental with sleep 0.25 [PR #2366]
 - storage: improve tapealert plugin [PR #2382]
+- cats: fix missing locks [PR #2378]
 
 ## [24.0.5] - 2025-09-02
 
@@ -527,6 +528,7 @@ It is therefore strongly suggested to immediately schedule a full backup of your
 [PR #2371]: https://github.com/bareos/bareos/pull/2371
 [PR #2374]: https://github.com/bareos/bareos/pull/2374
 [PR #2377]: https://github.com/bareos/bareos/pull/2377
+[PR #2378]: https://github.com/bareos/bareos/pull/2378
 [PR #2382]: https://github.com/bareos/bareos/pull/2382
 [PR #2387]: https://github.com/bareos/bareos/pull/2387
 [unreleased]: https://github.com/bareos/bareos/tree/master


core/src/cats/ddl/creates/postgresql.sql
```diff
@@ -527,7 +528,7 @@ SELECT
     ELSE j.jobbytes / 1000000
   END AS unknown_mb,
   CASE
-    WHEN f.filesettext ILIKE '%{%{%meta%=' THEN NULL
+    WHEN f.filesettext ILIKE '%{%{%meta%=%' THEN NULL
     WHEN f.filesettext ILIKE '%{%{%Plugin%=%mssqlvdi:%' THEN NULL
     WHEN f.filesettext ILIKE '%{%{%Plugin%=%python%:module_name=bareos-fd-mariabackup%' THEN NULL
     WHEN f.filesettext ILIKE '%{%{%Plugin%=%python%:module_name=bareos-fd-percona%' THEN NULL
```
arogge, line 533: we should not change schema in a backport

core/CMakeLists.txt
```diff
@@ -40,13 +40,9 @@ set(CMAKE_INSTALL_DEFAULT_COMPONENT_NAME common)
...
 
-# switch on CXX 17 Support
+# switch on CXX 20 Support
 #
-if(${CMAKE_SYSTEM_NAME} MATCHES "Windows")
-  set(CMAKE_CXX_STANDARD 20)
-else()
-  set(CMAKE_CXX_STANDARD 17)
-endif()
+set(CMAKE_CXX_STANDARD 20)
```
arogge: This would require C++20 for Bareos 24, do we actually want/need that?

sebsura: I think it is required for the new test.  I removed the test as you suggested.  Lets see if it still builds or if I need to change something else.

core/src/cats/ddl/creates/postgresql.sql
bruno-at-bareos: From my point of view as we only change views there's no literal schema changes.
As no updates of the database will run there's no risk at all for users. But on support side we had to already applied manually those changes to some infrastructure, due to bad filer reports.
The benefit to have it in the backport is the following:
- all new 24 installations will have the fix for free
- applying the fix is just a question of re-running the update catalog script which is more convenient for end user than copy/pasting sql instruction on psql console.

bruno-at-bareos: After internal discussion we agreed we won't change the catalog schema in 24 updates.
People using ndmp wanting to have accurate report will have to do the view update manually. contact support@bareos.com.

bruno-at-bareos: @sebsura would you mind to care to remove the commit ?

sebsura: removed it

PR #2536, opened 2026-02-05
cleanup: remove `basefiles` table

### LINKED ISSUE #2334: Remove basefile table and friends
As part of phasing out base jobs as announced in the Bareos 24 release notes the `basefile` table should be removed from the catalog. Also all related information (like the `has_base` column in the `job` table) should be removed.

This will require all queries to be reviewed to make sure the table is not used any more. Also where there are different queries with a basefile/non-basefile version we need to check that the right version of the query is used.

The database upgrade script needs to check that the basefile table is actually empty or that there is no job record with `has_base` enabled before proceeding. It must fail otherwise.

pstorz:

Base Jobs are already deprecated for quite some time, we now also remove the internal parts that are not needed anymore to support that feature.

#### Please check

##### General

##### Source code quality

##### Tests


core/src/dird/ua_cmds.cc
```diff
@@ -166,7 +166,6 @@ static bool DoTruncate(UaContext* ua,
 bool quit_cmd(UaContext* ua, const char* cmd);
 
 const char list_cmd_usage[] = NT_(
-    "basefiles ( jobid=<jobid> | ujobid=<complete_name> ) | "
```
joergsteffens, line 169: I think, `ujobid=<complete_name> ) | ` has been removed accidentally.

joergsteffens (dismissed):

core/src/dird/ua_cmds.cc
sebsura: I think these were options to the `list basefiles` command, which is now removed. Or are you reading this differently ?

I.e. `list volumes` still has `ujobid` as argument.

core/src/cats/sql_list.cc
```diff
@@ -648,14 +648,9 @@ void BareosDb::ListFilesForJob(JobControlRecord* jcr,
...
        "FROM (SELECT PathId, Name FROM File WHERE JobId=%s "
-       "UNION ALL "
-       "SELECT PathId, Name "
-       "FROM BaseFiles JOIN File "
-       "ON (BaseFiles.FileId = File.FileId) "
-       "WHERE BaseFiles.JobId = %s"
        ") AS F, Path "
        "WHERE Path.PathId=F.PathId ",
-       edit_int64(jobid, ed1), ed1);
+       edit_int64(jobid, ed1));
```
arogge, line 654: no need to nest this into a subquery anymore

core/src/cats/ddl/creates/postgresql.sql
```diff
@@ -321,27 +321,6 @@ CREATE TABLE counters
...
-
--- This table seems to be obsolete
--- CREATE TABLE UnsavedFiles
--- (
---     UnsavedId         INTEGER     NOT NULL,
---     JobId             INTEGER     NOT NULL,
---     PathId            INTEGER     NOT NULL,
---     FilenameId        INTEGER     NOT NULL,
---     PRIMARY KEY (UnsavedId)
--- );
```
arogge, line 344: schema change needs to update the versionid
also, the migration script is still missing

arogge (changes requested):
the database versionid needs to be updated and a migration script removing the table is missing.
Also, the upgrade script should fail if there is still data in the table. This can be done as follows:
```
DO $$
BEGIN
    IF EXISTS (SELECT 1 FROM basefiles LIMIT 1) THEN
        RAISE EXCEPTION 'Table basefiles is not empty';
    END IF;
END
$$;
```

core/src/cats/bvfs.cc
```diff
@@ -588,13 +585,8 @@ static void build_ls_files_query(JobControlRecord*,
...
 {
-  if (db->GetTypeIndex() == SQL_TYPE_POSTGRESQL) {
-    db->FillQuery(query, BareosDb::SQL_QUERY::bvfs_list_files, JobId, PathId,
-                  JobId, PathId, filter, limit, offset);
-  } else {
-    db->FillQuery(query, BareosDb::SQL_QUERY::bvfs_list_files, JobId, PathId,
-                  JobId, PathId, limit, offset, filter, JobId, JobId);
-  }
+  db->FillQuery(query, BareosDb::SQL_QUERY::bvfs_list_files, JobId, PathId,
+                filter, limit, offset);
```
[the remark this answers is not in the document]
pstorz: good catch

core/src/cats/postgresql.cc
```diff
@@ -3,7 +3,7 @@
 
    Copyright (C) 2003-2011 Free Software Foundation Europe e.V.
    Copyright (C) 2011-2016 Planets Communications B.V.
-   Copyright (C) 2013-2025 Bareos GmbH & Co. KG
+   Copyright (C) 2013-2026 Bareos GmbH & Co. KG
```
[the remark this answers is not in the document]
pstorz: good catch, will remove the wrong comment

core/src/cats/dml/0053_batch_fill_filename_query
```diff
@@ -1,6 +1,11 @@
 INSERT INTO Filename (Name)
```
arogge: I know I'm nitpicking, but the Filename table is long gone, so this query is pointless.

core/src/cats/dml/0050_batch_lock_filename_query
```diff
@@ -1 +1 @@
-BEGIN
+BEGIN; LOCK TABLE Filename IN SHARE ROW EXCLUSIVE MODE
```
arogge: That tables does not exist anymore.

core/src/cats/dml/0055_match_query2
```diff
@@ -1 +1 @@
- AND Filename.Name LIKE '%s'
+ AND Filename.Name ~ '%s'
```
arogge: This can never work, because there is no Filename relation in the `bvfs_list_files` query. Thus, setting a filter on `bvfs_lsfiles` will always fail.

core/src/cats/dml/0034_select_recent_version_with_basejob
```diff
@@ -7,35 +7,35 @@
...
-       LStat,
-       MD5,
-       DeltaSeq,
-       Fhinfo,
-       Fhnode,
+       File.LStat,
+       File.MD5,
+       File.DeltaSeq,
+       File.Fhinfo,
+       File.Fhnode,
```
arogge: we should either alias all or none of the fields, but not just some.

core/src/cats/bvfs.cc
```diff
@@ -588,13 +585,8 @@ static void build_ls_files_query(JobControlRecord*,
                                  int64_t limit,
                                  int64_t offset)
 {
-  if (db->GetTypeIndex() == SQL_TYPE_POSTGRESQL) {
```
arogge: AFAICT this is the only use of `GetTypeIndex()` and `SQL_TYPE_POSTGRESQL`, so these could/should be removed.

arogge (changes requested):
Looks pretty good already, but there are a few things that should still be improved. Also the commit messages need work.

core/src/cats/bvfs.cc
```diff
@@ -174,11 +174,8 @@ bool BareosDb::UpdatePathHierarchyCache(JobControlRecord* jcr,
        "INSERT INTO PathVisibility (PathId, JobId) "
        "SELECT DISTINCT PathId, JobId "
        "FROM (SELECT PathId, JobId FROM File WHERE JobId = %s "
-       "UNION "
-       "SELECT PathId, BaseFiles.JobId "
-       "FROM BaseFiles JOIN File AS F USING (FileId) "
-       "WHERE BaseFiles.JobId = %s) AS B",
-       jobid, jobid);
+       ") AS B",
+       jobid);
```
arogge: I don't see a reason for an inner query here.

systemtests/tests/upgrade-database-basefiles-not-empty/CMakeLists.txt
```diff
@@ -0,0 +1,27 @@
...
+#   Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+#   02110-1301, USA.
+
+get_filename_component(BASENAME ${CMAKE_CURRENT_BINARY_DIR} NAME)
+if(postgresql)
+  create_systemtest(${SYSTEMTEST_PREFIX} ${BASENAME})
+else()
+  create_systemtest(
+    ${SYSTEMTEST_PREFIX} ${BASENAME} DISABLED COMMENT "requires postgresql"
+  )
```
arogge: that condition doesn't make sense

systemtests/tests/upgrade-database-basefiles-not-empty/testrunner
```diff
@@ -0,0 +1,47 @@
...
+
+#shellcheck source=../../environment.in
+. ./environment
+
+#shellcheck source=../../scripts/functions
+. "${BAREOS_SCRIPTS_DIR}"/functions
+"${BAREOS_SCRIPTS_DIR}"/cleanup
+
+# use the legacy schema so update_bareos_tables runs the 2250 -> 2260 migration
+export SQL_DDL_DIR="${current_test_directory}/../upgrade-database/bareos_15_db_2004_ddl"
```
arogge: I really don't like this. The test should not depend on files from other systemtest directories and it should test an upgrade from 2250 instead of 2004.

systemtests/tests/CMakeLists.txt
```diff
@@ -116,6 +116,7 @@ add_subdirectory(tlsrestricted)
 add_subdirectory(tls-suites)
 add_subdirectory(truncate-command)
 add_subdirectory(upgrade-database)
+add_subdirectory(upgrade-database-basefiles-not-empty)
```
arogge: IIRC when a test directory name is the prefix of another test's directory name, (i.e. "upgrade-database" and "upgrade-database-...") the setup/cleanup of the shorter name will potentially kill the processes of the test with the longer name leading to spurious test failures.

arogge (changes requested):

core/src/cats/bdb_query_enum_class.h
```diff
@@ -41,8 +41,8 @@ class BareosDbQueryEnum {
     uar_jobid_fileindex = 30,
     uar_jobids_fileindex = 31,
     uar_jobid_fileindex_from_table = 32,
-    select_recent_version_with_basejob = 33,
-    select_recent_version_with_basejob_and_delta = 34,
+    select_recent_version_from_jobids = 33,
+    select_recent_version_from_jobids_and_delta = 34,
     select_recent_version = 35,
```
sebsura: This seems to be completely unused.  Please check each query if it has actual uses and delete all of the ones that are unused, as otherwise they will just rot anyways.

core/src/cats/dml/0044_uar_jobid_fileindex_from_dir
```diff
@@ -1,4 +1,8 @@
 # Query to get all files in a directory -- no recursing
+# Note, for PostgreSQL since it respects the Single Value
+# rule, the results of the SELECT will be unoptimized.
+# I.e. the same file will be restored multiple times, once
+# for each time it was backed up.
```
sebsura: This sounds like a bug.  The fix seems simple: Use `GROUP BY` like the "default" query and add `MAX(JobId)` to make the query unambiguous.

core/src/cats/dml/0048_bvfs_list_files
```diff
@@ -1,44 +1,29 @@
...
+      File.JobId IN (%s)
+      AND File.PathId = %s
+      AND File.Name != ''
+      %s
+   ORDER BY File.Name,
+            Job.StartTime DESC
+) AS A
+WHERE A.FileIndex > 0
+LIMIT %lld
+OFFSET %lld
```
sebsura: These are the wrong specifiers.  The correct one is PRId64

core/src/cats/dml/0055_match_query2
```diff
@@ -1 +1 @@
- AND Filename.Name LIKE '%s'
+ AND File.Name LIKE '%s' ESCAPE '\'
```
sebsura, line 3: Why was the `ESCAPE` added ? It would be great if this file (and match_query) could have comments explaining why this is necessary.

core/src/cats/dml/0074_bvfs_ls_special_dirs_3
sebsura: Should this be `directories` ?

core/src/cats/postgresql_queries.inc
```diff
@@ -636,69 +587,45 @@ R"SQL(CREATE TABLE %s AS
...
+      AND File.Name != ''
       %s
-   ORDER BY FileName,
-            StartTime DESC
+   ORDER BY File.Name,
+            Job.StartTime DESC
 ) AS A
 WHERE A.FileIndex > 0
 LIMIT %)SQL" PRId64 R"SQL(
 OFFSET %)SQL" PRId64 R"SQL(
```
sebsura, line 565: The queryfile was clearly not regenerated.  This still has the correct specifiers.

sebsura (changes requested):
The query file was not regenerated which makes most of the tests useless.  In fact regenerating the file shows that bareos does not even build anymore.  Please regenerate the file and fix the build issues.

systemtests/tests/upgrade-database-basefiles-not-empty/testrunner
pstorz: addressed in 4712bfd9e6ebf93cea3ed6de7cb1984fd66453f9

systemtests/tests/upgrade-database-basefiles-not-empty/CMakeLists.txt
pstorz: addressed in 4712bfd9e6ebf93cea3ed6de7cb1984fd66453f9

systemtests/tests/CMakeLists.txt
pstorz: addressed in 4712bfd9e6ebf93cea3ed6de7cb1984fd66453f9

core/src/cats/bvfs.cc
pstorz: addressed in f1982ff35d2da505be54849277454983a071ea84

core/src/cats/dml/0055_match_query2
pstorz: addressed in bebe1d67ce5615c7a78690538ce601edeff7f4f0

core/src/cats/postgresql_queries.inc
pstorz: Fixed in f12c7fca0d; generator was corrected, files regenerated, and postgresql_queries.inc

core/src/cats/ddl/updates/postgresql.2250_2260.sql
```diff
@@ -0,0 +1,32 @@
...
+
+  IF remaining_rows != 0 THEN
+    RAISE EXCEPTION
+      'Refusing to drop non-empty basefiles table during 2250 to 2260 migration (% rows remain).',
+      remaining_rows;
+  END IF;
+END
+$$ LANGUAGE 'plpgsql';
+
+DROP TABLE IF EXISTS public.basefiles;
```
sebsura, line 23: Is there a difference between `public.basefiles` and `basefiles` ?

core/src/cats/dml/0004_fill_jobhisto
```diff
@@ -4,13 +4,12 @@
...
 	       VolSessionId, VolSessionTime, JobFiles, JobBytes, ReadBytes, JobErrors, JobMissingFiles, PoolId, FileSetId,
-	       PriorJobId, PurgedFiles,
-	       HasBase, Reviewed, Comment
+	       PriorJobId, PurgedFiles, HasCache,
+	       Reviewed, Comment
        )
 SELECT JobId, Job, Name, Type, Level, ClientId, JobStatus, SchedTime, StartTime, EndTime, RealEndTime, JobTDate, VolSessionId,
        VolSessionTime, JobFiles, JobBytes, ReadBytes, JobErrors, JobMissingFiles, PoolId, FileSetId, PriorJobId, PurgedFiles,
-       HasBase,
-       Reviewed, Comment
```
sebsura, line 13: Why do we want to save `HasCache` ? `HasCache` is just the bvfs status, I do not think it makes sense to actually store that.

core/src/cats/CMakeLists.txt
```diff
@@ -37,6 +37,23 @@ target_sources(
...
+  check-cats-queryfiles
+  COMMAND ${CMAKE_COMMAND} -E chdir ${CMAKE_CURRENT_SOURCE_DIR}/dml
+          ./create_queryfiles.sh
+  COMMAND git -C ${CMAKE_SOURCE_DIR} diff --no-ext-diff -p --exit-code --stat=80 --
+          ${cats_generated_query_files}
+  COMMENT "Comparing generated catalog query files with git repo"
+  VERBATIM
+)
+add_dependencies(check-git check-cats-queryfiles)
+
```
sebsura, line 56: I am not a fan of this.  Why do something manually, when you can automate it ?
```
set(cats_generated_query_files
    ${CMAKE_CURRENT_SOURCE_DIR}/bdb_query_enum_class.h
    ${CMAKE_CURRENT_SOURCE_DIR}/bdb_query_names.inc
    ${CMAKE_CURRENT_SOURCE_DIR}/postgresql_queries.inc
)

# TODO: instead of using a file glob here
#       we should specify each file specifically
#       and give these files as parameter to create_queryfiles
#       i.e. that script should not search for the files itself anymore
file(GLOB cats_input_files CONFIGURE_DEPENDS "dml/????_*")

add_custom_command(
  OUTPUT ${cats_generated_query_files}
  COMMAND ${CMAKE_COMMAND} -E chdir ${CMAKE_CURRENT_SOURCE_DIR}/dml
          ./create_queryfiles.sh
  DEPENDS ${cats_input_files}
  VERBATIM
)

if (NOT HAVE_WIN32)
  add_custom_target(generate-query-files ALL DEPENDS ${cats_generated_query_files})
  add_dependencies(bareossql generate-query-files)
endif()
```

This should cause cmake to automatically regenerate the query files if they change.
Since our CI checks if the source tree is dirty, this will also cause it to fail if they were not pushed.
But this should not often happen, because git will tell you locally that your tree is dirty.

systemtests/tests/upgrade-database-basefiles-not-empty/CMakeLists.txt
sebsura: As this file still exists, I guess it is not fully addressed.

systemtests/tests/upgrade-database-basefiles-not-empty/testrunner
sebsura: As this file still exists, I guess it is not fully addressed.

core/src/filed/accurate.cc
```diff
@@ -311,11 +303,7 @@ bool AccurateCheckFile(JobControlRecord* jcr, FindFilesPacket* ff_pkt)
   /* In Incr/Diff accurate mode, we mark all files as seen
    * When in Full+Base mode, we mark only if the file match exactly */
   if (jcr->getJobLevel() == L_FULL) {
-    if (!status) {
-      // Compute space saved with basefile.
-      jcr->fd_impl->base_size += ff_pkt->statp.st_size;
-      jcr->fd_impl->file_list->MarkFileAsSeen(payload);
-    }
+    if (!status) { jcr->fd_impl->file_list->MarkFileAsSeen(payload); }
```
sebsura: Is it even possible for full jobs to still be accurate ? Where would the old file list even come from ?

core/src/filed/accurate.cc
```diff
@@ -120,19 +120,11 @@ bool AccurateFinish(JobControlRecord* jcr)
   }
 
   if (jcr->accurate && jcr->fd_impl->file_list) {
-    if (jcr->is_JobLevel(L_FULL)) {
-      if (!jcr->rerunning) {
-        retval = jcr->fd_impl->file_list->SendBaseFileList();
-      }
-    } else {
+    if (!jcr->is_JobLevel(L_FULL)) {
```
sebsura, line 123: Does this check still make sense without base files ?

core/src/cats/cats.h
```diff
@@ -901,14 +876,18 @@ class BareosDb : public BareosDbQueryEnum {
   template <SQL_QUERY query, typename... Args>
   void FillQuery(POOLMEM*& storage, Args&&... args)
   {
-    printf_check(queries[static_cast<int>(query)], args...);
+    if constexpr (sizeof...(Args) > 0) {
+      printf_check(queries[static_cast<int>(query)], args...);
+    }
     FillQuery(storage, query, std::forward<Args>(args)...);
```
sebsura, line 881: Im not a fan of this.  If I change a query from
```
SELECT X;
```
to 
```
SELECT X FROM %s
```
then I will not get any warnings, as all function calls currently pass nothing.

Did you run into any problems ? I thought I fixed all the compiler warnings that popped up here.

core/src/cats/dml/0055_match_query2.postgresql
```diff
@@ -1 +0,0 @@
- AND Filename.Name ~ '%s'
```
sebsura, line 1: Are we sure that this is even used ? `Filename.Name` does not seem to be correct to me.

The only occurance I could find with a quick grep is a test that was added in this pr.  Should we just remove this instead ?

core/src/cats/dml/0042_uar_create_temp
```diff
@@ -1,6 +1,5 @@
 CREATE TEMPORARY TABLE
        temp (
-	  JobId INTEGER UNSIGNED NOT NULL, JobTDate BIGINT UNSIGNED, ClientId INTEGER UNSIGNED, Level CHAR, JobFiles INTEGER
-	  UNSIGNED, JobBytes BIGINT UNSIGNED, StartTime TEXT, VolumeName TEXT, StartFile INTEGER UNSIGNED, VolSessionId INTEGER
-	  UNSIGNED, VolSessionTime INTEGER UNSIGNED
+          JobId INTEGER NOT NULL, JobTDate BIGINT, ClientId INTEGER, Level CHAR, JobFiles INTEGER, JobBytes BIGINT, StartTime
+          TEXT, VolumeName TEXT, StartFile INTEGER, VolSessionId INTEGER, VolSessionTime INTEGER
```
sebsura, line 4: Could we use one line per decleration here ? Currently its really hard to see the difference, and doing this should make that much easier for future changes.

sebsura (dismissed):
Looks great.  I just found a few minor things.

core/src/cats/cats.h
sebsura: Maybe as a compromise, we could check that the query does not contain any `%` if `sizeof...(Args) == 0`.

systemtests/tests/upgrade-database-basefiles-not-empty/CMakeLists.txt
pstorz: You are absolutely right, that does not make sense.

systemtests/tests/basefiles-not-empty-upgrade-database/testrunner
```diff
@@ -0,0 +1,61 @@
...
+    FileIndex         INTEGER,
+    BaseJobId         INTEGER,
+    PRIMARY KEY (BaseId)
+);
+
+CREATE INDEX basefiles_jobid_idx ON public.basefiles (JobId);
+
+INSERT INTO public.basefiles (JobId, FileId, FileIndex, BaseJobId)
+VALUES (1, 1, 1, 1);
+SQL
```
arogge, line 43: better, but still not good: it doesn't really test an upgrade from 2250. The other upgrade-test just uses a copy of the old creates sql-file and I don't see any reason not to just copy the 2250 creates into this test.

arogge (changes requested):
I only re-checked the comments that I had. Most of them were properly addressed, but the systemtest still doesn't do a proper upgrade test.

core/src/cats/CMakeLists.txt
arogge: Maybe we just don't touch this at all right now and create a task to move all of this to a pure CMake implementation (i.e. replace `create_queryfiles.sh`) and not have the generated files in the repository anymore?

core/src/tests/sql_query_test.cc
arogge, line 60: First of all, this whole test seems to be pointless as it just checks if a certain query contains a certain string. I'd vote to just remove it, then we don't have to talk about all the other issues I have with it.

core/src/cats/CMakeLists.txt
sebsura: Sure, but I do not see that happening in the near future in that case.  I am fine with just dropping this change all together, I just want to have it completely automated if we change it at all.

core/src/cats/dml/0004_fill_jobhisto
arogge: As `JobHisto` is created like `Job` in the schema, it probably does make sense to copy all the fields. It may not make sense to have that column in the first place, but if it is there I guess filling it is less surprising than not filling it.

core/src/cats/dml/0044_uar_jobid_fileindex_from_dir
```diff
@@ -1,14 +1,13 @@
-# Query to get all files in a directory -- no recursing
-SELECT Job.JobId,
-       File.FileIndex
-FROM Job,
-     File,
-     Path,
-     Client
+# Query to get the latest version of each file in a directory -- no recursing.
+SELECT DISTINCT ON (File.Name) Job.JobId,
```
arogge, line 2: How is this functionally equivalent to the old query?
Instead of returning all `FileIndex` values, we only get one now.
This query orders by `Job.StartTime`, while everything else ordering versions seems to order by Job.JobTDate`.
Last, but not least, `Job.JobId IN (%s)` already selects a client. Thus, joining the client table is only helpful if the JobId list was wrong.

core/src/cats/dml/0048_bvfs_list_files
```diff
@@ -1,44 +1,29 @@
...
+                                  File.FileIndex
+   FROM File
+   JOIN Job USING (JobId)
+   WHERE
+      File.JobId IN (%s)
+      AND File.PathId = %s
+      AND File.Name != ''
+      %s
+   ORDER BY File.Name,
+            Job.StartTime DESC
```
arogge, line 25: Everything else sorts on `JobTDate`.

core/src/cats/dml/0048_bvfs_list_files
```diff
@@ -1,44 +1,29 @@
...
+) AS A
+WHERE A.FileIndex > 0
+LIMIT %)SQL" PRId64 R"SQL(
```
arogge, line 28: Couldn't this just go into the subquery, too.
Because that would allow removing the surrounding query.

core/src/cats/dml/0074_bvfs_ls_special_dirs_3
```diff
@@ -26,5 +29,8 @@ LEFT JOIN (
     FROM File
     WHERE File.Name = ''
       AND File.JobId IN (%s)
+    ORDER BY PathId ASC, JobId DESC
 ) AS DirAttribute
 ON (SpecialDir.PathId = DirAttribute.PathId)
+ORDER BY SpecialDir.PathId ASC, JobId DESC
```
arogge, line 35: is `JobId DESC` actually correct or should we order by `JobTDate`, too?

core/src/cats/dml/create_queryfiles.sh
```diff
@@ -113,3 +109,7 @@ cat >>$QUERY_ENUM_FILE <<EOF
 
 #endif  // BAREOS_CATS_BDB_QUERY_ENUM_CLASS_H_
 EOF
+
+for db in $DATABASES; do
+  queryincludefile=$(get_query_include_filename $db)
+done
```
arogge, line 115: what is this supposed to do?

core/src/cats/dml/create_queryfiles.sh
arogge, line 49: as we only do PostgreSQL this is static now

core/src/cats/dml/create_queryfiles.sh
```diff
@@ -63,48 +69,38 @@ EOF
 
 for db in $DATABASES; do
   DB=${db^}
-  queryincludefile=$(get_query_include_filename "${db}")
-  : >"${queryincludefile}"
-  print_note >>"${queryincludefile}"
+  queryincludefile=$(get_query_include_filename $db)
+  >$queryincludefile
+  print_note >>$queryincludefile
 done
```
arogge, line 75: pointless loop, only PostgreSQL is done

core/src/cats/dml/create_queryfiles.sh
```diff
@@ -63,48 +69,38 @@ EOF
...
-      # remove comments and empty lines
-      sed -r -e "/^#/d" -e "/^$/d" <"${queryfile}"
-      printf ')SQL",\n'
-    } >>"${queryincludefile}"
+    if [ "$i" -gt 0 ]; then printf '\n' >>$queryincludefile; fi
+    printf '/* %s */\nR"SQL(' "$queryfile" >>$queryincludefile
+    # remove comments and empty lines
+    sed -r -e "/^#/d" -e "/^$/d" <"$queryfile" >>"$queryincludefile"
+    printf ')SQL",\n' >>$queryincludefile
   done
```
arogge, line 96: same here, no point in looping

core/src/filed/accurate.cc
arogge: At least it doesn't make sense to try and send a DeletedList on a Full.

core/src/filed/accurate.cc
arogge, line 298: Can we check if `file_changed` can be `false` with a Full job?
If not, we can (and should) simplify this.

docs/manuals/CMakeLists.txt
```diff
@@ -302,7 +302,10 @@ add_custom_target(
   COMMENT
     "Comparing json files in \"${PROJECT_SOURCE_DIR}/source/include/autogenerated\" with git repo"
 )
-add_dependencies(check-git check-freebsd-pkglist)
+if(TARGET check-freebsd-pkglist)
+  add_dependencies(check-docs-git check-freebsd-pkglist)
+endif()
+add_dependencies(check-git check-docs-git)
```
arogge, line 309: as discussed above, we shouldn't touch this part right now.

CMakeLists.txt
```diff
@@ -123,6 +123,7 @@ add_custom_target(
   uninstall "${CMAKE_COMMAND}" -P
             "${CMAKE_CURRENT_SOURCE_DIR}/cmake/uninstall.cmake"
 )
+add_custom_target(check-git)
```
arogge, line 126: we don't want this change currently.

arogge (changes requested):
Still quite some findings that were not (yet) addressed.

core/src/cats/dml/0044_uar_jobid_fileindex_from_dir
pstorz: You were right that the  DISTINCT ON  rewrite was not functionally equivalent. It changed the query from returning all matching  JobId / FileIndex  entries to returning only one entry per filename, which could alter directory-restore behavior. 

However this probably would be an optimization for the future as it makes probably sense to only restore the newest version of each file instead of multiple versions.

The query still uses the original implicit joins and does not change the  StartTime / JobTDate  ordering, because there is no version-selection ordering in the restored query.

core/src/cats/dml/0048_bvfs_list_files
pstorz: Agreed. The BVFS file-list query now orders versions by:

ORDER BY File.Name,
         Job.JobTDate DESC,
         Job.JobId DESC

core/src/cats/dml/0048_bvfs_list_files
pstorz: Moving  FileIndex > 0  into the subquery would change BVFS behavior.

The subquery selects the newest catalog entry for each filename. A newer directory entry can have  FileIndex = 0 , while an older regular-file entry for the same name has  FileIndex > 0 . If the filter moved inside, the older file could be selected and exposed incorrectly.

core/src/cats/dml/0074_bvfs_ls_special_dirs_3
pstorz: JobTDate  should be the primary ordering field because it represents the backup timestamp used by the other recent-version queries.  JobId DESC  is retained as a deterministic tie-breaker when multiple jobs have the same  JobTDate .

The query now orders special-directory entries by:

ORDER BY SpecialDir.PathId ASC,
         JobTDate DESC,
         JobId DESC

core/src/cats/dml/create_queryfiles.sh
pstorz: Agreed

core/src/filed/accurate.cc
pstorz: Agreed. Full jobs must not send a  DeletedList , because a Full backup is not using the accurate list to determine deletions in the same way as Incremental/Differential jobs.

 AccurateFinish()  now sends the deleted list only for non-Full jobs:

if (!jcr->is_JobLevel(L_FULL)) {
  retval = jcr->fd_impl->file_list->SendDeletedList();
}

The Full-job accurate list is still cleaned up normally afterward.

core/src/filed/accurate.cc
pstorz: Yes,  file_changed  can be  false  for a Full accurate job.  AccurateCheckFile()  compares the current file against the previous accurate payload and the configured accurate options. An unchanged file therefore remains unchanged, and the function returns  false .

The Full/non-Full distinction must remain:

• Full jobs mark only unchanged files as seen so they can be reused/skipped.
• Incremental/Differential jobs mark every visited file as seen so changed files are not later reported as deleted.

The earlier simplification was reverted because it caused Incremental/Differential accurate-backup regressions. No further change is required.

docs/manuals/CMakeLists.txt
pstorz: Agreed, was reverted.

CMakeLists.txt
pstorz: Agreed, reverted.

systemtests/tests/basefiles-not-empty-upgrade-database/testrunner
pstorz: Ok, adapted to uses a sel-contained 2250 catalog.

PR #2647, opened 2026-05-01
core: fixup fadvise usage

clockfort:
- Reduce page cache thrashing
- Possibly improve despool perf/reduce shoe-shining via larger readahead

My environment's spool/despool server has a lot of memory pressure, and pushing 12TiB of data through the page cache twice to do LTO-8 tape spool/despools does not help the problem. We tested this change by ld_preload shimming bareos' fadvise calls ( https://gist.github.com/clockfort/120a5ea7f92915a2547f86227b8f2a40 ) and it greatly improved memory pressure / overall perf.

There's probably also some other places that could use FADV_SEQUENTIAL, if I poked my head around more, or other low hanging fruit to avoid thrashing the page caches unnecessarily of the servers being backed up as well, which could make backups less disruptive to whatever main application is actively running on those servers.

As a flag, on old 2006 linux (<2.6.18) WILLNEED and NOREUSE were treated the same; only in linux 6.3+ (2023) does it have the good page marking behavior. ([manpage](https://man7.org/linux/man-pages/man2/posix_fadvise.2.html#:~:text=POSIX_FADV_NOREUSE%0A%20%20%20%20%20%20%20%20%20%20%20%20%20%20The%20specified,streaming%20large%20files.))
Old FreeBSD before like ?freebsd10? (2014) this change would be detrimental, as FreeBSD previously very questionably implemented NOREUSE by just making things O_DIRECT with no caching at all, which is undesirable for this application.

If we think it's necessary, I could add some detection for linux <6.3 to have the old behavior, though I kind of hate it because it's just extra code kludge that will stick around forever.

#### Please check

##### General

##### Source code quality

##### Tests

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -46,6 +46,7 @@ Carlos A. Molina G
 Carlos Rodrigues
 Carsten Paeth
 Chris Lee
+Chris Lockfort
 Chris Matthews
 Christer Fletcher
 Christian Eichelmann
--- a/core/src/dird/catreq.cc
+++ b/core/src/dird/catreq.cc
@@ -735,8 +735,8 @@ bool DespoolAttributesFromFile(JobControlRecord* jcr, const char* file)
     goto bail_out;
   }
 
-#if defined(HAVE_POSIX_FADVISE) && defined(POSIX_FADV_WILLNEED)
-  posix_fadvise(spool_fd, 0, 0, POSIX_FADV_WILLNEED);
+#if defined(HAVE_POSIX_FADVISE) && defined(POSIX_FADV_NOREUSE)
+  posix_fadvise(spool_fd, 0, 0, POSIX_FADV_NOREUSE);
 #endif
 
   while ((nbytes = read(spool_fd, (char*)&pktsiz, sizeof(int32_t)))
--- a/core/src/findlib/bfile.cc
+++ b/core/src/findlib/bfile.cc
@@ -1048,11 +1048,11 @@ int bopen(BareosFilePacket* bfd,
   bfd->win32Decomplugin_private_context.bIsInData = false;
   bfd->win32Decomplugin_private_context.liNextHeader = 0;
 
-#  if defined(HAVE_POSIX_FADVISE) && defined(POSIX_FADV_WILLNEED)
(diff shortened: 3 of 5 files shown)


sebsura:
I am suprised that you run into these issues with `WILLNEED`. The man page says
```
The amount of data read may be decreased by the kernel depending on virtual memory load.  (A few
              megabytes will usually be fully satisfied, and more is rarely useful.)
```

I will have to try it out on a big spool file and see what happens.

clockfort:
To be honest with you this spool/despool tape host box is under more load than it should be, but for at least backup clients I think the overall concept of "backing up the server is never the primary purpose of a server, serving something else is, and that should take priority" is why I think a change like this would be good; it's more about making bareos, an application that regularly pushes many TiB of data through the page cache more of a friendly neighbor to other programs

and fully ignoring willneed vs noreuse, a lot of bareos workloads are just a perfect example for sequential hinting and longer readaheads

sebsura:
Sorry for taking so long to come back to you.

I ran some backups with my standard fileset (~900k files, ~13GB data; consists of git checkout of chromium, firefox and the linux-kernel) and got the following results:

| | Master | PR-2647 | Delta |
|-|---------|---|---|
| No Spooling | 2:31| 2:44 | +8.6% |
| Spooling | 2:40 | 2:54 | +8.8% |

This is not great news.  I have not been able to confirm the reduced memory pressure.  Do you have a good idea on how i could check that ?

sebsura:
Seems like most of the time loss comes from the bfile changes.  Maybe its enough to just change it on the sd/dir for now ?

clockfort:
I'd have to know more about the test setup (linux version, filesystem type) but I'm not surprised that it's a bit slower, that's more or less the point, that it's more conservative with memory usage.

Yes, I'd accept just the sd/dir changes, or at least keeping in the sequential hinting everywhere because that should purely be a positive or neutral in all cases.

We could also gate this behaviour behind a config flag if we wanted, but I didn't want to introduce more config entropy / code coverage nightmares.

sebsura:
My kernel is `7.0.12-201.fc44.x86_64`, fs is ext4, but the spool file was on a ramdisk.
Ill test a bit more and also check additional `SEQUENTIAL` hints.

PR #2654, opened 2026-05-06
tray-monitor: animate icon as 3D shield

pstorz:

- [ ] Add qt svg as runtime dependency

#### Please check

##### General

##### Source code quality


core/src/qt-tray-monitor/systemtrayicon.cc
```diff
@@ -21,27 +21,71 @@
...
+#include <algorithm>
+#include <cmath>
+
+namespace {
+const int kNormalIcon = 0;
+const int kErrorIcon = 2;
+const int kAnimationIntervalMs = 120;
+const int kAnimationFrameCount = 12;
+const qreal kMinimumWidthScale = 0.15;
+const qreal kPi = 3.14159265358979323846;
```
sebsura: Whats wrong with `M_PI` and so on ? See https://doc.qt.io/qt-6/qtmath.html

core/src/qt-tray-monitor/systemtrayicon.cc
```diff
@@ -21,27 +21,71 @@
...
+const int kShadowBlueBlue = 156;
+const int kHighlightBlueRed = 48;
+const int kHighlightBlueGreen = 150;
+const int kHighlightBlueBlue = 230;
+const int kShadowGreenRed = 12;
+const int kShadowGreenGreen = 120;
+const int kShadowGreenBlue = 30;
+const int kHighlightGreenRed = 72;
+const int kHighlightGreenGreen = 220;
+const int kHighlightGreenBlue = 96;
```
sebsura: Giving numbers a name is great.  But its not clear at all what these numbers stand for.  Are these rgb values, some index, etc.  A comment explaining what these constants are for with an example would be great.

core/src/qt-tray-monitor/systemtrayicon.cc
```diff
@@ -21,27 +21,71 @@
...
+}  // namespace
 
 SystemTrayIcon::SystemTrayIcon(QMainWindow* mainWindow)
-    : QSystemTrayIcon(mainWindow), iconIdx(0), timer(new QTimer(this))
+    : QSystemTrayIcon(mainWindow)
+    , iconIdx(kNormalIcon)
+    , animationFrameIdx(0)
+    , animationRequested(false)
+    , normalIcon(QIcon(createShieldPixmap(0.0, 1.0, false)))
+    , errorIcon(":/images/W.png")
```
sebsura: Why did we give `kTrayAnimationIcon` a name, but not this path ?

core/src/qt-tray-monitor/systemtrayicon.cc
```diff
@@ -50,24 +94,232 @@ SystemTrayIcon::~SystemTrayIcon() { timer->stop(); }
...
+    frames << QIcon(createShieldPixmap(angle, widthScale, true));
+  }
+
+  return frames;
+}
+
+QPixmap SystemTrayIcon::createBaseShieldPixmap(bool recolor_green) const
+{
+  QPixmap basePixmap(kTrayAnimationIcon);
+  if (basePixmap.isNull() || !recolor_green) { return basePixmap; }
```
sebsura: Why do we load the file repeatedly ? To minimise errors (what happens when the first few loads work, but then they do not for some reason ?) we should try to load resources _once_, and then copy them if necessary.

In this specific case `basePixmap.toImage()` will create a copy anyways already, so we should only load the base image _once_, probably outside of this function, in some kind of setup.  Maybe the system tray class could take it as a constructor parameter.

core/src/qt-tray-monitor/systemtrayicon.cc
```diff
@@ -50,24 +94,232 @@ SystemTrayIcon::~SystemTrayIcon() { timer->stop(); }
...
+          || color.hsvSaturation() <= kShieldSaturationThreshold) {
+        continue;
+      }
+
+      QColor recolored;
+      recolored.setHsv(kShieldGreenHue, color.hsvSaturation(), color.value(),
+                       color.alpha());
+      scanLine[x] = recolored.rgba();
     }
   }
```
sebsura: Why does `createBaseShieldPixmap` colour stuff green ?

This function should be called something like `addGreenShading()` and only be done when this is actually necessary. It should take in the pixmap as `const&` as it needs to create a copy anyways.
Also now that we probably seperated concerns, this doesnt even need to be a class function anymore, but can be a simple free function `QPixmap addGreenShading(const Pixmap&)` (please find a better fitting name).

core/src/qt-tray-monitor/systemtrayicon.cc
```diff
@@ -50,24 +94,232 @@ SystemTrayIcon::~SystemTrayIcon() { timer->stop(); }
...
+  const int shadowRed = use_green_palette ? kShadowGreenRed : kShadowBlueRed;
+  const int shadowGreen
+      = use_green_palette ? kShadowGreenGreen : kShadowBlueGreen;
+  const int shadowBlue = use_green_palette ? kShadowGreenBlue : kShadowBlueBlue;
+  const int highlightRed
+      = use_green_palette ? kHighlightGreenRed : kHighlightBlueRed;
+  const int highlightGreen
+      = use_green_palette ? kHighlightGreenGreen : kHighlightBlueGreen;
+  const int highlightBlue
+      = use_green_palette ? kHighlightGreenBlue : kHighlightBlueBlue;
```
sebsura: this reads really bad.  Why not just create a struct with the relevant members, and then have one instant of the struct for the green values and one instance of the struct for the normal values ?

We could even name the struct "palette" and the function could accept any instead of just green/default.

core/src/qt-tray-monitor/systemtrayicon.cc
```diff
@@ -50,24 +94,232 @@ SystemTrayIcon::~SystemTrayIcon() { timer->stop(); }
...
+
+  return QPixmap::fromImage(image);
+}
+
+QPixmap SystemTrayIcon::createShieldPixmap(qreal angle,
+                                           qreal widthScale,
+                                           bool use_green_palette) const
+{
+  QPixmap basePixmap(createBaseShieldPixmap(use_green_palette));
+  if (basePixmap.isNull()) { return QPixmap(); }
```
sebsura: This should just not be possible if we set up the class correctly.

core/src/qt-tray-monitor/systemtrayicon.cc
```diff
@@ -50,24 +94,232 @@ SystemTrayIcon::~SystemTrayIcon() { timer->stop(); }
...
+  painter.drawPixmap(0, 0, shadowOverlay);
+  painter.drawPixmap(0, 0, shieldPixmap);
+
+  QPixmap depthOverlay(framePixmap.size());
+  depthOverlay.fill(Qt::transparent);
+  {
+    QPainter depthPainter(&depthOverlay);
+    QLinearGradient verticalGradient(0, 0, 0, framePixmap.height());
+    verticalGradient.setColorAt(0.0, QColor(highlightRed, highlightGreen,
+                                            highlightBlue, kShieldDepthAlpha));
```
sebsura: If only there was a struct that would bundle together r, g, b values of our highlight color.  Please use QRgb or QColor and not 3 different variables.  This is not only the case for this but also for the global variables ...

core/src/qt-tray-monitor/systemtrayicon.cc
```diff
@@ -50,24 +94,232 @@ SystemTrayIcon::~SystemTrayIcon() { timer->stop(); }
...
     }
   }
+
+  return QPixmap::fromImage(image);
+}
+
+QPixmap SystemTrayIcon::createShieldPixmap(qreal angle,
+                                           qreal widthScale,
+                                           bool use_green_palette) const
+{
```
sebsura: This function needs some documentation.  What is it supposed to do ?
#. Rotate by `angle`, but in which direction ?
#. Scale the width somehow, but how ? Does it scale only the width, i.e. squish the image ?
#. Optionally, apply some kind of hue ?

Are these steps interdependent on each other  or could we maybe split up the logic here ?

core/src/qt-tray-monitor/systemtrayicon.cc
```diff
@@ -50,24 +94,232 @@ SystemTrayIcon::~SystemTrayIcon() { timer->stop(); }
...
+  QPixmap basePixmap(createBaseShieldPixmap(true));
+
+  if (basePixmap.isNull()) { return frames; }
+
+  for (int frame = 0; frame < kAnimationFrameCount; ++frame) {
+    const qreal angle
+        = (static_cast<qreal>(frame) / kAnimationFrameCount) * 2.0 * kPi;
+    const qreal cosine = std::cos(angle);
+    const qreal widthScale = std::max(kMinimumWidthScale, std::abs(cosine));
+    frames << QIcon(createShieldPixmap(angle, widthScale, true));
```
sebsura: Why does createShildPixmap take both a width scale _and_ an angle, when one already determines the other ?
This just invites mixups.  createShieldPixmap should just take the angle and compute the width scale itself.

core/src/qt-tray-monitor/systemtrayicon.cc
```diff
@@ -50,24 +94,232 @@ SystemTrayIcon::~SystemTrayIcon() { timer->stop(); }
...
+    shadowPainter.drawPixmap(shadowX, kShieldShadowYOffset, shieldPixmap);
+    shadowPainter.setCompositionMode(QPainter::CompositionMode_SourceIn);
+    shadowPainter.fillRect(shadowOverlay.rect(),
+                           QColor(shadowRed / 2, shadowGreen / 2,
+                                  shadowBlue / 2, kShieldShadowAlpha));
+  }
+  painter.drawPixmap(0, 0, shadowOverlay);
+  painter.drawPixmap(0, 0, shieldPixmap);
+
+  QPixmap depthOverlay(framePixmap.size());
```
sebsura: It would be nice if each step could have a small comment describing what it is trying to achieve.

core/src/qt-tray-monitor/systemtrayicon.cc
```diff
@@ -50,24 +94,232 @@ SystemTrayIcon::~SystemTrayIcon() { timer->stop(); }
...
+      sideGradient.setColorAt(0.55,
+                              QColor(shadowRed, shadowGreen, shadowBlue, 0));
+      sideGradient.setColorAt(
+          1.0, QColor(shadowRed, shadowGreen, shadowBlue, sideShading));
+    }
+    sidePainter.fillRect(sideOverlay.rect(), sideGradient);
+    sidePainter.setCompositionMode(QPainter::CompositionMode_DestinationIn);
+    sidePainter.drawPixmap(0, 0, shieldPixmap);
+    painter.drawPixmap(0, 0, sideOverlay);
+  }
```
sebsura: At least on my monitor, this is basically unnoticable.  What do you think @pstorz ? If you cannot notice it either, then we should either remove it or boost it.

core/src/qt-tray-monitor/systemtrayicon.cc
```diff
@@ -50,24 +94,232 @@ SystemTrayIcon::~SystemTrayIcon() { timer->stop(); }
...
+  animationFrameIdx = (animationFrameIdx + 1) % animationIcons.size();
+  updateIcon();
 }
 
 void SystemTrayIcon::animateIcon(bool on)
 {
-  if (on) {
+  animationRequested = on;
+  if (animationRequested && iconIdx != kErrorIcon) {
+    animationFrameIdx = 0;
```
sebsura: Why are we resetting the index without checking if the animation is not already currently running ? This causes extremely jerky animations.

sebsura (changes requested):
The points brought up by coderabbit are also correct, but its even worse.  Currently there is _no_ transition at all between "rotating image" and "still image", so when your backup finishes while the logo is currently spinning, it just snaps back to the default position.

When the backup is done, it should continue spinning until it reaches the default position and it should stop there.

The color should probably also transition from blue to green/red (and back) in multiple frames as well (or it could happen during the flip), but this is not as important.

core/src/qt-tray-monitor/systemtrayicon.cc
sebsura: better yet: use actual types (i.e. qrgb and so on).

core/src/qt-tray-monitor/systemtrayicon.cc
```diff
@@ -50,24 +94,232 @@ SystemTrayIcon::~SystemTrayIcon() { timer->stop(); }
...
   iconIdx = icon;
-  setIcon(QIcon(icons[icon]));
+  if (iconIdx == kErrorIcon) {
+    timer->stop();
+  } else if (animationRequested && !timer->isActive()) {
+    timer->start();
+  }
+
+  updateIcon();
 }
```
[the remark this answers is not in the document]
pstorz: Addressed

core/src/qt-tray-monitor/systemtrayicon.cc
pstorz: addressed

core/src/qt-tray-monitor/systemtrayicon.cc
pstorz: addressed

core/src/qt-tray-monitor/systemtrayicon.cc
pstorz: Addressed

core/src/qt-tray-monitor/systemtrayicon.cc
pstorz: addressed

core/src/qt-tray-monitor/systemtrayicon.cc
pstorz: addressed

core/src/qt-tray-monitor/systemtrayicon.cc
pstorz: addressed

core/src/qt-tray-monitor/systemtrayicon.cc
pstorz: addressed

core/src/qt-tray-monitor/systemtrayicon.cc
pstorz: Good  point. It was even worse, the shadow also rotated with the image itself. The shadow is now fixed.

PR #2664, opened 2026-05-13
filemarks: use immediate filemarks when writing to tape, set default maxfilesize to 100GB

pstorz:

When writing to tape we now use immediate filemarks every `Maximum File Size` bytes
instead of using the former non-immediate filemarks.
The old behaviour forced the drive to flush all buffers which had a big negative performance impact on
the write performance.

Also the `d_flush()` call now is no longer is a no-op but really flushes data.

* [ ] update default-value for `Maximum File Size` to `100 GB`

#### Please check

##### General

##### Source code quality

##### Tests


arogge (changes requested):
IMO that's way to much change. We can just use the non-waiting filemarks everywhere as long as we flush properly where it is needed.

core/src/stored/backends/generic_tape_device.cc
```diff
@@ -328,54 +328,103 @@ bool generic_tape_device::offline()
...
+/**
+ * Write an end of file on the device
+ *
+ * Returns: true on success
+ *          false on failure
+ */
+bool generic_tape_device::weof(int num)
+{
+  return WriteFilemark(this, num, true);
+}
```
arogge: this refactoring doesn't make sense as the function is only called once. It also obscures the actual change.

core/src/stored/backends/generic_tape_device.cc
```diff
@@ -328,54 +328,103 @@ bool generic_tape_device::offline()
...
+  if (::fsync(fd) < 0 && errno != EINVAL && errno != ENOTSUP
+      && errno != ENOSYS) {
+    BErrNo be;
+    Mmsg2(errmsg, T_("fsync error on %s. ERR=%s.\n"), prt_name, be.bstrerror());
+    return false;
+  }
+#endif
+
+  return true;
+}
```
arogge: We have `generic_tape_device`, `win32_tape_device` and `unix_tape_device`.
`win32_tape_device` has its own implementation of `d_flush()`, thus this should go into `unix_tape_device` instead of `generic_tape_device`.
This would also remove the need for that `HAVE_WIN32` ifdef.

core/src/stored/backends/generic_tape_device.cc
```diff
@@ -328,54 +328,103 @@ bool generic_tape_device::offline()
...
+  if (d_ioctl(fd, MTIOCTOP, (char*)&mt_com) < 0) {
+    BErrNo be;
+    clrerror(mt_com.mt_op);
+    Mmsg2(errmsg, T_("ioctl MTNOP error on %s. ERR=%s.\n"), prt_name,
+          be.bstrerror());
+    return false;
+  }
+
+#if !defined(HAVE_WIN32)
+  if (::fsync(fd) < 0 && errno != EINVAL && errno != ENOTSUP
```
arogge: all Unix tape drivers I know ignore `fsync()`, so this seems pointless to me.

core/src/stored/btape.cc
```diff
@@ -614,6 +614,9 @@ static void weofcmd()
   if (!g_dev->weof(num)) {
     Pmsg1(0, T_("Bad status from weof. ERR=%s\n"), g_dev->bstrerror());
     return;
+  } else if (!g_dev->d_flush(g_dcr)) {
+    Pmsg1(0, T_("Bad status from flush. ERR=%s\n"), g_dev->bstrerror());
+    return;
```
arogge: This will add implicit flushes when writing a file-mark with `btape`, which might render it useless for performance testing.

core/src/stored/btape.cc
```diff
@@ -1533,7 +1536,7 @@ static int autochanger_test()
   /* Start with sleep_time 0 then increment by 30 seconds if we get
    * a failure. */
   Bmicrosleep(sleep_time, 0);
-  if (!g_dev->rewind(g_dcr) || !g_dev->weof(1)) {
+  if (!g_dev->rewind(g_dcr) || !g_dev->weof(1) || !g_dev->d_flush(g_dcr)) {
```
arogge: same here

core/src/stored/btape.cc
```diff
@@ -1545,7 +1548,7 @@ static int autochanger_test()
     Pmsg1(0, T_("Rewound %s\n"), g_dev->print_name());
   }
 
-  if (!g_dev->weof(1)) {
+  if (!g_dev->weof(1) || !g_dev->d_flush(g_dcr)) {
```
arogge: and here

core/src/tests/CMakeLists.txt
```diff
@@ -345,8 +345,16 @@ if(NOT client-only)
...
   )
   bareos_add_test(
-    test_config_parser_sd LINK_LIBRARIES Bareos::SD Bareos::LibSD Bareos::Lib
-                                         GTest::gtest_main
+    test_config_parser_sd
+    LINK_LIBRARIES Bareos::SD Bareos::LibSD Bareos::Lib GTest::gtest_main
+  )
+  if(TARGET bareossd-file)
+    add_dependencies(test_config_parser_sd bareossd-file)
+  endif()
```
arogge: why is this suddenly required without any change to test_config_parser_fd?

core/src/tests/test_generic_tape_device.cc
```diff
@@ -0,0 +1,119 @@
...
+
+TEST(GenericTapeDevice, tape_flush_issues_mtnop)
+{
+  TestTapeDevice dev{/*immediate_supported=*/true};
+
+  ASSERT_TRUE(dev.d_flush(nullptr));
+  ASSERT_EQ(dev.operations.size(), 1U);
+  EXPECT_EQ(dev.operations[0], MTNOP);
+}
+
```
arogge: That test doesn't seem to be too useful to me.

arogge (changes requested):
I like this version a lot more than the previous one I looked at. But I think there's still a few things that need to be addressed.

core/src/stored/backends/generic_tape_device.cc
pstorz: agreed

core/src/stored/backends/generic_tape_device.cc
pstorz: Good catch

core/src/tests/test_generic_tape_device.cc
pstorz: Agreed.

core/src/stored/backends/generic_tape_device.cc
pstorz: Agreed

core/src/tests/CMakeLists.txt
pstorz: removed

core/src/stored/btape.cc
pstorz: agreed

core/src/stored/btape.cc
pstorz: agreed

core/src/stored/btape.cc
pstorz: agreed

core/src/stored/backends/generic_tape_device.cc
```diff
@@ -328,12 +328,6 @@ bool generic_tape_device::offline()
   return true;
 }
 
-/**
- * Write an end of file on the device
- *
- * Returns: true on success
- *          false on failure
- */
```
arogge, line 336: what was wrong with this comment?

core/src/stored/backends/generic_tape_device.cc
```diff
@@ -357,8 +351,20 @@ bool generic_tape_device::weof(int num)
...
   ClearEot();
   mt_com.mt_op = MTWEOF;
+#if defined(MTWEOFI)
+  mt_com.mt_op = MTWEOFI;
+#endif
   mt_com.mt_count = num;
   status = d_ioctl(fd, MTIOCTOP, (char*)&mt_com);
+#if defined(MTWEOFI)
+  if (status < 0 && mt_com.mt_op == MTWEOFI
+      && (errno == ENOTTY || errno == ENOSYS)) {
```
arogge: that is not the proper `errno` for this case.
Maybe `EIO` or `EINVAL` are correct, but I can't find anything in the docs or kernel driver sources right now.

core/src/stored/backends/generic_tape_device.h
```diff
@@ -2,7 +2,7 @@
    BAREOS® - Backup Archiving REcovery Open Sourced
 
    Copyright (C) 2014-2014 Planets Communications B.V.
-   Copyright (C) 2014-2023 Bareos GmbH & Co. KG
+   Copyright (C) 2014-2026 Bareos GmbH & Co. KG
```
arogge, line 5: file unchanged, no year change

core/src/stored/backends/generic_tape_device.cc
arogge: I double-checked and if I don't misunderstand the st driver in the Linux kernel, the response for a failed SCSI command in this case is `EIO`.

core/src/stored/backends/unix_tape_device.cc
```diff
@@ -108,6 +109,29 @@ ssize_t unix_tape_device::d_read(int t_fd, void* buffer, size_t count)
...
+
+  mt_com.mt_op = MTNOP;
+  mt_com.mt_count = 1;
+  if (d_ioctl(fd, MTIOCTOP, (char*)&mt_com) < 0) {
+    BErrNo be;
+    clrerror(mt_com.mt_op);
+    Mmsg2(errmsg, T_("ioctl MTNOP error on %s. ERR=%s.\n"), prt_name,
+          be.bstrerror());
+    return false;
+  }
```
arogge, line 130: Sadly, this only flushes the kernel buffer to the device, but not the device buffer onto the tape.
Technically, the only way to force such a buffer flush this is to write a non-immediate ("normal") filemark or a setmark, which we both don't want when we flush.

arogge (changes requested):
We may need some more changes and we have to decide whether or not we require the tape drive buffer to be flushed to the tape by `d_flush()`. In that case we would have to do a bit more than what we have right now, as we cannot reliably flush the tape drive buffer without writing a filemark.

PR #2669, opened 2026-05-16
sd reservation: fix some problems that can lead to endless loops

pstorz:
#### Please check

##### General

##### Source code quality

##### Tests

### DIFF
--- a/core/src/stored/askdir.cc
+++ b/core/src/stored/askdir.cc
@@ -52,6 +52,8 @@ namespace storagedaemon {
 
 static const int debuglevel = 50;
 static pthread_mutex_t vol_info_mutex = PTHREAD_MUTEX_INITIALIZER;
+static constexpr std::size_t kMaxAppendVolumeLookups = 200;
+static constexpr std::size_t kMaxUnwantedVolumesLength = 32 * 1024;
 
 /* Requests sent to the Director */
 inline constexpr const char Find_media[]
@@ -225,36 +227,16 @@ bool StorageDaemonDeviceControlRecord::DirFindNextAppendableVolume()
 
     PmStrcpy(unwanted_volumes, "");
 
-    std::size_t device_count = 0;
+    for (std::size_t vol_index = 1; vol_index <= kMaxAppendVolumeLookups;
+         vol_index++) {
+      if (strlen(unwanted_volumes.c_str()) >= kMaxUnwantedVolumesLength) {
+        Dmsg1(debuglevel,
+              "Stopping append volume lookup after unwanted list reached %zu "
+              "bytes.\n",
+              strlen(unwanted_volumes.c_str()));
+        break;
+      }
 
-    {
-      ResLocker _{my_config};
-
-      BareosResource* found_dev = nullptr;
-      foreach_res (found_dev, R_DEVICE) { device_count += 1; }
-    }
-
-    if (device_count == 0) {
-      Emsg0(M_ERROR, 0,
-            "Trying to find a volume, but there are apparently no devices.");
-      // not sure what happened here, but this should hopefully be enough
-      device_count = 100;
-    }
-
-    // x >> 3 == x / 8 ~ 10% * x
-    // there should never be a need to ask for more volumes than there are
-    // devices because each device should only be allowed to reserve one volume,
-    // but we add some headroom here in case some recommended volume is broken
-    // or in case some devices are currently reserving multiple volumes (maybe
-    // they are currently switching volumes)
-    std::size_t ask_limit = device_count + (device_count >> 3);
-
-    Dmsg0(400, "device count = %llu => ask limit = %llu\n",
-          static_cast<long long unsigned>(device_count),
-          static_cast<long long unsigned>(ask_limit));
-
-    // use <= because we start counting at 1
-    for (std::size_t vol_index = 1; vol_index <= ask_limit; vol_index++) {
       BashSpaces(media_type);
       BashSpaces(pool_name);
       BashSpaces(unwanted_volumes.c_str());
--- a/core/src/stored/askdir.cc
+++ b/core/src/stored/askdir.cc
@@ -201,8 +201,8 @@ bool StorageDaemonDeviceControlRecord::DirGetVolumeInfo(
  * Returns: true  on success dcr->VolumeName is volume
  *                reserve_volume() called on Volume name
  *          false on failure dcr->VolumeName[0] == 0
- *                also sets dcr->FoundInUse if at least one
- *                in use volume was found.
+ *                also sets dcr->AppendVolumeBusy if at least one
+ *                appendable volume was found busy on another device.
  *
  * Volume information returned in dcr
  */
@@ -221,6 +221,7 @@ bool StorageDaemonDeviceControlRecord::DirFindNextAppendableVolume()
   with_volume_lock([&] {
     lock_mutex(vol_info_mutex);
     ClearFoundInUse();
+    ClearAppendVolumeBusy();
 
     PmStrcpy(unwanted_volumes, "");
 
@@ -286,9 +287,7 @@ bool StorageDaemonDeviceControlRecord::DirFindNextAppendableVolume()
           goto get_out;
         } else {
           Dmsg1(debuglevel, "Volume %s is in use.\n", VolumeName);
-
-          // If volume is not usable, it is in use by someone else
-          SetFoundInUse();
+          if (IsAppendVolumeBusyOnAnotherDevice()) { SetAppendVolumeBusy(); }
           continue;
         }
       }
--- a/core/src/stored/askdir.cc
+++ b/core/src/stored/askdir.cc
@@ -46,13 +46,14 @@
 #include "lib/berrno.h"
 #include "lib/bsock.h"
 #include "lib/serial.h"
+#include <algorithm>
 #include <cinttypes>
 
 namespace storagedaemon {
 
 static const int debuglevel = 50;
 static pthread_mutex_t vol_info_mutex = PTHREAD_MUTEX_INITIALIZER;
-static constexpr std::size_t kMaxAppendVolumeLookups = 200;
+static constexpr std::size_t kMinAppendVolumeLookups = 200;
 static constexpr std::size_t kMaxUnwantedVolumesLength = 32 * 1024;
 
 /* Requests sent to the Director */
@@ -217,9 +218,9 @@ bool StorageDaemonDeviceControlRecord::DirFindNextAppendableVolume()
   Dmsg2(debuglevel, "DirFindNextAppendableVolume: reserved=%d Vol=%s\n",
         IsReserved(), VolumeName);
 
-  /* Try the twenty oldest or most available volumes. Note,
-   * the most available could already be mounted on another
-   * drive, so we continue looking for a not in use Volume. */
+  /* Ask the director for append candidates, oldest or most available first.
+   * The most available one could already be mounted on another drive, so we
+   * keep asking for the next one until we find a volume we can use. */
   with_volume_lock([&] {
     lock_mutex(vol_info_mutex);
     ClearFoundInUse();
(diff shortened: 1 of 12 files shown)


bruno-at-bareos:
Testing since a week, with more than 4000 jobs done, there was not storage error nor lock-down.
Also testing same configuration during a week with 6000 jobs done with master code. There's was two time error on reserving a tape drive leading to a deadlock between drives.

core/src/tests/sd_reservation.cc
```diff
@@ -40,12 +40,15 @@
 #include "lib/edit.h"
 #include "lib/parse_conf.h"
 #include "stored/device_control_record.h"
+#include "stored/sd_device_control_record.h"
+#include "stored/acquire.h"
 #include "stored/stored_jcr_impl.h"
```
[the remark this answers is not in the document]
bruno-at-bareos: @pstorz might make sense what the rabbit want ;-)

core/src/stored/wait.cc
```diff
@@ -41,9 +42,33 @@
...
+
+void SetWaitForDeviceTimeoutForTesting(int timeout_in_seconds)
+{
+  wait_for_device_timeout = timeout_in_seconds >= 0 ? timeout_in_seconds : 0;
+}
+
+void ResetWaitForDeviceTimeoutForTesting()
+{
+  wait_for_device_timeout = kDefaultWaitForDeviceTimeout;
+}
```
sebsura, line 71: Please do not add mutable global variables for no reason.  You can do this by simply adding an additional parameter to WaitForDevice.  Its ok to give that one a default value, so we do not have to touch everything.

```C++
bool WaitForDevice(JobControlRecord* jcr,
                   int& retries,
                   std::chrono::seconds max_wait_duration = std::chrono::seconds(60));
```

core/src/stored/wait.cc
```diff
@@ -229,16 +255,35 @@ bool WaitForDevice(JobControlRecord* jcr, int& retries)
          edit_uint64(jcr->JobId, ed1), jcr->Job);
   }
 
+  if (wait_times.rem_wait_sec > 0) {
+    current_wait = std::min(current_wait, wait_times.rem_wait_sec);
+  } else if (wait_for_device_timeout > 0) {
+    current_wait = wait_for_device_timeout;
+  }
```
sebsura, line 262: Either we customize the wait times via a (constant) global variable, or the jcr, or with an additional argument, but please only use one way to do it.

core/src/stored/wait.cc
```diff
@@ -208,17 +233,18 @@ int WaitForSysop(DeviceControlRecord* dcr)
...
 bool WaitForDevice(JobControlRecord* jcr, int& retries)
 {
   struct timeval tv;
   struct timespec timeout;
   int status = 0;
   bool ok = true;
-  const int max_wait_time = 1 * 60; /* wait 1 minute */
   char ed1[50];
+  auto& wait_times = jcr->sd_impl->device_wait_times;
+  int current_wait = wait_for_device_timeout;
```
sebsura, line 247: Why is this an int ?

core/src/stored/wait.cc
```diff
@@ -229,16 +255,35 @@ bool WaitForDevice(JobControlRecord* jcr, int& retries)
...
   gettimeofday(&tv, NULL);
   timeout.tv_nsec = tv.tv_usec * 1000;
-  timeout.tv_sec = tv.tv_sec + max_wait_time;
+  timeout.tv_sec = tv.tv_sec + current_wait;
 
   Dmsg0(debuglevel, "Going to wait for a device.\n");
 
   /* Wait required time */
   status = pthread_cond_timedwait(&wait_device_release, &device_release_mutex,
                                   &timeout);
```
sebsura, line 272: std::condition_variables can use `std::chrono::X` directly, so we should probably just switch to those.
This condition variable (and the associated mutex) are only used once/twice, so this is not a big change.

core/src/stored/wait.cc
```diff
@@ -41,9 +42,33 @@
...
 static pthread_mutex_t device_release_mutex = PTHREAD_MUTEX_INITIALIZER;
 static pthread_cond_t wait_device_release = PTHREAD_COND_INITIALIZER;
+static int wait_for_device_timeout = kDefaultWaitForDeviceTimeout;
+
+static bool AdvanceWaitBudget(DeviceWaitTimes& wait_times)
+{
+  wait_times.wait_sec *= 2;
+  if (wait_times.wait_sec > wait_times.max_wait) {
+    wait_times.wait_sec = wait_times.max_wait;
+  }
```
sebsura, line 56: Why would we need exponential backoff here ? We are not communicating with another service, we are mostly just waiting for ourselves.

core/src/stored/wait.cc
```diff
@@ -229,16 +255,35 @@ bool WaitForDevice(JobControlRecord* jcr, int& retries)
...
   status = pthread_cond_timedwait(&wait_device_release, &device_release_mutex,
                                   &timeout);
   Dmsg1(debuglevel, "Wokeup from sleep on device status=%d\n", status);
+  if (status == ETIMEDOUT) {
+    if (current_wait == 0 || wait_times.rem_wait_sec <= current_wait) {
+      wait_times.rem_wait_sec = 0;
+      ok = AdvanceWaitBudget(wait_times);
+    } else {
+      wait_times.rem_wait_sec -= current_wait;
+    }
```
sebsura, line 280: Why do we implement part of the logic here and some in another function ? Please move the whole logic into a single function, e.g. something like:

```
auto time_waited = <measure actual time waited>;

bool time_left = AdvanceWaitBudget(wait_times, time_waited);

if (status = ETIMEDOUT && !time_left) {
  ok = false;
}
```

Also, instead of using current_wait, we should actually check how long we waited.  It is totally normal for `pthread_cond_timedwait` to wait longer/shorter than the given timeout.

core/src/stored/askdir.cc
```diff
@@ -221,39 +223,20 @@ bool StorageDaemonDeviceControlRecord::DirFindNextAppendableVolume()
...
-    // but we add some headroom here in case some recommended volume is broken
-    // or in case some devices are currently reserving multiple volumes (maybe
-    // they are currently switching volumes)
-    std::size_t ask_limit = device_count + (device_count >> 3);
-
-    Dmsg0(400, "device count = %llu => ask limit = %llu\n",
-          static_cast<long long unsigned>(device_count),
-          static_cast<long long unsigned>(ask_limit));
+    for (std::size_t vol_index = 1; vol_index <= kMaxAppendVolumeLookups;
+         vol_index++) {
```
sebsura, line 231: This limit was changed by me in 03313de4cfbc22fd6f51f20b67cb81c3be28da7d to be dynamic (instead of a static 20) because a customer had problems. Now it seems like we are back to a static limit.  Why ?

This just makes sure that we will encounter the exact same problem again in a few years.

sebsura: The commit comment does not go into detail, when this actually fails.  If the headroom is too small, then we can make it larger:  Maybe do (2*num_devices) or even (3*num_devices), or w/e.

core/src/stored/vol_mgr.cc
```diff
@@ -796,6 +796,29 @@ bool DeviceControlRecord::Can_i_write_volume()
   return Can_i_use_volume();
 }
 
+bool DeviceControlRecord::IsAppendVolumeBusyOnAnotherDevice()
+{
+  VolumeReservationItem* vol = nullptr;
+  bool busy = false;
+
+  if (find_read_volume(VolumeName)) { return false; }
```
sebsura, line 804: Could you please add comments here explaining what the idea here is ?

I would have guessed that the volume _is busy_, when its currently getting read, so why are we returning false here ? Im probably misunderstanding something, so a comment would be helpful.

sebsura (changes requested):

pstorz:
Pushed 26a7aecd42, which addresses two of @sebsura's review points:

**"instead of using `current_wait`, we should actually check how long we waited"** — `WaitForDevice()` now records the time around `pthread_cond_timedwait()` and charges the measured elapsed time against the budget instead of the nominal slice. A timeout charges at least one second, because the granted slice did elapse and the budget still has to make progress when the measurement rounds down to zero.

**"Please move the whole logic into a single function"** — the budget accounting moved to the new `stored/device_wait_policy.h`, next to `DeviceWaitTimes` itself (relocated there from `stored_jcr_impl.h`). The local `AdvanceWaitBudget()` helper in `wait.cc` is gone, so the rules are no longer split across two places. They are `constexpr`, so `core/src/tests/device_wait_policy_test.cc` covers them with compile-time assertions plus a few gtest cases, without waiting for real time to pass.

The three cases now read as: a real wakeup charges what it used and rescans, a timeout charges its slice and only fails once the budget is exhausted, and a genuine pthread error still fails immediately.

Still open from your review, not touched here:
- the mutable global `wait_for_device_timeout` / `SetWaitForDeviceTimeoutForTesting()` — you suggested an extra defaulted parameter instead
- switching the condition variable and mutex to `std::condition_variable` + `std::chrono`
- the exponential backoff question
- the static limit in `askdir.cc` vs. your dynamic one in 03313de4cfbc22fd6f51f20b67cb81c3be28da7d
- the explanatory comment in `vol_mgr.cc`

Existing `ReservationTest` cases (including `wait_for_device_times_out` and `wait_for_device_uses_total_wait_budget`) still pass unchanged; full unit suite is green.

pstorz:
Thanks for the review — all points are addressed in `d0d2086bc0`, `4e12edb331` and `c1442b2dd8`.

### `wait.cc` — no mutable global, one way to customize, typed timeout, C++ primitives

`wait_for_device_timeout` and the `Set/ResetWaitForDeviceTimeoutForTesting` helpers are gone. `WaitForDevice` now takes `std::chrono::seconds max_wait` (defaulted), so the parameter is the only way to configure it and the tests no longer couple through global state. The `pthread_mutex_t`/`pthread_cond_t` pair became `std::mutex`/`std::condition_variable` and the untyped `int` timeout became a `std::chrono` duration.

### `wait.cc` — why exponential backoff?

You were right to ask: it never had an observable effect. On master `WaitForDevice` uses a flat `max_wait_time = 1*60` and never reads `device_wait_times` at all, and in this PR's version `current_wait` was likewise clamped to a flat 60s — so the doubling only ever accumulated a *total*. It is replaced by a plain remaining-time budget, which is what the code was actually computing.

One judgement call there: master waited **forever**, so there is no prior finite behaviour to preserve. I picked `kMaxTotalDeviceWait = 5 days` as a conservative, generous bound. Happy to change it if you'd prefer a different value or want it configurable.

The per-*device* twins (`dev->min_wait` and friends via `DoubleDevWaitTime`) are still live for `WaitForSysop` and are untouched.

### `askdir.cc` — back to a static limit?

Not quite a revert — I think both pure forms are wrong. `03313de4cfbc` moved to `device_count + device_count/8` because a customer with many devices hit the static cap, which is a real bug. But the dynamic form fails the opposite way: this PR's own test has 1 device and 8 volumes, so the limit collapses to 1 and no alternative volume is ever considered.

The device count is a *lower* bound on how many candidates may have to be skipped, not an upper bound. So the limit is now `max(kMinAppendVolumeLookups, device_count + device_count/8)`: a large device count cannot hit the cap, and neither can a large volume count on few devices. `append_scan_reaches_volumes_beyond_device_count` covers this. The stale "twenty oldest" comment above it is updated too.

### `vol_mgr.cc` — add comments

Done. The rationale only existed in `2dd998754f`'s commit message. The key point, now in the code: the flag does not mean "this volume is busy" generically, it means "busy in a way that retrying with `PreferMountedVols` would resolve". A reader never becomes an append owner, so retrying against one would wait for a condition that can never occur.

### Not addressed, on purpose

There is a lost-wakeup window between a failed rescan and entering the wait — `ReleaseDeviceCond()` notifies without holding the mutex. It is pre-existing and identical with the old `pthread_cond_broadcast`, and it is bounded by the timeout, so I left it rather than grow the scope of this PR. Say the word if you'd rather have it fixed here.

Build is clean and 433/433 unit tests pass.

PR #2683, opened 2026-05-29
tools: bextract and friends should be able to use autochangers

pstorz:

Currently, an autochanger will not work for bextract/bscan/etc. you have to provide a single device. As this is non-obvious esp. with the automatic file autochangers, providing an autochanger should "just work" or ask which device should be used.

#### Please check

##### General

##### Source code quality

##### Tests

### DIFF
--- a/core/src/stored/bcopy.cc
+++ b/core/src/stored/bcopy.cc
@@ -148,7 +148,8 @@ int main(int argc, char* argv[])
                   "Specify the input device name "
                   "(either as name of a Bareos Storage Daemon Device resource "
                   "or identical to the "
-                  "Archive Device in a Bareos Storage Daemon Device resource).")
+                  "Archive Device in a Bareos Storage Daemon Device resource, "
+                  "or a local file volume path).")
       ->type_name(" ");
 
   std::string output_archive;
@@ -157,7 +158,8 @@ int main(int argc, char* argv[])
                   "Specify the output device name "
                   "(either as name of a Bareos Storage Daemon Device resource "
                   "or identical to the "
-                  "Archive Device in a Bareos Storage Daemon Device resource).")
+                  "Archive Device in a Bareos Storage Daemon Device resource, "
+                  "or a local file volume path).")
       ->type_name(" ");
 
   ParseBareosApp(bcopy_app, argc, argv);
@@ -168,9 +170,6 @@ int main(int argc, char* argv[])
 
   working_directory = work_dir.c_str();
 
-  my_config = InitSdConfig(configFile.c_str(), M_CONFIG_ERROR);
-  ParseSdConfig(configFile.c_str(), M_CONFIG_ERROR);
-
   if (input_archive.empty()) {
     printf(T_("Missing input device. %sNothing done.\n"),
            AvailableDevicesListing().c_str());
@@ -179,6 +178,20 @@ int main(int argc, char* argv[])
     printf(T_("Missing output device. %sNothing done.\n"),
            AvailableDevicesListing().c_str());
     return BEXIT_CLI_PARSING_ERROR;
+  }
+
+  const bool config_explicit = !configFile.empty();
+  const bool use_sd_config
+      = config_explicit || !IsDirectLocalVolumePath(input_archive.c_str())
+        || !IsDirectLocalVolumePath(output_archive.c_str());
+
+  my_config = nullptr;
+  if (use_sd_config) {
+    my_config = InitSdConfig(configFile.c_str(), M_CONFIG_ERROR);
+    ParseSdConfig(configFile.c_str(), M_CONFIG_ERROR);
+  } else if (!DirectorName.empty()) {
+    Emsg0(M_ERROR_TERM, 0,
+          T_("--director requires a Storage Daemon configuration.\n"));
   }
 
 
@@ -195,10 +208,12 @@ int main(int argc, char* argv[])
     }
   }
 
-  LoadSdPlugins(me->plugin_directory, me->plugin_names);
+  if (my_config) {
+    LoadSdPlugins(me->plugin_directory, me->plugin_names);
 
-  ReadCryptoCache(me->working_directory, "bareos-sd",
-                  GetFirstPortHostOrder(me->SDaddrs));
+    ReadCryptoCache(me->working_directory, "bareos-sd",
+                    GetFirstPortHostOrder(me->SDaddrs));
+  }
 
   // Setup and acquire input device for reading
   Dmsg0(100, "About to setup input jcr\n");
--- a/core/src/stored/bcopy.cc
+++ b/core/src/stored/bcopy.cc
@@ -170,12 +170,12 @@ int main(int argc, char* argv[])
   working_directory = work_dir.c_str();
 
   if (input_archive.empty()) {
-    LoadSdConfigForDeviceListingIfAvailable(configFile.c_str());
+    LoadSdConfigIfAvailable(configFile.c_str());
     printf(T_("Missing input device. %sNothing done.\n"),
            AvailableDevicesListing().c_str());
     return BEXIT_CLI_PARSING_ERROR;
   } else if (output_archive.empty()) {
-    LoadSdConfigForDeviceListingIfAvailable(configFile.c_str());
+    LoadSdConfigIfAvailable(configFile.c_str());
     printf(T_("Missing output device. %sNothing done.\n"),
            AvailableDevicesListing().c_str());
     return BEXIT_CLI_PARSING_ERROR;
--- a/core/src/stored/bcopy.cc
+++ b/core/src/stored/bcopy.cc
@@ -146,20 +146,18 @@ int main(int argc, char* argv[])
   bcopy_app
       .add_option("input-archive", input_archive,
                   "Specify the input device name "
-                  "(either as name of a Bareos Storage Daemon Device resource "
-                  "or identical to the "
-                  "Archive Device in a Bareos Storage Daemon Device resource, "
-                  "or a local file volume path).")
+                  "(either the name of a Bareos Storage Daemon Device "
+                  "resource, the Archive Device in such a resource, or a "
+                  "local file volume path).")
       ->type_name(" ");
 
   std::string output_archive;
   bcopy_app
       .add_option("output-archive", output_archive,
                   "Specify the output device name "
-                  "(either as name of a Bareos Storage Daemon Device resource "
-                  "or identical to the "
-                  "Archive Device in a Bareos Storage Daemon Device resource, "
-                  "or a local file volume path).")
+                  "(either the name of a Bareos Storage Daemon Device "
+                  "resource, the Archive Device in such a resource, or a "
+                  "local file volume path).")
       ->type_name(" ");
 
   ParseBareosApp(bcopy_app, argc, argv);
--- a/core/src/stored/bcopy.cc
+++ b/core/src/stored/bcopy.cc
@@ -171,10 +171,12 @@ int main(int argc, char* argv[])
   working_directory = work_dir.c_str();
 
   if (input_archive.empty()) {
+    LoadSdConfigForDeviceListingIfAvailable(configFile.c_str());
     printf(T_("Missing input device. %sNothing done.\n"),
            AvailableDevicesListing().c_str());
     return BEXIT_CLI_PARSING_ERROR;
   } else if (output_archive.empty()) {
+    LoadSdConfigForDeviceListingIfAvailable(configFile.c_str());
     printf(T_("Missing output device. %sNothing done.\n"),
            AvailableDevicesListing().c_str());
     return BEXIT_CLI_PARSING_ERROR;
--- a/core/src/stored/bcopy.cc
+++ b/core/src/stored/bcopy.cc
@@ -183,8 +183,9 @@ int main(int argc, char* argv[])
 
   const bool config_explicit = !configFile.empty();
   const bool use_sd_config
-      = config_explicit || !IsDirectLocalVolumePath(input_archive.c_str())
-        || !IsDirectLocalVolumePath(output_archive.c_str());
+      = config_explicit
+        || !IsLocalFilesystemVolumePath(input_archive.c_str())
+        || !IsLocalFilesystemVolumePath(output_archive.c_str());
 
   my_config = nullptr;
   if (use_sd_config) {
--- a/core/src/stored/bcopy.cc
+++ b/core/src/stored/bcopy.cc
@@ -97,7 +97,7 @@ int main(int argc, char* argv[])
       ->type_name("<bootstrap>");
 
   std::string configFile;
-  bcopy_app
+  auto config_option = bcopy_app
       .add_option("-c,--config", configFile,
                   "Use <path> as configuration file or directory.")
       ->check(CLI::ExistingPath)
@@ -108,6 +108,7 @@ int main(int argc, char* argv[])
       .add_option("-D,--director", DirectorName,
                   "Specify a director name specified in the storage. "
                   "Configuration file for the Key Encryption Key selection.")
+      ->needs(config_option)
       ->type_name("<director>");
 
   AddDebugOptions(bcopy_app);
@@ -189,9 +190,6 @@ int main(int argc, char* argv[])
   if (use_sd_config) {
     my_config = InitSdConfig(configFile.c_str(), M_CONFIG_ERROR);
     ParseSdConfig(configFile.c_str(), M_CONFIG_ERROR);
-  } else if (!DirectorName.empty()) {
-    Emsg0(M_ERROR_TERM, 0,
-          T_("--director requires a Storage Daemon configuration.\n"));
   }
 
 
--- a/core/src/stored/bls.cc
+++ b/core/src/stored/bls.cc
@@ -217,7 +217,7 @@ int main(int argc, char* argv[])
   ParseBareosApp(bls_app, argc, argv);
 
   if (device_names.size() == 0) {
-    LoadSdConfigForDeviceListingIfAvailable(configfile);
+    LoadSdConfigIfAvailable(configfile);
     printf(T_("Missing input device. %sNothing done.\n"),
            AvailableDevicesListing().c_str());
     return BEXIT_CLI_PARSING_ERROR;
--- a/core/src/stored/bls.cc
+++ b/core/src/stored/bls.cc
@@ -227,7 +227,7 @@ int main(int argc, char* argv[])
   const bool all_direct_local
       = std::all_of(device_names.begin(), device_names.end(),
                     [](const std::string& device_name) {
-                      return IsDirectLocalVolumePath(device_name.c_str());
+                      return IsLocalFilesystemVolumePath(device_name.c_str());
                     });
   const bool use_sd_config = config_explicit || !all_direct_local;
 
--- a/core/src/stored/bls.cc
+++ b/core/src/stored/bls.cc
@@ -29,6 +29,7 @@
 #if !defined(HAVE_MSVC)
 #  include <unistd.h>
 #endif
+#include <algorithm>
 #include "include/bareos.h"
 #include "include/exit_codes.h"
 #include "include/streams.h"
@@ -209,18 +210,33 @@ int main(int argc, char* argv[])
                   "Specify the input device name "
                   "(either as name of a Bareos Storage Daemon Device resource "
                   "or identical to the "
-                  "Archive Device in a Bareos Storage Daemon Device resource).")
+                  "Archive Device in a Bareos Storage Daemon Device resource, "
+                  "or a local file volume path).")
       ->type_name(" ");
 
   ParseBareosApp(bls_app, argc, argv);
-
-  my_config = InitSdConfig(configfile, M_CONFIG_ERROR);
-  ParseSdConfig(configfile, M_CONFIG_ERROR);
 
   if (device_names.size() == 0) {
     printf(T_("Missing input device. %sNothing done.\n"),
            AvailableDevicesListing().c_str());
     return BEXIT_CLI_PARSING_ERROR;
+  }
+
+  const bool config_explicit = configfile != nullptr;
+  const bool all_direct_local
+      = std::all_of(device_names.begin(), device_names.end(),
+                    [](const std::string& device_name) {
+                      return IsDirectLocalVolumePath(device_name.c_str());
+                    });
+  const bool use_sd_config = config_explicit || !all_direct_local;
+
+  my_config = nullptr;
+  if (use_sd_config) {
+    my_config = InitSdConfig(configfile, M_CONFIG_ERROR);
+    ParseSdConfig(configfile, M_CONFIG_ERROR);
+  } else if (!DirectorName.empty()) {
+    Emsg0(M_ERROR_TERM, 0,
+          T_("--director requires a Storage Daemon configuration.\n"));
   }
 
   if (!DirectorName.empty()) {
@@ -235,10 +251,12 @@ int main(int argc, char* argv[])
     }
   }
 
-  LoadSdPlugins(me->plugin_directory, me->plugin_names);
+  if (my_config) {
+    LoadSdPlugins(me->plugin_directory, me->plugin_names);
 
-  ReadCryptoCache(me->working_directory, "bareos-sd",
-                  GetFirstPortHostOrder(me->SDaddrs));
+    ReadCryptoCache(me->working_directory, "bareos-sd",
+                    GetFirstPortHostOrder(me->SDaddrs));
+  }
 
   if (ff->included_files_list == nullptr) { AddFnameToIncludeList(ff, 0, "/"); }
 
--- a/core/src/stored/bls.cc
+++ b/core/src/stored/bls.cc
@@ -217,6 +217,7 @@ int main(int argc, char* argv[])
   ParseBareosApp(bls_app, argc, argv);
 
   if (device_names.size() == 0) {
+    LoadSdConfigForDeviceListingIfAvailable(configfile);
     printf(T_("Missing input device. %sNothing done.\n"),
            AvailableDevicesListing().c_str());
     return BEXIT_CLI_PARSING_ERROR;
--- a/core/src/stored/bscan.cc
+++ b/core/src/stored/bscan.cc
@@ -242,7 +242,7 @@ int main(int argc, char* argv[])
   ParseBareosApp(bscan_app, argc, argv);
 
   if (device_name.empty()) {
-    LoadSdConfigForDeviceListingIfAvailable(configfile);
+    LoadSdConfigIfAvailable(configfile);
     printf(T_("Missing input device. %sNothing done.\n"),
            AvailableDevicesListing().c_str());
     return BEXIT_CLI_PARSING_ERROR;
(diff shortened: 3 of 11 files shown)


core/src/stored/bcopy.cc
```diff
@@ -148,7 +148,8 @@ int main(int argc, char* argv[])
                   "Specify the input device name "
                   "(either as name of a Bareos Storage Daemon Device resource "
                   "or identical to the "
-                  "Archive Device in a Bareos Storage Daemon Device resource).")
+                  "Archive Device in a Bareos Storage Daemon Device resource, "
+                  "or a local file volume path).")
```
sebsura: There are two `or`s here, is that on purpose ? 

Maybe this should be `Specify the input device name (either ... or ...) or a path to a local file volume`.  Currently it says that a local file volume path is a input device name which does not sound correct.

core/src/stored/bcopy.cc
```diff
@@ -168,19 +170,32 @@ int main(int argc, char* argv[])
...
+        || !IsDirectLocalVolumePath(output_archive.c_str());
+
+  my_config = nullptr;
+  if (use_sd_config) {
+    my_config = InitSdConfig(configFile.c_str(), M_CONFIG_ERROR);
+    ParseSdConfig(configFile.c_str(), M_CONFIG_ERROR);
+  } else if (!DirectorName.empty()) {
+    Emsg0(M_ERROR_TERM, 0,
+          T_("--director requires a Storage Daemon configuration.\n"));
+  }
```
sebsura: Im pretty sure that there is a builtin way in CLI11 to do this: `->needs()`, see https://github.com/CLIUtils/CLI11:

> ->needs(opt): This option requires another option to also be present, opt is an Option pointer. Options can be removed from the needs with remove_needs(opt). The option can also be specified with a string containing the name of the option

docs/manuals/source/include/autogenerated/usage/bcopy.txt
```diff
@@ -4,12 +4,12 @@ Positionals:
     input-archive  
         Specify the input device name (either as name of a Bareos Storage 
         Daemon Device resource or identical to the Archive Device in a Bareos 
-        Storage Daemon Device resource). 
+        Storage Daemon Device resource, or a local file volume path).
```
sebsura: Same here (and in the other docs).

core/src/stored/bscan.cc
```diff
@@ -262,15 +274,19 @@ int main(int argc, char* argv[])
...
+    ReadCryptoCache(me->working_directory, "bareos-sd",
+                    GetFirstPortHostOrder(me->SDaddrs));
+  }
 
   /* Check if -w option given, otherwise use resource for working directory
    */
   if (!work_dir.empty()) {
     working_directory = work_dir.c_str();
+  } else if (!my_config) {
+    working_directory = "/tmp";
```
sebsura, line 300: I didnt test it yet, but im pretty sure this does not work on windows. There it should be `%TMP%` or something like this.

core/src/stored/butil.cc
```diff
@@ -39,25 +39,331 @@
...
+
+static bool TryStatPath(const std::string& path, struct stat& statp)
+{
+  return !path.empty() && stat(path.c_str(), &statp) == 0;
+}
+
+static bool IsDirectLocalVolumeStat(const struct stat& statp)
+{
+  return S_ISREG(statp.st_mode) || S_ISDIR(statp.st_mode);
+}
```
sebsura: Please do not create these single line, single use functions.  It makes it much harder to read than just having that text in the actual caller.

core/src/stored/butil.cc
```diff
@@ -39,25 +39,331 @@
...
+    normalized.pop_back();
+  }
+
+  return normalized;
+}
+
+static std::string AccessDescription(bool readonly)
+{
+  return readonly ? "reading" : "writing";
+}
```
sebsura: Do we need to allocate here ? Would `std::string_view` / `char*` not suffice ?

core/src/stored/butil.cc
```diff
@@ -39,25 +39,331 @@
...
+{
+  std::string normalized{resource_name};
+  if (normalized.size() >= 2 && normalized.front() == '"'
+      && normalized.back() == '"') {
+    normalized.erase(0, 1);
+    normalized.pop_back();
+  }
+
+  return normalized;
+}
```
sebsura: This should take in a `std::string_view` and it should first determine whether to remove quotes, and only then create the string, so we do not have to copy it multiple times (which is caused by `erase`).

core/src/stored/butil.cc
```diff
@@ -39,25 +39,331 @@
...
+  return S_ISREG(statp.st_mode) || S_ISDIR(statp.st_mode);
+}
+
+static std::string DirectoryName(const std::string& path)
+{
+  auto separator = path.find_last_of("/\\");
+  if (separator == std::string::npos) { return "."; }
+  if (separator == 0) { return path.substr(0, 1); }
+  return path.substr(0, separator);
+}
```
sebsura, line 127: Please add some comments here. Even something as simple as `directory name` can have multiple different meanings.

1. Does this include the trailing slash ?
2.  Does this _always_ include a trailing slash ?
3. Can I _always_ reconstruct the original path from both directoryname + basename the same way ? I.e. is it guaranteed that `DirectoryName() + BaseName() == path` ?

These kinds of functions always leads to weird bugs later down the line because the author and user have slight different understandings what DirectoryName/BaseName are supposed to do in some edge cases.

core/src/stored/butil.cc
```diff
@@ -39,25 +39,331 @@
...
+  if (len > 0 && len < sizeof(path)) {
+    return DirectoryName(std::string(path, len));
+  }
+#else
+  char path[PATH_MAX];
+  auto len = readlink("/proc/self/exe", path, sizeof(path) - 1);
+  if (len > 0) {
+    path[len] = '\0';
+    return DirectoryName(path);
+  }
```
sebsura, line 159: I am under the impression that mac / freebsd do not have `/proc`

core/src/stored/butil.h
```diff
@@ -43,6 +43,8 @@ JobControlRecord* SetupJcr(const char* name,
                            DeviceControlRecord* dcr,
                            const std::string& VolumeName,
                            bool readonly);
+bool IsDirectLocalVolumePath(const char* device_name);
```
sebsura: What is meant by `direct` and `local` ? To me that sounds like we do not support volumes on i.e. nfs shares.

core/src/stored/butil.cc
```diff
@@ -127,6 +433,12 @@ JobControlRecord* SetupJcr(const char* name,
 
 std::string AvailableDevicesListing()
 {
+  if (!my_config) {
+    return T_(
+        "Direct local file volume paths may be used without an SD "
+        "configuration.\n");
+  }
```
sebsura: Either the function needs to be renamed or this needs to be moved somewhere else, as this does not really make sense as a return value for "AvailableDevicesListing()" IMO.

sebsura: Maybe we should have a function `ListDevicesInConfig` which takes a (nonnull) config and returns what this function does, and a wrapping function that contains this if.

core/src/stored/butil.cc
```diff
@@ -139,9 +451,45 @@ std::string AvailableDevicesListing()
...
+    autochanger_str += " \"";
+    autochanger_str += autochanger->resource_name_;
+    autochanger_str += "\"";
+
+    std::vector<std::string> members;
+    DeviceResource* device = nullptr;
+    int i;
+    foreach_alist_index (i, device, autochanger->device_resources) {
+      members.emplace_back(FormatDeviceChoice(device));
+    }
```
sebsura: Since `i` is not used here, could we not use a c++ for each loop ? I thought we implemented that for alist.

core/src/stored/butil.cc
```diff
@@ -139,9 +451,45 @@ std::string AvailableDevicesListing()
...
+    std::sort(members.begin(), members.end());
+
+    if (!members.empty()) {
+      autochanger_str += " ->";
+      for (const auto& member : members) {
+        autochanger_str += "\n    ";
+        autochanger_str += member;
+      }
+    }
```
sebsura, line 487: Do I select an autochanger or a device inside an autochanger ?
Do devices in an autochanger also end up on the list of devices above ?

core/src/stored/butil.cc
```diff
@@ -39,25 +39,331 @@
...
+void LoadSdConfigForDeviceListingIfAvailable(const char* config_path)
+{
+  if (my_config || !config_path || config_path[0] == '\0') { return; }
+
+  struct stat statp;
+  if (stat(config_path, &statp) != 0) { return; }
+
+  my_config = InitSdConfig(config_path, M_CONFIG_ERROR);
+  ParseSdConfig(config_path, M_CONFIG_ERROR);
+}
```
sebsura: This function does not have anything do to  with device listings as far as I can tell.
Function( name)s should make sense on their own, even without looking at their (probably only) caller.

core/src/stored/butil.cc
```diff
@@ -39,25 +39,331 @@
...
 static void MyFreeJcr(JobControlRecord* jcr);
 
+bool IsDirectLocalVolumePath(const char* device_name)
+{
+  if (!device_name) { return false; }
+
+  auto normalized_path = StripQuotes(device_name);
+  struct stat statp;
+  return TryStatPath(normalized_path, statp) && IsDirectLocalVolumeStat(statp);
+}
```
sebsura, line 370: Please add a description of what this is actually trying to accomplish.  What is a direct/local path, etc.

core/src/stored/butil.cc
sebsura: Also I do not see where we check for anything `direct` or `local` here.

core/src/stored/butil.cc
```diff
@@ -282,54 +633,79 @@ static void MyFreeJcr(JobControlRecord* jcr)
...
   }
 
   if (!found) {
-    Pmsg2(0, T_("Could not find device \"%s\" in config file %s.\n"),
-          archive_device_string, configfile);
-    return NULL;
+    result = ResolveDirectLocalVolumePath(normalized_resource_name);
+    device_resource = result.device_resource;
+    found = (device_resource != nullptr);
+  }
```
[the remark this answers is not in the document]
sebsura: Was this addressed already ?

sebsura (changes requested):
I had a quick look.

core/src/stored/bcopy.cc
```diff
@@ -146,18 +147,18 @@ int main(int argc, char* argv[])
   bcopy_app
       .add_option("input-archive", input_archive,
                   "Specify the input device name "
-                  "(either as name of a Bareos Storage Daemon Device resource "
-                  "or identical to the "
-                  "Archive Device in a Bareos Storage Daemon Device resource).")
+                  "(either the name of a Bareos Storage Daemon Device "
+                  "resource, the Archive Device in such a resource, or a "
+                  "local file volume path).")
       ->type_name(" ");
```
sebsura, line 153: This still does not make sense to me.

```
Specify the input device name ( ..., or a local file volume path)
```
A "local file volume path" is _not_ an input device.  What goes against just putting the local file volume path outside of the parens ?

core/src/stored/bscan.cc
```diff
@@ -262,15 +274,31 @@ int main(int argc, char* argv[])
...
+    auto temp_directory_length
+        = GetTempPathA(sizeof(temp_directory), temp_directory);
+    if (temp_directory_length > 0
+        && temp_directory_length < sizeof(temp_directory)) {
+      working_directory = temp_directory;
+    } else {
+      working_directory = "C:\\Windows\\Temp";
+    }
+#else
+    working_directory = "/tmp";
```
sebsura, line 300: This should probably just use `std::filesystem::temporary_directory::path()`

sebsura: You literally created a function for this `CreateTempDirectory()`

core/src/stored/butil.cc
sebsura: The comment should still mention whether the return value _always_ is supposed to contain a trailing slash or not.

core/src/stored/butil.cc
```diff
@@ -39,25 +39,347 @@
...
+#endif
+  return path.substr(0, separator);
+}
+
+#if defined(HAVE_WIN32)
+std::string DirectoryNameForTest(const std::string& path)
+{
+  return DirectoryName(path);
+}
+#endif
```
sebsura, line 134: What is this supposed to be ???

core/src/stored/butil.cc
```diff
@@ -39,25 +39,347 @@
...
+static std::string BaseName(const std::string& path)
+{
+  // Return the final path component. If the input ends in a separator or is
+  // the root path, the basename is empty.
+  auto separator = path.find_last_of("/\\");
+  if (separator == std::string::npos) { return path; }
+  return path.substr(separator + 1);
+}
+
+static std::string GetExecutableDirectory()
```
sebsura, line 145: Why do need this special logic here ? Would it not be better if all bareos things do the things in the same way ?
As far as I can tell `exepath` normally gets set by `MyNameIs`, which is also called by `bscan`, so why do we not just simply use `exepath` everywhere ?

core/src/stored/butil.cc
```diff
@@ -39,25 +39,347 @@
...
-                                       bool readonly);
+static DeviceResolutionResult find_device_res(char* archive_device_string,
+                                              bool readonly);
 static void MyFreeJcr(JobControlRecord* jcr);
 
+bool IsLocalFilesystemVolumePath(const char* device_name)
+{
+  // Direct local paths are regular files or directories that can be used
+  // without an SD device resource.
+  if (!device_name) { return false; }
```
sebsura, line 364: I still do not see what is direct, or local about this.

core/src/stored/butil.cc
```diff
@@ -39,25 +39,347 @@
...
                                    char* dev_name,
                                    const std::string& VolumeName,
                                    bool readonly);
-static DeviceResource* find_device_res(char* archive_device_string,
-                                       bool readonly);
+static DeviceResolutionResult find_device_res(char* archive_device_string,
+                                              bool readonly);
 static void MyFreeJcr(JobControlRecord* jcr);
 
+bool IsLocalFilesystemVolumePath(const char* device_name)
```
sebsura, line 360: https://en.wikipedia.org/wiki/File_system

```
A local file system is a capability of an [operating system](https://en.wikipedia.org/wiki/Operating_system) that services the applications running on the same [computer](https://en.wikipedia.org/wiki/Computer).[[1]](https://en.wikipedia.org/wiki/File_system#cite_note-1)[[2]](https://en.wikipedia.org/wiki/File_system#cite_note-2) A [distributed file system](https://en.wikipedia.org/wiki/Distributed_file_system) is a [protocol](https://en.wikipedia.org/wiki/Communication_protocol) that provides file access between [networked](https://en.wikipedia.org/wiki/Computer_network) computers.
```

I do not see any check to see if the filesystem is not a distributed one.

core/src/stored/butil.cc
```diff
@@ -39,25 +39,347 @@
...
+{
+  // Direct local paths are regular files or directories that can be used
+  // without an SD device resource.
+  if (!device_name) { return false; }
+
+  auto normalized_path = StripQuotes(device_name);
```
sebsura, line 366: This does not belong here.  This is a parsers job, and should be done during startup.

core/src/stored/butil.cc
sebsura: A good name might be `IsRegularFileOrDirectory`

core/src/stored/butil.h
```diff
@@ -43,9 +43,16 @@ JobControlRecord* SetupJcr(const char* name,
...
                            bool readonly);
+// Return true for a directly stat-able local file or directory path.
+bool IsLocalFilesystemVolumePath(const char* device_name);
+void LoadSdConfigIfAvailable(const char* config_path);
 void DisplayTapeErrorStatus(JobControlRecord* jcr, Device* dev);
 std::string AvailableDevicesListing();
 
+#if defined(HAVE_WIN32)
+std::string DirectoryNameForTest(const std::string& path);
+#endif
```
sebsura, line 54: Why does this exist ?

core/src/stored/butil.cc
```diff
@@ -282,54 +667,79 @@ static void MyFreeJcr(JobControlRecord* jcr)
...
  */
-static DeviceResource* find_device_res(char* archive_device_string,
-                                       bool readonly)
+static DeviceResolutionResult find_device_res(char* archive_device_string,
+                                              bool readonly)
 {
+  DeviceResolutionResult result;
   bool found = false;
   DeviceResource* device_resource;
+  auto normalized_resource_name = StripQuotes(archive_device_string);
```
sebsura, line 676: This is not where this belongs.  We should put parsing on the outer boundary.  Functions in the heart of the application, should not deal with that.  They should expect that this is done correctly by their caller.

core/src/stored/bls.cc
```diff
@@ -209,20 +210,36 @@ int main(int argc, char* argv[])
...
 
   ParseBareosApp(bls_app, argc, argv);
 
-  my_config = InitSdConfig(configfile, M_CONFIG_ERROR);
-  ParseSdConfig(configfile, M_CONFIG_ERROR);
-
   if (device_names.size() == 0) {
+    LoadSdConfigIfAvailable(configfile);
     printf(T_("Missing input device. %sNothing done.\n"),
            AvailableDevicesListing().c_str());
```
sebsura, line 222: If `configfile == nullptr` then it does not make sense to try to list available devices, as currently loadsdconfigifavailable does not do anything in that case.

core/src/stored/bls.cc
```diff
@@ -209,20 +210,36 @@ int main(int argc, char* argv[])
...
+  const bool use_sd_config = config_explicit || !all_direct_local;
+
+  my_config = nullptr;
+  if (use_sd_config) {
+    my_config = InitSdConfig(configfile, M_CONFIG_ERROR);
+    ParseSdConfig(configfile, M_CONFIG_ERROR);
+  } else if (!DirectorName.empty()) {
+    Emsg0(M_ERROR_TERM, 0,
+          T_("--director requires a Storage Daemon configuration.\n"));
+  }
```
sebsura, line 241: This should be checked by cli11 (and afaik it does do that), so this should be at most an assertion, but it can also be removed.

core/src/stored/bls.cc
```diff
@@ -235,10 +252,12 @@ int main(int argc, char* argv[])
...
 
-  LoadSdPlugins(me->plugin_directory, me->plugin_names);
+  if (my_config) {
+    LoadSdPlugins(me->plugin_directory, me->plugin_names);
 
-  ReadCryptoCache(me->working_directory, "bareos-sd",
-                  GetFirstPortHostOrder(me->SDaddrs));
+    ReadCryptoCache(me->working_directory, "bareos-sd",
+                    GetFirstPortHostOrder(me->SDaddrs));
+  }
```
sebsura, line 260: This means that e.g. autoxflate will not work if you use this without the config. For a lot of things this is fine, but we should probably mention it somewhere in the docs (i.e. for bextract, bscan).

core/src/stored/butil.cc
```diff
@@ -39,25 +39,347 @@
...
+      && (S_ISREG(statp.st_mode) || S_ISDIR(statp.st_mode));
+}
+
+void LoadSdConfigIfAvailable(const char* config_path)
+{
+  if (my_config || !config_path || config_path[0] == '\0') { return; }
+
+  struct stat statp;
+  if (stat(config_path, &statp) != 0) { return; }
+
```
sebsura, line 378: The logic is not the same as the one that the actual loading logic uses.  It has fallbacks for default directories/default paths, that are not checked here.

core/src/stored/bls.cc
```diff
@@ -209,20 +210,36 @@ int main(int argc, char* argv[])
...
     return BEXIT_CLI_PARSING_ERROR;
   }
 
+  const bool config_explicit = configfile != nullptr;
+  const bool all_direct_local
+      = std::all_of(device_names.begin(), device_names.end(),
+                    [](const std::string& device_name) {
+                      return IsLocalFilesystemVolumePath(device_name.c_str());
+                    });
+  const bool use_sd_config = config_explicit || !all_direct_local;
```
sebsura, line 232: We should always at least try to load the config, i.e. to load plugins.

core/src/stored/butil.cc
```diff
@@ -39,25 +39,347 @@
...
+extern char* exepath;
 
 namespace storagedaemon {
 
+namespace {
+
+struct DeviceResolutionResult {
+  DeviceResource* device_resource{nullptr};
+  std::string detail;
```
sebsura, line 70: What is this supposed to be? Choose a better name and also add a comment describing what this is supposed to be used for.

core/src/stored/butil.cc
```diff
@@ -39,25 +39,347 @@
...
+#include <string_view>
+#include <sys/stat.h>
+#include <vector>
+#if defined(HAVE_WIN32)
+#  include <windows.h>
+#else
+#  include <unistd.h>
+#endif
+
+extern char* exepath;
```
sebsura, line 62: Do not "import" external things without using a header for it if possible.

core/src/stored/butil.cc
```diff
@@ -39,25 +39,347 @@
...
 
+namespace {
+
+struct DeviceResolutionResult {
+  DeviceResource* device_resource{nullptr};
+  std::string detail;
+  std::string inferred_volume_name;
+};
+
+static std::vector<std::unique_ptr<DeviceResource>> synthetic_device_resources;
```
sebsura, line 74: Try to not introduce new global variables.

core/src/stored/butil.cc
sebsura: Just export DirectoryName if its necessary.

Why is the same not done for BaseName, etc.  Are these functions untested ?

sebsura: This also looks like a function that does not belong here, but (probably) in lib/util.h.

core/src/stored/butil.cc
```diff
@@ -39,25 +39,347 @@
...
+  if (!executable_directory.empty()) {
+    std::string build_tree_backend_dir{std::move(executable_directory)};
+    if (!IsPathSeparator(build_tree_backend_dir.back())) {
+      build_tree_backend_dir += PathSeparator;
+    }
+    build_tree_backend_dir += "backends";
+    backend_directories.emplace_back(std::move(build_tree_backend_dir));
+  }
+
+  return LoadStorageBackend(device_type, backend_directories);
```
sebsura, line 184: Are we sure that this is never called when a config is loaded ? Because this currently seems to ignore the configured backend path completely if a config _is_ loaded.

core/src/stored/butil.cc
```diff
@@ -39,25 +39,347 @@
...
+#else
+  return true;
+#endif
+}
+
+static DeviceResolutionResult ResolveDirectLocalVolumePath(std::string path)
```
sebsura, line 190: Im not sure what we are really resolving here, but this does not make it clear at all that all this does is trying to setup a simple file device + a volume for it.

core/src/stored/butil.cc
```diff
@@ -39,25 +39,347 @@
...
+    prompt += "  ";
+    prompt += std::to_string(i + 1);
+    prompt += ") ";
+    prompt += FormatDeviceChoice(devices[i]);
+    if (devices[i]->autoselect) { prompt += " [autoselect]"; }
+    prompt += '\n';
+  }
+
+  prompt += T_("Enter selection number: ");
+  return prompt;
```
sebsura, line 248: Please use a `std::stringstream` here.

core/src/stored/butil.cc
```diff
@@ -39,25 +39,347 @@
...
+  synthetic_device_resources.emplace_back(std::move(device_resource));
+
+  return result;
+}
+
+static std::string FormatDeviceChoice(const DeviceResource* device_resource)
+{
+  return "\"" + std::string(device_resource->resource_name_) + "\" ("
+         + device_resource->archive_device_string + ")";
+}
```
sebsura, line 228: This should also directly take in a `stringstream`

core/src/stored/butil.cc
```diff
@@ -39,25 +39,347 @@
...
+  detail += T_(". Please specify one of these Device resources explicitly:\n");
+
+  for (const auto* device_resource : devices) {
+    detail += "  ";
+    detail += FormatDeviceChoice(device_resource);
+    if (device_resource->autoselect) { detail += " [autoselect]"; }
+    detail += '\n';
+  }
+
+  return detail;
```
sebsura, line 269: same here: stringstream.

core/src/stored/butil.cc
```diff
@@ -39,25 +39,347 @@
...
+  }
+
+  if (eligible_devices.empty()) {
+    result.detail = T_("Autochanger \"");
+    result.detail += autochanger_resource->resource_name_;
+    result.detail += T_("\" has no devices suitable for ");
+    result.detail += AccessDescription(readonly);
+    result.detail += ".\n";
+    return result;
+  }
```
sebsura, line 336: stringstream.

core/src/stored/butil.cc
```diff
@@ -39,25 +39,347 @@
...
+
+  result.device_resource
+      = PromptForAutochangerDevice(autochanger_resource, eligible_devices);
```
sebsura, line 339: This should probably _not_ be here, deeply buried in other logic.  The decision to prompt or not should lie in the actual "application code".  This will just lead to problems (i.e. when you want to implement some `--non-interactive` flag).

core/src/stored/butil.cc
```diff
@@ -108,7 +430,7 @@ JobControlRecord* SetupJcr(const char* name,
 {
   JobControlRecord* jcr = SetupDummyJcr(name, bsr, director);
 
-  InitAutochangers();
+  if (my_config) { InitAutochangers(); }
```
sebsura, line 433: Are we sure that this is not something that is always required ?

core/src/stored/butil.cc
```diff
@@ -139,11 +462,58 @@ std::string AvailableDevicesListing()
...
+      devices_str += autochanger;
+    }
+  }
   return devices_str;
 }
+
+std::string AvailableDevicesListing()
+{
+  if (!my_config) {
+    return T_(
```
sebsura, line 510: Do not use `T_()` for new strings.  Nothing else here is translated, and we are trying to phase it out anyways.

core/src/stored/butil.cc
```diff
@@ -282,54 +667,79 @@ static void MyFreeJcr(JobControlRecord* jcr)
...
+      foreach_res (autochanger_resource, R_AUTOCHANGER) {
+        Dmsg2(900, "Compare %s and %s\n", autochanger_resource->resource_name_,
+              normalized_resource_name.c_str());
+        if (bstrcmp(autochanger_resource->resource_name_,
+                    normalized_resource_name.c_str())) {
+          result = ResolveAutochangerDevice(autochanger_resource, readonly);
+          device_resource = result.device_resource;
+          found = (device_resource != nullptr);
+          break;
+        }
```
sebsura, line 714: There should be a dmsg here that outputs something when the comparision is true, so that its clearer when reading the debug log what actually happened.

sebsura (changes requested):

PR #2701, opened 2026-06-19
filed: Enable sleep suppression for client-initiated jobs

pstorz:
For client-initiated director sessions in filed, start sleep suppression when processing backup or restore commands, while keeping verify unchanged.

Windows support existed before, but was not being called by client-initiated connects.
This PR now also adds Linux and MacOS so that also sleeping is supporessed there while jobs are running.

#### Please check

##### General

##### Source code quality

##### Tests


sduehr:
On Linux inhibiting suspend only when a jobs runs works now.
But on Windows it inhibits suspend already when bareos-fd is running.

sduehr:
With the last change it now works as expected also on Windows.

core/src/filed/CMakeLists.txt
```diff
@@ -20,6 +20,12 @@ message("Entering ${CMAKE_CURRENT_SOURCE_DIR}")
 
 include(BareosConfigureFile)
 include(BareosWindowsResource)
+include(CheckIncludeFiles)
+
+option(fd-client-sleep-inhibition
```
sebsura: every other option is `UPPER_SNAKE_CASE`, why was kebab case chosen here ?

core/src/filed/CMakeLists.txt
```diff
@@ -29,6 +35,29 @@ if(HAVE_WIN32)
...
+  if(NOT HAVE_SYSTEMD)
+    message(
+      FATAL_ERROR
+        "fd-client-sleep-inhibition requires systemd support on Linux. "
+        "Set -Dfd-client-sleep-inhibition=OFF to disable manually."
+    )
+  endif()
+  set(_previous_required_includes "${CMAKE_REQUIRED_INCLUDES}")
+  set(CMAKE_REQUIRED_INCLUDES ${SYSTEMD_INCLUDE_DIRS})
+  check_include_files("systemd/sd-bus.h" HAVE_SYSTEMD_SD_BUS_H)
```
sebsura: This probably belongs in `BareosCheckIncludes.cmake`

core/src/filed/CMakeLists.txt
```diff
@@ -59,6 +88,23 @@ target_link_libraries(
   fd_objects_common PRIVATE Bareos::Lib Bareos::Findlib Bareos::Fastlz
                             ZLIB::ZLIB
 )
+if(fd-client-sleep-inhibition)
+  target_compile_definitions(fd_objects_common
+                             PRIVATE HAVE_FD_CLIENT_SLEEP_INHIBITION=1
+  )
```
sebsura: Why do we not add this to config.h instead ?

core/src/filed/CMakeLists.txt
```diff
@@ -59,6 +88,23 @@ target_link_libraries(
...
+  target_compile_options(fd_objects_common PRIVATE ${SYSTEMD_CFLAGS_OTHER})
+  if(SYSTEMD_LIBRARIES)
+    target_link_libraries(fd_objects_common PRIVATE ${SYSTEMD_LIBRARIES})
+  else()
+    target_link_libraries(fd_objects_common PRIVATE systemd)
+  endif()
+endif()
+if(HAVE_DARWIN_OS)
+  target_link_libraries(fd_objects_common PRIVATE "-framework IOKit")
+endif()
```
sebsura, line 117: Do we need this even without fd-client-sleep-inhibition ?

core/src/filed/CMakeLists.txt
```diff
@@ -59,6 +88,23 @@ target_link_libraries(
...
+if(HAVE_SYSTEMD AND fd-client-sleep-inhibition)
+  target_include_directories(fd_objects_common PRIVATE ${SYSTEMD_INCLUDE_DIRS})
+  target_compile_options(fd_objects_common PRIVATE ${SYSTEMD_CFLAGS_OTHER})
+  if(SYSTEMD_LIBRARIES)
+    target_link_libraries(fd_objects_common PRIVATE ${SYSTEMD_LIBRARIES})
+  else()
+    target_link_libraries(fd_objects_common PRIVATE systemd)
+  endif()
+endif()
```
sebsura: Why is this not just part of the upper if ? I do not get why this pr has so many
```
if (x & a)
if (x & b)
if (x & c)
```

instead of

```
if (x)
  if (a)
  if (b)
  if (c)
```

core/src/filed/dir_cmd.cc
```diff
@@ -390,6 +539,28 @@ void* process_director_commands(JobControlRecord* jcr, BareosSocket* dir)
...
+        ActivateLinuxSleepInhibition(jcr, linux_sleep_inhibitor_fd,
+                                     linux_sleep_inhibit_warning_logged);
+      }
+#endif
+#if defined(HAVE_FD_CLIENT_SLEEP_INHIBITION) && defined(HAVE_DARWIN_OS)
+      if (is_backup_or_restore_command) {
+        ActivateDarwinSleepInhibition(jcr, darwin_sleep_assertion_id,
+                                      darwin_sleep_inhibit_warning_logged);
+      }
+#endif
```
sebsura: Why does only windows care whether sleep is already prevented ?

Why is this not just
```
#if HAVE_FD_CLIENT_SLEEP_INHIBITION
if (is_backup_or_restore_command) {
  PreventOsSuspensions();
}
#endif
```
 And simply define `PreventOsSuspension()` depending on the target.  We should probably just move everything in `filed/os_suspension.h` + an impl file per platform.

core/src/filed/dir_cmd.cc
```diff
@@ -438,9 +609,16 @@ void* process_director_commands(JobControlRecord* jcr, BareosSocket* dir)
...
+#if defined(HAVE_FD_CLIENT_SLEEP_INHIBITION) && defined(HAVE_WIN32)
   AllowOsSuspensions();
 #endif
+#if defined(HAVE_FD_CLIENT_SLEEP_INHIBITION) && defined(HAVE_LINUX_OS) \
+    && defined(HAVE_SYSTEMD)
+  DeactivateLinuxSleepInhibition(linux_sleep_inhibitor_fd);
+#endif
+#if defined(HAVE_FD_CLIENT_SLEEP_INHIBITION) && defined(HAVE_DARWIN_OS)
+  DeactivateDarwinSleepInhibition(darwin_sleep_assertion_id);
+#endif
```
sebsura: Same here.  This should just be
```
#if ...
  AllowOsSuspensions();
#endif
```

core/src/filed/filed.cc
```diff
@@ -52,6 +52,30 @@ using namespace filedaemon;
...
+
+  return has_outbound_director && !has_inbound_director;
+}
+
+static void WaitUntilTerminated()
+{
+  // Without a listening socket server we still need to keep the daemon process
+  // alive so client-initiated connection threads can run.
+  for (;;) { Bmicrosleep(30, 0); }
+}
```
sebsura, line 174: There has to be a better way to do this.  What condition are we actually waiting for ? If we know that, we can actually make the code wait on that condition (e.g. via a condition_variable, an atomic_flag, ....).

core/src/filed/dir_cmd.cc
```diff
@@ -362,6 +498,19 @@ static s_fd_dir_cmds* SelectCommandByName(const char* name)
...
+#if defined(HAVE_FD_CLIENT_SLEEP_INHIBITION) && defined(HAVE_LINUX_OS) \
+    && defined(HAVE_SYSTEMD)
+  int linux_sleep_inhibitor_fd = -1;
+  bool linux_sleep_inhibit_warning_logged = false;
+#endif
+#if defined(HAVE_FD_CLIENT_SLEEP_INHIBITION) && defined(HAVE_DARWIN_OS)
+  IOPMAssertionID darwin_sleep_assertion_id = kIOPMNullAssertionID;
+  bool darwin_sleep_inhibit_warning_logged = false;
+#endif
```
sebsura: If the sleep prevention needs some kind of state, then we should have one `SleepPrevention` struct or so (with one implementation per os) instead of this.

sebsura (changes requested):

core/src/filed/CMakeLists.txt
sebsura: I guess this is only relevant for the fd.

core/src/filed/CMakeLists.txt
pstorz: Renamed to FD_CLIENT_SLEEP_INHIBITION

core/src/filed/CMakeLists.txt
pstorz: moved to BareosCheckIncludes.cmake

core/src/filed/CMakeLists.txt
pstorz: Yes, it is only relevant for the FD. Renamed to FILED_CLIENT_SLEEP_INHIBITION

core/src/filed/CMakeLists.txt
pstorz: Good point, changed to HAVE_DARWIN_OS AND FD_CLIENT_SLEEP_INHIBITION

core/src/filed/CMakeLists.txt
pstorz: Ok, good point, was refactored.

core/src/filed/dir_cmd.cc
pstorz: Good point, os_suspension.h and os_suspension_{linux,darwin,win32}.cc were added.

core/src/filed/filed.cc
pstorz: That makes sense. Was changed to condition variable.

core/src/filed/dir_cmd.cc
pstorz: Also, now all OS check for the current state in the corresponding files.

core/src/filed/dir_cmd.cc
pstorz: That was addressed in the files mentioned above.

sebsura (changes requested):
There are merge conflicts

core/src/filed/CMakeLists.txt
```diff
@@ -55,10 +78,43 @@ target_sources(
...
+if(FD_CLIENT_SLEEP_INHIBITION
+   AND HAVE_LINUX_OS
+   AND HAVE_SYSTEMD
+)
+  target_sources(fd_objects_common PRIVATE os_suspension_linux.cc)
+endif()
+if(FD_CLIENT_SLEEP_INHIBITION AND HAVE_DARWIN_OS)
+  target_sources(fd_objects_common PRIVATE os_suspension_darwin.cc)
+endif()
+
```
sebsura: why is this not part of the if below ? And this still has the weird ifs.

core/src/filed/os_suspension.h
```diff
@@ -0,0 +1,72 @@
...
+                                  SleepPrevention& sleep_prevention);
+void DeactivateLinuxSleepPrevention(SleepPrevention& sleep_prevention);
+#endif
+
+#if defined(FILED_CLIENT_SLEEP_INHIBITION) && defined(HAVE_DARWIN_OS)
+void ActivateDarwinSleepPrevention(JobControlRecord* jcr,
+                                   SleepPrevention& sleep_prevention);
+void DeactivateDarwinSleepPrevention(SleepPrevention& sleep_prevention);
+#endif
+
```
sebsura: Please do not repeat the same #if stuff all over. Either use a single
```
#if defined(FILED_CLIENT_SLEEP_INHIBITION)
```
at the top of the file, or let the build system only include this file when that is set.

sebsura: Also why does each operating system need a differently called function ? Just call them all the same.

core/src/filed/filed.cc
sebsura: This is still a big loop.  Where is the condition variable that you talked about ?

sebsura (changes requested):

core/src/filed/CMakeLists.txt
pstorz: Has been refactored

core/src/filed/os_suspension.h
pstorz: Was addressed

core/src/filed/CMakeLists.txt
```diff
@@ -55,10 +77,44 @@ target_sources(
...
+      target_link_libraries(fd_objects_common PRIVATE ${SYSTEMD_LIBRARIES})
+    else()
+      target_link_libraries(fd_objects_common PRIVATE systemd)
+    endif()
+  endif()
+
+  if(HAVE_DARWIN_OS)
+    target_link_libraries(fd_objects_common PRIVATE "-framework IOKit")
+  endif()
+endif()
```
sebsura, line 117: Why is this seperate from the target_sources calls above ?

systemtests/tests/client-initiated/testrunner
```diff
@@ -24,6 +24,39 @@ exit_with_error()
...
+        sock.settimeout(0.5)
+        rc = sock.connect_ex((host, port))
+        sock.close()
+        if rc == 0:
+            print(f"FD listener accepts connections on {host}:{port}")
+            sys.exit(1)
+    except OSError:
+        pass
+
+sys.exit(0)
```
sebsura, line 51: Please put this into its own python file.  It would be much easier to read&maintain if you get syntax highlighting etc for this.

core/cmake/BareosCheckIncludes.cmake
```diff
@@ -21,4 +21,12 @@ if(NOT MSVC)
...
   check_include_files("sys/types.h;sys/acl.h" HAVE_SYS_ACL_H)
   check_include_files(sys/capability.h HAVE_SYS_CAPABILITY_H)
+
+  if(HAVE_SYSTEMD AND HAVE_LINUX_OS)
+    set(_previous_required_includes "${CMAKE_REQUIRED_INCLUDES}")
+    set(CMAKE_REQUIRED_INCLUDES "${SYSTEMD_INCLUDE_DIRS}")
+    check_include_files("systemd/sd-bus.h" HAVE_SYSTEMD_SD_BUS_H)
+    set(CMAKE_REQUIRED_INCLUDES "${_previous_required_includes}")
+    unset(_previous_required_includes)
+  endif()
```
sebsura, line 31: In findxxhash, we use
    cmake_push_check_state()
and
    cmake_pop_check_state()

We should keep stuff consistent if possible.

core/src/filed/CMakeLists.txt
```diff
@@ -29,6 +33,24 @@ if(HAVE_WIN32)
...
+    )
+  endif()
+  if(NOT HAVE_SYSTEMD_SD_BUS_H)
+    message(
+      FATAL_ERROR
+        "FD_CLIENT_SLEEP_INHIBITION requires <systemd/sd-bus.h> on Linux. "
+        "Install systemd development headers or set "
+        "-DFD_CLIENT_SLEEP_INHIBITION=OFF to disable manually."
+    )
+  endif()
```
sebsura, line 51: This lhould happen where we add os_suspension_linux.cc.  I do not see any upside to splitting this up into X parts all over this file.

core/src/filed/dir_cmd.cc
```diff
@@ -391,6 +394,12 @@ void* process_director_commands(JobControlRecord* jcr, BareosSocket* dir)
       }
 
       Dmsg1(100, "Executing %s command.\n", to_execute->cmd);
+      const bool is_backup_or_restore_command
+          = (to_execute->func == BackupCmd || to_execute->func == RestoreCmd);
```
sebsura, line 398: Please add a field to the cmd strutc (e.g. bool prevent_os_suspension), and then check that, as that is the point of that struct.

core/src/filed/os_suspension_darwin.cc
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+  sleep_prevention.darwin_assertion_id = assertion_id;
+}
+
+void DeactivateSleepPrevention(SleepPrevention& sleep_prevention)
+{
+  IOPMAssertionID assertion_id
+      = static_cast<IOPMAssertionID>(sleep_prevention.darwin_assertion_id);
+  if (assertion_id != kIOPMNullAssertionID) {
```
sebsura, line 68: https://github.com/pstorz/bareos/blob/13ffefc3b069368eced45ce9fd9de2c054c3bf9a/core/src/filed/os_suspension_darwin.cc#L55

We set this to 0 in Activate, but compare to kIOPMNullAssertionID here.  These should be the same.

core/src/filed/os_suspension_darwin.cc
```diff
@@ -0,0 +1,74 @@
...
+    IOPMAssertionRelease(assertion_id);
+    sleep_prevention.darwin_assertion_id = 0;
```
sebsura, line 70: Why are we setting it to 0 here again ?

sebsura: This should either be static_cast<...>(kIOPNullAssertionID), or even better this should just be the correct type.

core/src/filed/os_suspension.h
```diff
@@ -0,0 +1,55 @@
...
+#  if defined(HAVE_DARWIN_OS)
+  uintptr_t darwin_assertion_id = 0;
+  bool darwin_warning_logged = false;
+#  endif
+#endif
+};
+
+void ActivateSleepPrevention(JobControlRecord* jcr,
+                             SleepPrevention& sleep_prevention);
+void DeactivateSleepPrevention(SleepPrevention& sleep_prevention);
```
sebsura, line 51: We should not have this ifdef stuff here.  The functions should probably look like:

```
struct SleepPrevention;
SleepPrevention* ActivateSleepPrevention(jcr);
void DeactiveateSleepPrevention(SleepPrevention*);
```

This would allow us to use the correct types in sleep prevention (see e.g. darwin comments).

core/src/filed/os_suspension_linux.cc
```diff
@@ -0,0 +1,115 @@
...
+
+  sd_bus* bus = nullptr;
+  sd_bus_message* reply = nullptr;
+  sd_bus_error error = SD_BUS_ERROR_NULL;
+
+  auto cleanup = [&]() {
+    sd_bus_error_free(&error);
+    sd_bus_message_unref(reply);
+    sd_bus_unref(bus);
+  };
```
sebsura, line 59: Please use unique_ptr here

core/src/filed/os_suspension_linux.cc
```diff
@@ -0,0 +1,115 @@
...
+    return;
+  }
+
+  int dupfd = fcntl(fd, F_DUPFD_CLOEXEC, 3);
+  if (dupfd < 0) {
+    BErrNo be;
+    warn_and_return(be.bstrerror());
+    return;
+  }
+
```
sebsura, line 102: So what happens to fd ? Do we not want to close that ?

core/src/filed/os_suspension_win32.cc
```diff
@@ -0,0 +1,45 @@
...
+
+#include "compat.h"
+
+namespace filedaemon {
+
+void ActivateSleepPrevention(JobControlRecord*,
+                             SleepPrevention& sleep_prevention)
+{
+  if (!sleep_prevention.windows_active) {
+    PreventOsSuspensions();
```
sebsura, line 32: This does not make sense.   These functions are defined in compat.cc.  Please delete them and insert their contents here.

core/src/filed/filed.cc
```diff
@@ -49,12 +50,129 @@
...
 #include "lib/address_conf.h"
 #include "lib/alist.h"
+#include "lib/berrno.h"
 
 using namespace filedaemon;
 
 /* Imported Functions */
 extern void* handle_connection_request(void* dir_sock);
 
+static bool use_signal_pipe_termination = false;
```
sebsura, line 60: This would be much easier to read if there was only a single #ifdef HAVE_WIN32.

sebsura (changes requested):

PR #2720, opened 2026-07-09
dird: remove unused write verify list

sebsura:
This removes the `WriteVerifyList` option in the directors job resource, which was completely unused except for statements printing it.

#### Please check

- [x] Short description and the purpose of this PR is present _above this paragraph_
- [x] Your name is present in the AUTHORS file (optional)

If you have any questions or problems, please give a comment in the PR.

### Helpful documentation and best practices

- [Git Workflow](https://docs.bareos.org/DeveloperGuide/gitworkflow.html)
- [Automatic Sourcecode Formatting](https://docs.bareos.org/DeveloperGuide/generaldevel.html#automatic-sourcecode-formatting)
- [Check your commit messages](https://docs.bareos.org/DeveloperGuide/gitworkflow.html#commits)
- [Boy Scout Rule](https://docs.bareos.org/DeveloperGuide/generaldevel.html#boy-scout-rule)

### Checklist for the _reviewer_ of the PR (will be processed by the Bareos team)
Make sure you check/merge the PR using `devtools/pr-tool` to have some simple automated checks run and a proper changelog record added.

##### General
- [x] Is the PR title usable as CHANGELOG entry?
- [x] Purpose of the PR is understood
- [x] Commit descriptions are understandable and well formatted
- [x] Required backport PRs have been created
- If a release should wait for this PR to be finished, set that release's milestone.

##### Source code quality
- [x] Source code changes are understandable
- [x] Variable and function names are meaningful
- [x] Code comments are correct (logically and spelling)
- [x] Required documentation changes are present and part of the PR

PR #2721, opened 2026-07-09
core: make sure to cross check tls-psk with claimed identity in auth

sebsura:
Currently tls and authentication are completely independent.  This is not a problem currently as TLS in and off itself is not trusted at all from an authentication stand point.

This PR changes this, so that at least for tls-psk, the used identity is reported back to the authentication routine, where it cross checks it with the claimed identity in the authentication step.

#### Please check

##### General

##### Source code quality

##### Tests


sebsura:
@coderabbitai full review

core/src/lib/try_tls_handshake_as_a_server.h
```diff
@@ -22,9 +22,151 @@
...
+    return std::nullopt;
+  }
+
+  JobControlRecord* get_job_for_used_secret() const { return found_jcr; }
+
+  bool psk_used() const
+  {
+    return get_job_for_used_secret() != nullptr
+           || get_resource_for_used_secret() != nullptr;
+  }
```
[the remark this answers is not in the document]
sebsura: When a jcr based psk is accepted, the resource is not set, so `check_resource_name()` should return false. Can you confirm this ?

core/src/dird/ua_cmds.cc
```diff
@@ -440,7 +440,6 @@ static struct ua_cmdstruct commands[] = {
          "backupclient=<client-name> backupformat=<format> "
          "nextpool=<pool-name> "
          "since=<universal-time-specification> verifyjob=<job-name> "
-         "verifylist=<verify-list> "
```
pstorz, line 443: Why is the verify list not needed anymore?

core/src/filed/dir_cmd.cc
```diff
@@ -1436,11 +1436,8 @@ static bool StorageCmd(JobControlRecord* jcr)
   jcr->store_bsock = storage_daemon_socket;
 
   if (tls_policy == TlsPolicy::kBnetTlsAuto) {
-    std::string qualified_resource_name;
-    if (!my_config->GetQualifiedResourceNameTypeConverter()->ResourceToString(
-            jcr->Job, R_JOB, qualified_resource_name)) {
-      goto bail_out;
-    }
+    std::string qualified_resource_name
+        = global_resource::QualifiedName(global_resource::Type::Job, jcr->Job);
```
pstorz, line 1440: Why is there no check for failure anymore?

core/src/lib/try_tls_handshake_as_a_server.h
```diff
@@ -22,9 +22,151 @@
...
+      output[pw_len] = 0;
+      chosen_resource = res;
+      return pw_len;
+    } else {
+      Dmsg1(100, "Could not get tls resource for %d.\n", r_type);
+      return 0;
+    }
+  }
+
+  std::optional<std::string> check_resource_name(uint32_t resource_type,
```
pstorz: Do I understand right, that std::nullopt is the "success"  result?

core/src/lib/try_tls_handshake_as_a_server.h
```diff
@@ -22,9 +22,151 @@
...
+  JobControlRecord* found_jcr{nullptr};
+
+  unsigned int get_shared_secret_for(
+      std::string_view identity,
+      std::span<unsigned char> psk_buffer) override
+  {
+    auto [type, name] = global_resource::ParseQualifiedName(identity);
+
+    if (type == global_resource::Type::Job) {
+      std::string cpy{name};
```
pstorz: Is the temporarly variable cpy needed?

core/src/lib/try_tls_handshake_as_a_server.h
```diff
@@ -22,9 +22,151 @@
...
+      auto* jcr = get_jcr_by_full_name(cpy.c_str());
+      if (!jcr) { return 0; }
```
pstorz: Would it make sense to have both error exits (114 and118) distinguishable?

sebsura:
@coderabbitai full review

core/src/dird/ua_cmds.cc
sebsura: We removed it in a pr #2720, but that string was overlooked.

core/src/filed/dir_cmd.cc
sebsura: Because the function cannot fail anymore. It was failable before, because the config parser had to be manually told about which types existed, and if you did not do that, then the function would fail.

core/src/lib/try_tls_handshake_as_a_server.h
sebsura: Yes. The function returns an error (if any).  There may be a way to better communicate this.  Ill have a think about this (unless you already have an alternative in mind).

core/src/lib/try_tls_handshake_as_a_server.h
sebsura: Currently yes, since `name` is a `string_view` (i.e. may not be NUL terminated), and the jcr find function needs a c-string.  I will probably add an overload to that function taking in a `string_view` to fix this though.

core/src/lib/try_tls_handshake_as_a_server.h
sebsura: Distunguishable by whom ?
* The client => definitely not,
* The director => possibly, but im not sure what it would do with that information
* The user => yes, it probably makes sense to add some debug messages here, so one can understand what went wrong.

core/src/lib/try_tls_handshake_as_a_server.h
pstorz: Doing a bit of research, std::expected (c++23) would be exactly what we need here. Otherwise probably std::variant is a better approach.

core/src/lib/jcr.cc
```diff
@@ -602,6 +600,13 @@ JobControlRecord* get_jcr_by_full_name(char const* Job)
   return jcr;
 }
 
+JobControlRecord* get_jcr_by_full_name(char const* Job)
```
pstorz, line 603: I think we usually use `const char*` instead of `char const*`

core/src/lib/try_tls_handshake_as_a_server.h
```diff
@@ -22,9 +22,202 @@
...
+      auto* jcr = get_jcr_by_full_name(name);
+      if (!jcr) {
+        Dmsg1(100,
+              "Login attempt via job %.*s, but no such job is known to me\n",
+              (int)name.size(), name.data());
+        return 0;
+      }
+      found_jcr = jcr;
+
+      auto key_length = strlen(jcr->sd_auth_key);
```
pstorz: strlen(jcr->sd_auth_key)  can dereference a null pointer

pstorz (commented):
See comments. Also, I think the coderabbit also has some remarks.

core/src/lib/try_tls_handshake_as_a_server.h
sebsura: We can change it to std::expected once we switch over to c++23, but i am not a fan at all of using std::variant for this.

If you want to have this changed, then I can think of something else, but otherwise I would leave it like this.

core/src/lib/jcr.cc
sebsura: fixed

core/src/lib/try_tls_handshake_as_a_server.h
sebsura: fixed

core/src/lib/try_tls_handshake_as_a_server.h
```diff
@@ -22,9 +22,214 @@
...
+              "Login attempt via job %.*s, but no such job is known to me\n",
+              (int)name.size(), name.data());
+        return 0;
+      }
+      if (!jcr->sd_auth_key) {
+        Dmsg1(100, "Login attempt via job %.*s, but that job has no auth key\n",
+              (int)name.size(), name.data());
+        return 0;
+      }
+      if (strcmp(jcr->sd_auth_key, "dummy")) {
```
pstorz:
```suggestion
      if (!strcmp(jcr->sd_auth_key, "dummy")) {
```

core/src/lib/try_tls_handshake_as_a_server.h
```diff
@@ -22,9 +22,214 @@
...
+        Dmsg1(100,
+              "Login attempt via job %.*s, but that job has no real auth key\n",
+              (int)name.size(), name.data());
+        return 0;
+      }
+
+      found_jcr = jcr;
```
pstorz: This line should be moved between Lines 143 and 144 to avoid a use-count leak of the jcr, as  get_jcr_by_full_name()  increments the refcount before  found_jcr  is set.

pstorz (changes requested):
Please see the two comments.

core/src/lib/try_tls_handshake_as_a_server.h
sebsura: I switched up bstrcmp and strcmp.  We should probably remove that one at some point.

core/src/lib/try_tls_handshake_as_a_server.h
sebsura: done

pstorz (approved):
Good work!

PR #2739, opened 2026-07-30
Reject external post-login redirects

acts-1631:
The WebUI login return path accepted an absolute req URL. The route check
validated only its path, while the subsequent redirect used the original
absolute URL.

This allowed a URL such as https://attacker.example/dashboard to pass the
local /dashboard route check and redirect users off-site after login.

Require an origin-relative path and reject network-path, backslash, and
encoded path-separator forms before route matching and redirecting.

### DIFF
--- a/webui/module/Auth/src/Auth/Controller/AuthController.php
+++ b/webui/module/Auth/src/Auth/Controller/AuthController.php
@@ -167,12 +167,12 @@ class AuthController extends AbstractActionController
             $session->offsetSet('commands', $aclcheck);
         }
 
-        if ($this->params()->fromQuery('req')) {
-            $redirect = $this->params()->fromQuery('req');
+        $redirect = $this->params()->fromQuery('req');
+        if ($this->isLocalRedirect($redirect)) {
             $request = $this->getRequest();
             $request->setUri($redirect);
             if ($routeToBeMatched = $this->getServiceLocator()->get('Router')->match($request)) {
-                return $this->redirect()->toUrl($this->params()->fromQuery('req'));
+                return $this->redirect()->toUrl($redirect);
             }
         }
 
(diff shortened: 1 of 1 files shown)


webui/module/Auth/src/Auth/Controller/AuthController.php
```diff
@@ -167,12 +167,12 @@ public function loginAction()
             $session->offsetSet('commands', $aclcheck);
         }
 
-        if ($this->params()->fromQuery('req')) {
```
bruno-at-bareos, line 170: With if no more present, what will be the result if req is not present ?

webui/module/Auth/src/Auth/Controller/AuthController.php
```diff
@@ -243,6 +243,15 @@ private function clearAuthenticationSessionState()
         }
     }
 
+    private function isLocalRedirect($redirect)
```
bruno-at-bareos, line 246: Would be nice to extent systemtest

webui/module/Auth/src/Auth/Controller/AuthController.php
```diff
@@ -243,6 +243,15 @@ private function clearAuthenticationSessionState()
...
+    {
+        return is_string($redirect)
+            && substr($redirect, 0, 1) === '/'
+            && substr($redirect, 0, 2) !== '//'
+            && strpos($redirect, '\\') === false
+            && !preg_match('/%(2f|5c)/i', $redirect);
+    }
+
```
bruno-at-bareos, line 254: could you check and implement this recommendation.

bruno-at-bareos (changes requested):
Thanks for your proposal.
We would like to see few adjustments taking place, before going further. 
- First of all, this webui will become legacy in next 26 version and only really few efforts will be done in the future.
- Could you re-add the default template for the PR, which is the normal complete workflow which has to be be followed to get it merged.
- Can you check, how you can extent the systemtest to cover your changes ( Failed first without the code change, Fixed with the changes) ?

Thanks.

bruno-at-bareos:
@acts-1631 any news on this?

arogge:
No feedback, closing. Feel free to provide feedback and reopen!

PR #2740, opened 2026-07-30
Require CSRF confirmation for job cancellation

acts-1631:
GET /job/cancel/<id> immediately issued a Director cancel command after
checking only the authenticated session and normal command ACL.

A cross-site top-level navigation could therefore cancel a job using a
logged-in operator's authority.

GET now displays a confirmation form. The cancellation command is issued
only after a POST with a valid per-session CSRF token.

### DIFF
--- a/webui/module/Job/src/Job/Controller/JobController.php
+++ b/webui/module/Job/src/Job/Controller/JobController.php
@@ -29,6 +29,7 @@ use Laminas\Mvc\Controller\AbstractActionController;
 use Laminas\View\Model\ViewModel;
 use Laminas\Json\Json;
 use Job\Form\JobForm;
+use Job\Form\CancelJobForm;
 use Job\Form\RunJobForm;
 use Job\Model\Job;
 use Exception;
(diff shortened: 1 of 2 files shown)


bruno-at-bareos (changes requested):
Thanks for your proposal!
We would like to see few adjustments taking place, before going further.
- First of all, this webui will become legacy in next 26 version and only really few efforts will be done in the future.
- Can you check, how you can extent the systemtest to cover your changes ( Failed first without the code change, Fixed with the changes) ?
- Take into account the report of CodeRabbit about your csrf name.
- Could you re-add the default template for the PR, which is the normal complete workflow which has to be be followed to get it merged.


Thanks.

bruno-at-bareos:
@acts-1631 any news on this?

acts-1631:
Since the WebUI will be deprecated, you can close the issue(s) if you want. Probably not worth working on.

bruno-at-bareos:
Thanks for your feedback.

PR #2748, opened 2026-08-05
webui-vue: fix API/WS base URL when deep-linking or refreshing sub-routes

### LINKED ISSUE #2747: webui-new: API/WS base URL is computed from the current route path, breaking "Director list request failed" on any deep-linked or refreshed sub-route
**Environment**
- Bareos version: 25.1.0 (self compiled) (bareos-webui-new, bareos-webui-proxy, bareos-director)
- OS: Debian 13
- Web server: Apache 2.4.67, mod_proxy + mod_proxy_wstunnel, using the shipped bareos-webui-new.conf reverse-proxy config unmodified
- Deployment: single director, reached at /bareos-webui-new/ via Apache, proxying to bareos-webui-proxy on localhost:9104

**Summary**
Loading (or refreshing) the WebUI on any route other than the exact app root, e.g. /bareos-webui-new/dashboard, permanently fails with:

▎ Director list request failed: Failed to load directors

The app root (/bareos-webui-new/, trailing slash, no sub-route) works fine and logs in correctly. The bug is purely in how the frontend computes its own API base URL, not in the proxy or reverse-proxy config.

**Steps to reproduce**
1. Deploy 25.1.0 with bareos-webui-new + bareos-webui-proxy behind Apache using the packaged bareos-webui-new.conf.
2. Confirm the proxy itself is healthy: curl http://localhost/bareos-webui-new/api/directors returns {"type":"director_list","directors":["<name>"]}.
3. Open http://\<host\>/bareos-webui-new/dashboard directly in a browser (or refresh the page while on any sub-route, e.g. after client-side navigation to /clients, /jobs, etc.).
4. Observe the permanent "Director list request failed: Failed to load directors" banner; login is impossible from that URL.

**Expected behavior**
The director list (and all API/WS calls) should resolve relative to the app's mount path (/bareos-webui-new/) regardless of which route is currently displayed.

**Actual behavior / root cause**
In the built bundle (assets/settings-*.js), the API/WS base is computed once at module load from window.location.pathname, i.e. the current route's path, not the app's mount path:

function kt(){ let e = window.location.pathname || "/"; return e.endsWith("/") ? e : e + "/" }
function At(){ let e = window.location.protocol === "https:" ? "wss:" : "ws:"; let t = kt(); return `${e}//${window.location.host}${t}ws` }
function jt(){ let e = kt(); return `${window.location.protocol}//${window.location.host}${e}api` }
var Mt = At(), Nt = jt();

kt() just appends a trailing slash to the raw pathname instead of stripping back to the app's mount segment. So:

- At /bareos-webui-new/ → base = /bareos-webui-new/ → API = .../bareos-webui-new/api ✅
- At /bareos-webui-new/dashboard → base = /bareos-webui-new/dashboard/ → API = .../bareos-webui-new/dashboard/api ❌

The reverse proxy (correctly, per the shipped Apache config) only proxies /bareos-webui-new/api/ and /bareos-webui-new/ws, so the miscomputed URL falls through to the SPA catch-all rewrite and gets index.html back (200, text/html) instead of JSON. The frontend's fetch(...).json().catch(() => null) then yields null, tripping the "Failed to load directors" error path.

**Reproducible directly with curl:**
$ curl -s -i http://localhost/bareos-webui-new/api/directors
HTTP/1.1 200 OK
Content-Type: application/json
{"type":"director_list","directors":["bareos-dir"]}

$ curl -s -i http://localhost/bareos-webui-new/dashboard/api/directors
HTTP/1.1 200 OK
Content-Type: text/html
<!DOCTYPE html>...

The same issue affects the WebSocket base (At()/Mt), so any deep-linked/refreshed route likely also breaks the live-update websocket, not just the initial director list.

**Suggested fix**
Compute the API/WS base from the app's known mount path (e.g. Vite's import.meta.env.BASE_URL, fixed at build time) rather than from window.location.pathname of whatever route the user happens to be on. Alternatively, derive it from the router's configured base rather than the live browser URL.

**Workaround**
Only navigate to the app at its exact root (/bareos-webui-new/, trailing slash); any deep link or page refresh on a sub-route breaks the WebUI.

swenske:
Fixes #2747

`getAppBasePath()` derived the app's base path from `window.location.pathname`, which reflects whatever route the user is currently on rather than the path the app is actually deployed under. When a user opened or refreshed a deep link (e.g. `/bareos-webui-new/dashboard`), the computed base became `/bareos-webui-new/dashboard/` instead of `/bareos-webui-new/`, so `API_BASE_URL` and the director WebSocket URL pointed at a non-existent proxy path (`/bareos-webui-new/dashboard/api`, `.../dashboard/ws`). The reverse proxy doesn't recognize that path and returns HTML instead of JSON, which surfaces as a persistent "Director list request failed" error and blocks login.

Since the app already uses `createWebHashHistory()`, the browser path never actually changes during in-app navigation, only routes reached via a fresh page load (deep link or refresh) hit this bug. The fix uses Vite's `import.meta.env.BASE_URL`, which is fixed at build time to the app's real deployment path, instead of the mutable current URL. `API_BASE_URL` and `DIRECTOR_WS_URL` are computed once from this in `directorCommandSocket.js` and consumed everywhere else, so the fix is centralized in a single function.

Testing:

`npx vitest run`: 275/276 tests pass; the one failure (`i18nGeneration.spec.js`) is a pre-existing, unrelated stale-generated-file issue with no dependency on this code.

### DIFF
--- a/webui-vue/src/utils/directorCommandSocket.js
+++ b/webui-vue/src/utils/directorCommandSocket.js
@@ -20,8 +20,8 @@
  */
 
 export function getAppBasePath() {
-  const path = window.location.pathname || '/'
-  return path.endsWith('/') ? path : path + '/'
+  const base = import.meta.env.BASE_URL || '/'
+  return base.endsWith('/') ? base : base + '/'
 }
 
 export function defaultDirectorWsUrl() {


swenske:
fixed in #2754

PR #2779, opened 2026-08-25
dplcompat: fix SIGABRT when a chunk deletion fails during truncate

### LINKED ISSUE #2778: sd: dplcompat backend aborts (SIGABRT) when a chunk deletion fails during volume truncate
### Bareos component version

```
bareos-sd 25.1.1~pre20.a5aedd6bf-99
```

The bug is also present in current `master`.

### Steps to reproduce

1. Configure a `dplcompat` (Droplet-compatible / S3) device on the storage daemon, backed by an object store (in our case Backblaze B2).
2. Set `Action On Purge = Truncate` on the pool using that device, so that purging a volume invokes `DropletCompatibleDevice::TruncateRemoteVolume()`.
3. Arrange for one `remove` (chunk deletion) call to the object store to fail while a volume is being truncated — e.g. a transient backend error, or a chunk that no longer exists in the bucket.

The failure is in the *error-reporting* path, so any condition that makes a single chunk delete fail is sufficient to trigger it.

### Expected results

The chunk deletion failure is reported via `errmsg`, `TruncateRemoteVolume()` returns `false`, and the storage daemon continues running.

### Actual results

The storage daemon aborts with `SIGABRT` on an assertion inside `tl::expected`, killing all in-flight jobs.

The cause is in `core/src/stored/backends/dplcompat_device.cc`, in `TruncateRemoteVolume()`:

https://github.com/bareos/bareos/blob/a5aedd6bf060a8597641ab02117c350a1d7bb8a0/core/src/stored/backends/dplcompat_device.cc#L365-L384

```cpp
  const auto chunk_map = m_storage.list(vol_name);
  if (!chunk_map) {
    PmStrcpy(errmsg, chunk_map.error().c_str());   // ok, chunk_map is unexpected here
    dev_errno = EIO;
    return false;
  }
  for (const auto& [chunk_name, stat] : *chunk_map) {
    if (is_chunk_name(chunk_name)) {
      if (auto res = m_storage.remove(vol_name, chunk_name); !res) {
        PmStrcpy(errmsg, chunk_map.error().c_str());   // <-- should be res.error()
        dev_errno = EIO;
        return false;
      }
    }
  }
```

Inside the loop, `chunk_map` necessarily *holds a value* (it is being iterated). `tl::expected::error()` is defined as:

```cpp
constexpr const E &error() const & {
  TL_ASSERT(!has_value());
  return err().value();
}
```

so calling `chunk_map.error()` there trips the assertion and aborts the process. On a build with assertions disabled it would instead be undefined behaviour, reading the error member of an engaged `expected`.

The intent was clearly to report the error from `res`, the failed `remove()` call. This looks like a copy-paste slip from the guarded `chunk_map.error()` a few lines above.

### Suggested fix

```diff
-        PmStrcpy(errmsg, chunk_map.error().c_str());
+        PmStrcpy(errmsg, res.error().c_str());
```

The line has been present since the backend was introduced; commit 61fde89b ("dplcompat: rename list -> chunk_map", 2024-12-11) carried it over unchanged when the variable was renamed.

### Environment

- OS: Debian-based (dpkg)
- component: `bareos-sd`, `dplcompat` storage backend (`stored/backends/dplcompat_device.cc`)

### Relevant log output

```shell
bareos-sd[NNNNN]: bareos-sd: /usr/include/tl/expected.hpp:2014: constexpr const E& tl::expected<T, E>::error() const & [with T = std::map<std::__cxx11::basic_string<char>, CrudStorage::Stat>; E = std::__cxx11::basic_string<char>]: Assertion `!has_value()' failed.
bareos-sd[NNNNN]: BAREOS interrupted by signal 6: IOT trap
bareos-sd[NNNNN]: bareos-sd, example-sd got signal 6 - IOT trap. Attempting traceback.
bareos-sd[NNNNN]: exepath=/usr/sbin/
bareos-sd[NNNNN]: BAREOS interrupted by signal 6: IOT trap
bareos-sd[NNNNN]: Calling: /usr/sbin/btraceback /usr/sbin/bareos-sd NNNNN /var/lib/bareos
bareos-sd[NNNNN]: It looks like the traceback worked...
bareos-sd[NNNNN]: Dumping: /var/lib/bareos/example-sd.NNNNN.bactrace
systemd[1]: bareos-storage.service: Main process exited, code=exited, status=1/FAILURE
systemd[1]: bareos-storage.service: Failed with result 'exit-code'.
```

The template parameter `T = std::map<std::string, CrudStorage::Stat>` in the assertion message identifies the return type of `CrudStorage::list()`, which pins the crash to one of the `m_storage.list()` result sites; `TruncateRemoteVolume()` is the only one where `.error()` is reached with the `expected` in a valid state.

### Anything else?

`RemoteVolumeSize()`, immediately below, has a related but harmless cosmetic issue: it returns `false` on the error path from a function returning `ssize_t`, which yields `0` rather than an error indication like `-1`.

jsfrerot:
Fixes #2778

`TruncateRemoteVolume()` calls `chunk_map.error()` inside the chunk-deletion loop, but the object that actually failed is `res`. At that point `chunk_map` is engaged, so `tl::expected::error()` trips its `TL_ASSERT(!has_value())` and the storage daemon aborts with `SIGABRT` instead of reporting the error and returning `false`:

```
Assertion !has_value() failed at .../expected.hpp
  [T = std::map<std::string, CrudStorage::Stat>]
```

Any single failed chunk delete against the object store therefore kills the SD rather than failing the truncate. Since `label.cc` invokes `d_truncate()` on both relabel and recycle, normal volume reuse reaches this path — it is not limited to `Action On Purge = Truncate`.

This PR reports `res.error()` instead, so the failure surfaces as an `EIO` device error carrying the backend's own message in `errmsg`, which is what the surrounding code already expects. Verified against a Backblaze B2 backend: the crash was reproduced twice in production, and with the deletion succeeding a 1600-chunk / 107 GB volume truncated cleanly with the SD staying up.

The unrelated `RemoteVolumeSize()` return-type issue mentioned in #2778 is deliberately left out of scope; happy to follow up separately.

#### Please check

##### General

##### Source code quality

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -95,6 +95,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - cmake: fix alphabetic requirements when some tests are disabled [PR #2762]
 - new webui: Simplify configuration and improve robustness [PR #2754]
 - bvfs: fix stack buffer overflow in GetAllFileVersions [PR #2785]
+- dplcompat: fix SIGABRT when a chunk deletion fails during truncate [PR #2779]
 
 ### Documentation
 - update bareos-github-banner.png to 13th anniversary [PR #2483]
@@ -2378,5 +2379,6 @@ If you want to migrate from your manually configured disk autochanger to simply 
 [PR #2762]: https://github.com/bareos/bareos/pull/2762
 [PR #2764]: https://github.com/bareos/bareos/pull/2764
 [PR #2772]: https://github.com/bareos/bareos/pull/2772
+[PR #2779]: https://github.com/bareos/bareos/pull/2779
 [PR #2785]: https://github.com/bareos/bareos/pull/2785
 [unreleased]: https://github.com/bareos/bareos/tree/master
--- a/core/src/stored/backends/dplcompat_device.cc
+++ b/core/src/stored/backends/dplcompat_device.cc
@@ -374,7 +374,7 @@ bool DropletCompatibleDevice::TruncateRemoteVolume(DeviceControlRecord*)
   for (const auto& [chunk_name, stat] : *chunk_map) {
     if (is_chunk_name(chunk_name)) {
       if (auto res = m_storage.remove(vol_name, chunk_name); !res) {
-        PmStrcpy(errmsg, chunk_map.error().c_str());
+        PmStrcpy(errmsg, res.error().c_str());
         dev_errno = EIO;
         return false;
       }


sebsura:
@jsfrerot please use our pr template for prs!

jsfrerot:
Sorry about that — I had dropped the template header when I wrote the description. The PR body now carries `pull_request_template.md` verbatim, with the description above the checklist.

sebsura (approved):
Thanks for the find

sebsura:
Feel free to add yourself to the AUTHORS file.

PR #2781, opened 2026-08-27
dird: add "configure delete" console command

nmcvanson:

Adds a configure delete subcommand to the Director console, the counterpart to the existing configure add:
```
configure delete <resourcetype> name=<name>
```
#### Please check

- [ ] Your name is present in the AUTHORS file (optional)`

##### General

##### Source code quality

##### Tests


nmcvanson:
Would appreciate a look this PR @sebsura

core/src/dird/ua_configure.cc
```diff
@@ -301,6 +303,36 @@ static inline bool ConfigureCreateFdResource(UaContext* ua,
...
+                                           const char* clientname)
+{
+  PoolMem basedir(PM_FNAME);
+  PoolMem path(PM_FNAME);
+  const char* dirname = my_config->GetNextRes(R_DIRECTOR, NULL)->resource_name_;
+
+  basedir.bsprintf("bareos-dir-export/client/%s/bareos-fd.d", clientname);
+  if (!my_config->GetPathOfResource(path, basedir.c_str(), "director", dirname,
+                                    false)) {
+    return;
```
sebsura, line 327: Could you add a debug message here that the path could not be found for the resource ?

core/src/lib/parse_conf.h
```diff
@@ -190,6 +190,14 @@ struct DatatypeName {
   const char* description;
 };
 
+// Describes a directive of one resource that points at another resource,
+// as found by ConfigurationParser::FindResourceReferences().
+struct ResourceReference {
+  int rcode{};
+  std::string resource_name;
+  std::string directive_name;
```
sebsura, line 198: This should store the ResourceItem* directly instead so that it becomes useful for internal uses as well.

core/src/lib/parse_conf.cc
```diff
@@ -577,6 +577,62 @@ bool ConfigurationParser::RemoveResource(int rcode, const char* name)
...
+  for (int t = 0; t < r_num_; t++) {
+    const ResourceTable& table = resource_definitions_[t];
+    if (!table.items) { continue; }
+
+    for (BareosResource* res
+         = loaded_configuration->configuration_resources_[t];
+         res; res = res->next_) {
+      if (res == target) { continue; }
+
+      *table.allocated_resource_ = res;
```
sebsura, line 597: Since this modifies the tables, it is only safe to call with the lock held.  Could you add a comment for that in parse_conf.h ?

core/src/lib/parse_conf.cc
```diff
@@ -577,6 +577,62 @@ bool ConfigurationParser::RemoveResource(int rcode, const char* name)
...
+              references.push_back({static_cast<int>(table.rcode),
+                                    res->resource_name_, item.name});
+            }
+            break;
+          }
+          case CFG_TYPE_ALIST_RES: {
+            alist<BareosResource*>* list
+                = GetItemVariable<alist<BareosResource*>*>(item);
+            if (list) {
+              for (auto* referenced : *list) {
```
sebsura, line 616: IIRC it is possible to just do `for (auto* referenced : list)` instead.

core/src/dird/ua_configure.cc
```diff
@@ -425,6 +468,189 @@ static inline bool ConfigureAdd(UaContext* ua, int resource_type_parameter)
...
+  }
+
+  bool file_removed = (unlink(path.c_str()) == 0);
+  if (!file_removed && errno != ENOENT) {
+    ua->ErrorMsg(T_("failed to remove config resource file \"%s\": %s\n"),
+                 path.c_str(), strerror(errno));
+    return false;
+  }
+
+  my_config->RemoveResource(res_table->rcode, name);
```
sebsura, line 598: Sadly this does not work.  Just because the _configuration_ does not reference a resource anymore, does not mean that the _system_ as a whole does not reference it.

I.e. I can just define a client and reference it no where, but then execute a job via
```
run job=some-job client=not-referenced yes
```
and now we have a running job that expects that this client resource remains available for its runtime.
If you were to delete it, your director would just crash.

sebsura (changes requested):
Thanks for your awesome work.  As mentioned below, this approach sadly does not work.

It is never safe to modify the running configuration, as it will lead to crashes, some of them occurring well after the resource was deleted (and crashing is probably the best that can happen).

The only reason why it works for configure_add is that its almost impossible to actually use the resource that was added before it gets removed.

To do this correctly one would have to load a new config (but not via my_config), do the deletion steps you made in that temporary config, and then atomically swap out my_config with that temporary config.

The way the code is written currently (meaning the ConfigurationParser and so on), does not make this easy.
I have some refactorings in mind that would help here, but that is sadly not on the roadmap for now.

Sorry to be the bearer of bad news.  If you create an issue for this, then we will either let you know when we refactored this, or if we implemented this ourselves.

core/src/dird/ua_configure.cc
nmcvanson: Thanks — the in-place RemoveResource() is indeed wrong, and I'll drop it.
Looking at DoReloadConfig() in dird/reload.cc, though, it seems the mechanism you describe already exists:
BackupCurrentConfiguration() installs a fresh LoadedConfiguration and parses into it, the swap happens under ResLocker, the old generation is restored on validation failure, and SetNext() guards stray accesses. Running jobs pin their generation through DirectorJcrImpl::used_config_for_job (set at ua_run.cc:387, inherited by child JCRs in jobq.cc:645), so in your example the ClientResource stays alive for the job's full runtime.
 
Would you accept configure delete implemented as: check references → remove the resource file → DoReloadConfig() → restore the file and reload back if that fails? No mutation of the live config, and the same safety profile as the existing reload command. I'd move the reference check out of the ResLocker scope to keep lock ordering consistent with the reload path.

 If you'd rather wait for the refactoring you have in mind, I'll open an issue instead

sebsura: Yes that works, as that is basically doing the same thing an administrator would do anyways.
Ping me again once you have implemented that and ill review it.

nmcvanson: Implemented, can you take a look?

PR #2784, opened 2026-08-27
Introduce bareos setup tool

pstorz:
#### Please check

##### General

##### Source code quality

##### Tests


bruno-at-bareos:
If subscription is used, there's a 404 as the `latest` url doesn't exist on the .com
```
Dry run: subscription credentials would be requested before the repository script download.
[preview] curl --fail --silent --show-error --location --config - --output bareos-setup-repository.sh https://download.bareos.com/bareos/release/latest/EL_9/add_bareos_repositories.sh
 ```

File: core/src/bareos-setup/main.cc [hunk 1]
```diff
@@ -0,0 +1,219 @@
...
+                 "privileged commands.")
+      ->default_val("127.0.0.1");
+
+  bool no_browser = false;
+  app.add_flag("--no-browser", no_browser,
+               "Do not open the browser automatically");
+
+  bool dry_run = false;
+  app.add_flag("--dry", dry_run,
+               "Dry-run mode: print commands instead of executing them");
```
sebsura:
```suggestion
               "Print the commands instead of executing them");
```
Tho options should sound the same.

core/src/bareos-setup/main.cc
```diff
@@ -0,0 +1,219 @@
...
+  std::string listen_address = "127.0.0.1";
+  app.add_option("--listen,-l", listen_address,
+                 "IPv4 address to listen on. Defaults to 127.0.0.1 "
+                 "(loopback only, the secure default). To reach the "
+                 "wizard from another host, prefer forwarding the port "
+                 "over SSH (ssh -L 19101:127.0.0.1:19101 root@host) "
+                 "rather than binding to a reachable interface: the "
+                 "setup service speaks plain HTTP and executes "
+                 "privileged commands.")
+      ->default_val("127.0.0.1");
```
sebsura: This seems to be useless when we already set the address in its initalizer.

core/src/bareos-setup/main.cc
```diff
@@ -0,0 +1,219 @@
...
+  app.footer(std::string("Version: ") + BAREOS_FULL_VERSION);
+
+  int port = 19101;
+  app.add_option("--port,-p", port, "TCP port to listen on")
+      ->default_val(19101);
+
+  std::string listen_address = "127.0.0.1";
+  app.add_option("--listen,-l", listen_address,
+                 "IPv4 address to listen on. Defaults to 127.0.0.1 "
+                 "(loopback only, the secure default). To reach the "
```
sebsura: Useless fluff.  The default can be printed in a standard way with a cli11.  Either use `capture_default_str()` or keep `default_val` and remove initializer.

core/src/bareos-setup/main.cc
```diff
@@ -0,0 +1,219 @@
...
+                 "wizard from another host, prefer forwarding the port "
+                 "over SSH (ssh -L 19101:127.0.0.1:19101 root@host) "
+                 "rather than binding to a reachable interface: the "
+                 "setup service speaks plain HTTP and executes "
```
sebsura: Which service ?

core/src/bareos-setup/main.cc
```diff
@@ -0,0 +1,219 @@
...
+                 "commands as root and speaks plain HTTP, so the setup "
+                 "token in the URL is the only access control; only do "
+                 "this on a trusted network.\n"
+                 "Consider forwarding the port over SSH instead, which "
+                 "keeps the wizard on loopback and encrypts the "
+                 "connection:\n"
+                 "  ssh -L "
+              << port << ":127.0.0.1:" << port << " root@<this-host>\n";
+  }
+  if (listen_address == "localhost") listen_address = "127.0.0.1";
```
sebsura: Why ?

core/src/bareos-setup/main.cc
```diff
@@ -0,0 +1,219 @@
...
+               "Run as interactive terminal wizard instead of web UI");
+
+  CLI11_PARSE(app, argc, argv);
+
+  std::cout << "Bareos Setup Wizard " << BAREOS_FULL_VERSION;
+  if (dry_run) std::cout << " [dry-run]";
+  std::cout << "\n";
+
+  const bool listen_all_interfaces = (listen_address == "0.0.0.0");
+  if (!tui && listen_address != "127.0.0.1" && listen_address != "localhost") {
```
sebsura: The cli11 options should make sure that you cannot set tui & listen_address at the same time.  `listen_address` should also mention what it actually does.  What are we listening for ?

core/src/bareos-setup/main.cc
```diff
@@ -0,0 +1,219 @@
...
+    if (execlp(cmd, cmd, url.c_str(), nullptr) == 0) return;
+    // execlp only returns on failure
+  }
+  std::cerr << "Could not open browser automatically.\n"
+            << "Open this URL manually: " << url << "\n";
+}
+
+int main(int argc, char* argv[])
+{
+  CLI::App app{"Bareos Setup Wizard", "bareos-setup"};
```
sebsura: This should probably have a description on what it does and how it works. I.e. it host a webui or can be used in a tui etc.

core/src/bareos-setup/main.cc
```diff
@@ -0,0 +1,219 @@
...
+  if (listen_address == "localhost") listen_address = "127.0.0.1";
+
+  // Fail fast with a clear message if a required external tool (curl,
+  // systemctl, the detected package manager, ...) is missing, instead
+  // of surfacing a raw exec failure deep inside an install step later.
+  // Dry-run only previews commands and never executes anything, so the
+  // check is skipped in that mode.
+  if (!dry_run) {
+    const auto os = DetectOs();
+    const auto missing = MissingRequiredTools(os.pkg_mgr);
```
sebsura: These two functions together do not make much sense to me.
DetectOs() checks if we can execute "/usr/bin/apt-get" and if so writes "apt" into the `pkg_mgr` field.
`MissingRequiredTools()` now checks that _both_ `apt` and `apt-get` are in PATH.  Why ? Why do we not just use the package manager we just found (i.e. /usr/bin/apt-get) ?

sebsura: What happens here if the package manager was not found (i.e. its set to "unknown") ?  It looks like it will just get ignored ? Do we check that anywhere else ?  Or will the install just fail at some point with the error message that "unknown" is not a file ?

core/src/bareos-setup/main.cc
```diff
@@ -0,0 +1,219 @@
...
+  // "localhost": the server only ever binds an IPv4 socket, but many
+  // systems resolve "localhost" to the IPv6 loopback address (::1)
+  // first. When the browser is separated from the server by a NAT/port
+  // -forwarding layer (e.g. a container's published port), an IPv6
+  // connection attempt can succeed at the TCP layer yet still get reset
+  // once traffic reaches the IPv4-only listener, since there is no
+  // matching IPv6 socket to hand it off to. "0.0.0.0" itself is not a
+  // URL a browser can open either, so it maps to 127.0.0.1 too --
+  // the server always accepts loopback connections regardless of
+  // --listen.
```
sebsura: Whats stopping us from just listening on ipv6 as well ?

core/src/bareos-setup/main.cc
```diff
@@ -0,0 +1,219 @@
...
+  pid_t child = -1;
+  if (!no_browser) {
+    child = fork();
+    if (child == 0) {
+      // Child: wait briefly then open browser
+      sleep(1);
+      OpenBrowser(display_host, port, setup_token);
+      _exit(0);
+    }
+  }
```
sebsura: Why are this two different ifs ? Can we please keep this stuff together ?  Trying to understand the logic of the program gets extremely hard if the `if`s that reference a variable are all over the place.

core/src/bareos-setup/main.cc
```diff
@@ -0,0 +1,219 @@
...
+
+  // Fork before starting the server so the child opens the browser
+  // after the parent has started listening.
+  pid_t child = -1;
+  if (!no_browser) {
+    child = fork();
+    if (child == 0) {
+      // Child: wait briefly then open browser
+      sleep(1);
```
sebsura: I can read the text just fine myself, a better comment would be _why_ we wait briefly before opening the browser.

core/src/bareos-setup/main.cc
sebsura: Why ?

core/src/bareos-setup/main.cc
```diff
@@ -0,0 +1,219 @@
...
+  // connection attempt can succeed at the TCP layer yet still get reset
+  // once traffic reaches the IPv4-only listener, since there is no
+  // matching IPv6 socket to hand it off to. "0.0.0.0" itself is not a
+  // URL a browser can open either, so it maps to 127.0.0.1 too --
+  // the server always accepts loopback connections regardless of
+  // --listen.
+  const std::string display_host
+      = (listen_all_interfaces || listen_address == "127.0.0.1")
+            ? "127.0.0.1"
+            : listen_address;
```
sebsura: Would it be possible to not hardcode 127.0.0.1 all over the place ?  Is that even necessary ?

sebsura: If we ignore 0.0.0.0, then please reject it instead of silently doing something completely different.

core/src/bareos-setup/main.cc
```diff
@@ -0,0 +1,219 @@
...
+  if (!dry_run && !IsRoot()) {
+    std::cout << "Administrator privileges are required to continue.\n";
+    if (!PrimeSudoTicket()) {
+      std::cerr << "Fatal: could not obtain sudo privileges. Run "
+                   "bareos-setup as root, or as a user allowed to "
+                   "authenticate with sudo from this terminal.\n";
+      return 1;
+    }
+    StartSudoKeepAlive();
+  }
```
sebsura: Ok, but what is the upside of this ? Could we not simply tell the user to execute this under sudo ?

core/src/bareos-setup/setup_steps.cc
```diff
@@ -0,0 +1,658 @@
...
+  if (offset == random.size()) return;
+
+  std::ifstream urandom("/dev/urandom", std::ios::binary);
+  if (!urandom) { throw std::runtime_error("Unable to open /dev/urandom"); }
+  urandom.read(reinterpret_cast<char*>(random.data() + offset),
+               static_cast<std::streamsize>(random.size() - offset));
+  if (!urandom) {
+    throw std::runtime_error(
+        "Unable to obtain cryptographically secure random data");
+  }
```
sebsura: It does not make sense to have a fallback for getrandom().  Just fail if it does not work

core/src/bareos-setup/setup_steps.cc
```diff
@@ -0,0 +1,658 @@
...
+                                       "bareos-director",
+                                       "bareos-storage",
+                                       "bareos-storage-tape",
+                                       "bareos-storage-dedupable",
+                                       "bareos-database-tools",
+                                       "bareos-tools",
+                                       "bareos-webui-new",
+                                       "bareos-webui-proxy",
+                                       "policycoreutils"};
+  if (pkg_mgr == "dnf" || pkg_mgr == "yum") { packages.push_back("mod_ssl"); }
```
sebsura: if we only allow a certain, closed & finite, set of package managers, why do we not use an enum for this ?

That way we know whether we handled every case as well.

core/src/bareos-setup/os_detector.cc
```diff
@@ -0,0 +1,112 @@
...
+/** Strip leading/trailing whitespace and optional surrounding quotes. */
+static std::string Unquote(std::string s)
+{
+  // trim whitespace
+  auto b = s.find_first_not_of(" \t\r\n");
+  auto e = s.find_last_not_of(" \t\r\n");
+  if (b == std::string::npos) return {};
+  s = s.substr(b, e - b + 1);
+  // strip quotes
+  if (s.size() >= 2 && s.front() == '"' && s.back() == '"')
```
sebsura: > $ man os-release
....
Variable assignment values must be enclosed in double **or single** quotes

core/src/bareos-setup/os_detector.cc
```diff
@@ -0,0 +1,112 @@
...
+OsInfo ParseOsRelease(const std::string& content)
+{
+  OsInfo info;
+  std::istringstream stream(content);
+  std::string line;
+  while (std::getline(stream, line)) {
+    auto eq = line.find('=');
+    if (eq == std::string::npos) continue;
+    auto key = Unquote(line.substr(0, eq));
+    auto val = Unquote(line.substr(eq + 1));
```
sebsura: > $ man os-release
...
Shell special characters ("$", quotes, backslash, backtick) must be escaped with backslashes, following
       shell style.

I.e. we need to support escaping as well if we want to parse it correctly.

Its probably not necessary, but if we decide against it, we should at least mention it it in a comment, so the next reader will know that this was not an oversight.

core/src/bareos-setup/os_detector.cc
```diff
@@ -0,0 +1,112 @@
...
+  return words;
+}
+
+OsInfo ParseOsRelease(const std::string& content)
+{
+  OsInfo info;
+  std::istringstream stream(content);
+  std::string line;
+  while (std::getline(stream, line)) {
+    auto eq = line.find('=');
```
sebsura: > $ man os-release
...
Lines beginning with
       "#" are treated as comments. Blank lines are permitted and ignored.

The code already handles this correctly, but its not obvious at first reading (and especially to someone who wants to change this code), so it would be nice to add a comment that comments are already getting skipped in the if (or via the = search if no = is found).

core/src/bareos-setup/command_runner.cc
```diff
@@ -0,0 +1,294 @@
...
+      const pid_t pid = fork();
+      if (pid < 0) continue;
+      if (pid == 0) {
+        const int devnull = open("/dev/null", O_RDWR);
+        if (devnull >= 0) {
+          dup2(devnull, STDIN_FILENO);
+          dup2(devnull, STDOUT_FILENO);
+          dup2(devnull, STDERR_FILENO);
+          close(devnull);
+        }
```
sebsura: If this is important, why do do we ignore errors, if its not important why do we do this ? A comment here would be nice.

core/src/bareos-setup/command_runner.cc
```diff
@@ -0,0 +1,294 @@
...
+        execlp("sudo", "sudo", "-v", nullptr);
+        _exit(127);
+      }
+      int status = 0;
+      waitpid(pid, &status, 0);
+    }
+  }).detach();
```
sebsura: Who stops this thread ?

core/src/bareos-setup/command_runner.cc
```diff
@@ -0,0 +1,294 @@
...
+  return WIFEXITED(status) && WEXITSTATUS(status) == 0;
+}
+
+void StartSudoKeepAlive()
+{
+  if (IsRoot()) return;
+  std::thread([] {
+    for (;;) {
+      std::this_thread::sleep_for(std::chrono::seconds(60));
+      const pid_t pid = fork();
```
sebsura: This reads suspiciously like PrimeSudoTicket.  Why is this logic duplicated ?

core/src/bareos-setup/command_runner.cc
```diff
@@ -0,0 +1,294 @@
...
+    if (n < 0 && (errno == EAGAIN || errno == EWOULDBLOCK)) break;
+    break;
+  }
+  // Emit complete lines
+  size_t pos;
+  while ((pos = buf.find('\n')) != std::string::npos) {
+    std::string line = buf.substr(0, pos);
+    // Strip trailing \r
+    if (!line.empty() && line.back() == '\r') line.pop_back();
+    cb(line, stream);
```
sebsura: Why do we copy here ? Can we not just give the callback a string_view ?

core/src/bareos-setup/command_runner.cc
```diff
@@ -0,0 +1,294 @@
...
+  // and the remaining bytes would be lost.
+  for (;;) {
+    const ssize_t n = read(fd, tmp.data(), tmp.size());
+    if (n > 0) {
+      buf.append(tmp.data(), static_cast<size_t>(n));
+      continue;
+    }
+    if (n < 0 && errno == EINTR) continue;
+    if (n < 0 && (errno == EAGAIN || errno == EWOULDBLOCK)) break;
+    break;
```
sebsura: Should we not let the caller know that reading the fd failed ? Currently he cannot distinguish between complete & incomplete (because of a read error) data.

core/src/bareos-setup/command_runner.cc
```diff
@@ -0,0 +1,294 @@
...
+  }
+  // Emit complete lines
+  size_t pos;
+  while ((pos = buf.find('\n')) != std::string::npos) {
+    std::string line = buf.substr(0, pos);
+    // Strip trailing \r
+    if (!line.empty() && line.back() == '\r') line.pop_back();
+    cb(line, stream);
+    buf.erase(0, pos + 1);
+  }
```
sebsura: I have not tried this, but this looks like it misses the last line (unless it happens to end in a \n).

core/src/bareos-setup/command_runner.cc
```diff
@@ -0,0 +1,294 @@
...
+static int RunCommandImpl(const std::vector<std::string>& argv,
+                          const std::string* input,
+                          bool use_sudo,
+                          OutputCallback cb)
+{
+  // Build final argv with optional sudo prefix. Already running as root
+  // makes sudo pointless (and would add an avoidable dependency on the
+  // sudo binary being installed), so it is skipped in that case.
+  std::vector<std::string> full_argv;
+  if (use_sudo && !IsRoot()) full_argv.push_back("sudo");
```
sebsura: Looks like the argument is not "use_sudo" after all, but rather to execute the command with root privileges.

core/src/bareos-setup/command_runner.cc
```diff
@@ -0,0 +1,294 @@
...
+    buf.erase(0, pos + 1);
+  }
+}
+
+static int RunCommandImpl(const std::vector<std::string>& argv,
+                          const std::string* input,
+                          bool use_sudo,
+                          OutputCallback cb)
+{
+  // Build final argv with optional sudo prefix. Already running as root
```
sebsura: I do not think that `argv` gets a sudo prefix.

core/src/bareos-setup/command_runner.cc
```diff
@@ -0,0 +1,294 @@
...
+  // sudo binary being installed), so it is skipped in that case.
+  std::vector<std::string> full_argv;
+  if (use_sudo && !IsRoot()) full_argv.push_back("sudo");
+  full_argv.insert(full_argv.end(), argv.begin(), argv.end());
+
+  // Build C-style argv
+  std::vector<const char*> cargv;
+  cargv.reserve(full_argv.size() + 1);
+  for (const auto& s : full_argv) cargv.push_back(s.c_str());
+  cargv.push_back(nullptr);
```
sebsura: A comment here would be nice pointing out that exec* expects a terminating nullptr.

core/src/bareos-setup/command_runner.cc
```diff
@@ -0,0 +1,294 @@
...
+
+  // Create stdout, stderr, and (when requested) stdin pipes.
+  int pipe_out[2], pipe_err[2], pipe_in[2] = {-1, -1};
+  if (input != nullptr && input->size() > PIPE_BUF) {
+    throw std::invalid_argument("command stdin exceeds PIPE_BUF");
+  }
```
sebsura: Why is this a problem?

core/src/bareos-setup/command_runner.cc
```diff
@@ -0,0 +1,294 @@
...
+      if (devnull >= 0) {
+        dup2(devnull, STDIN_FILENO);
+        close(devnull);
+      }
+    }
+    execvp(cargv[0], const_cast<char* const*>(cargv.data()));
+    // execvp failed — write error to stderr and exit
+    const char* msg = strerror(errno);
+    [[maybe_unused]] auto _ = write(STDERR_FILENO, msg, strlen(msg));
+    _exit(127);
```
sebsura: Why do we exit with 127 ? Please use named constants for things like these.

core/src/bareos-setup/command_runner.cc
```diff
@@ -0,0 +1,294 @@
...
+                        OutputCallback cb)
+{
+  return RunCommandImpl(argv, &input, use_sudo, std::move(cb));
+}
+
+bool IsToolInPath(const std::string& name)
+{
+  if (name.empty()) return false;
+  // An absolute or relative path is checked directly rather than
+  // searched for in PATH.
```
sebsura: What is a path thats neither absolute nor relative ?  And this comment answers the wrong question again.  Please describe _why_ something is done, i.e. why are "paths" searched directly and not in PATH ?

sebsura: Why is the function called "IsToolInPath" if we do not actually always check if the tool is in the path ?

sebsura: Is there even a way for this function to be called with either an empty name (see check above) or a name containing a slash ? If not, then just assert that this does not happen, and move on.  Do not make the function more complicated that needs be.

core/src/bareos-setup/command_runner.cc
```diff
@@ -0,0 +1,294 @@
...
+  const char* path_env = getenv("PATH");
+  const std::string path = path_env != nullptr ? path_env : "/usr/bin:/bin";
+  std::istringstream stream(path);
+  std::string dir;
+  while (std::getline(stream, dir, ':')) {
+    if (dir.empty()) continue;
+    const std::string candidate = dir + "/" + name;
+    struct stat st{};
+    if (stat(candidate.c_str(), &st) == 0 && S_ISREG(st.st_mode)
+        && (st.st_mode & S_IXUSR)) {
```
sebsura: Why do we use `access()` for the package manager, but only care about stat here ? Should we not do the checks consistently ?

core/src/bareos-setup/command_runner.cc
```diff
@@ -0,0 +1,294 @@
...
+{
+  if (name.empty()) return false;
+  // An absolute or relative path is checked directly rather than
+  // searched for in PATH.
+  if (name.find('/') != std::string::npos) {
+    struct stat st{};
+    return stat(name.c_str(), &st) == 0 && (st.st_mode & S_IXUSR);
+  }
+  const char* path_env = getenv("PATH");
+  const std::string path = path_env != nullptr ? path_env : "/usr/bin:/bin";
```
sebsura: 1. Why do we set it to the hardcoded "/usr/bin:/bin" ? Please add comments for these choices
2. This is completely wrong.  The default search path is "/bin:/usr/bin", but more acurately is confstr(_CS_PATH);
3. Why do we even entertain to support systems that do not have PATH set ? This just makes the code more complicated for no reason.  If somebody wants to install bareos on a system without PATH, he can either set it to the empty string, or just install bareos manually.

core/src/bareos-setup/http_server.cc
```diff
@@ -0,0 +1,364 @@
...
+#include <openssl/bio.h>
+#include <openssl/buffer.h>
+#include <openssl/sha.h>
+
+#include "embedded_assets.h"
+#include "setup_steps.h"
+
+// ---- helpers ---------------------------------------------------------------
+
+namespace {
```
sebsura: Since both this and the webui-proxy do some simple http handling, maybe we should create a single (static) library that both can use, so that we do not have to maintain two seperate http servers.

core/src/lib/ws_codec.cc
```diff
@@ -0,0 +1,303 @@
...
+
+#include <sys/socket.h>
+#include <unistd.h>
+
+#include <openssl/bio.h>
+#include <openssl/buffer.h>
+#include <openssl/evp.h>
+
+// RFC 6455 magic GUID appended to Sec-WebSocket-Key before SHA-1
+static constexpr const char kWsMagic[] = "258EAFA5-E914-47DA-95CA-C5AB0DC85B11";
```
sebsura: So this also has a duplicate websocket implementation.  It would be good if we could just have one that both applications shared.

core/src/bareos-setup/tui_wizard.cc
```diff
@@ -0,0 +1,437 @@
...
+                      {"--output", repository_script.string()});
+  const bool repo_downloaded = use_curl_config
+                                   ? RunWithInput(add_repo_cmd, curl_config,
+                                                  dry_run, {login, password})
+                                   : Run(add_repo_cmd, dry_run);
+  if (!repo_downloaded) {
+    if (!dry_run) std::filesystem::remove(repository_script);
+    return 1;
+  }
+  if (!Run({"bash", repository_script.string()}, dry_run)) {
```
sebsura: If we do the calls this way, we have to manually keep the "required tools check" and the "tool usage" code in sync.  It would be great if we had a single file, which knew of all the tools that this needs and which implements a single function for each tool.

This way each other file which needs a tool just writes (in this case) `Bash(arg1, arg2, ...)` or something like that.  Then the compiler will tell us if we forgot to define such a function or not, and if we did, we add it to that single file, and hopefully remember to also add it to the required tools.

core/src/bareos-setup/tui_wizard.cc
```diff
@@ -0,0 +1,437 @@
...
+    if (!dry_run) std::filesystem::remove(repository_script);
```
sebsura: We have what feels like hundreds of checks involving `dry_run`.  Surely there is a better way to solve this.

Maybe `Run` should just be a member function of a struct that had
1. the list of tools and the paths we found for them (so that we do not need to use execp, but can use exec directly),
2. dry_run as a member, so that run always did the correct thing,
3. maybe had a function that removes files during a real run, so you can do
    S.remove(my-file); instead of if (!dry_run) std::filesystem::remove(my-file); everywhere

And there are probably other things that one could also manage that way.

sebsura (changes requested):
I only looked at a few files so far, but to prevent a single review from being too long, ill publish this already, so you can have a look.

core/src/lib/ws_codec.cc
sebsura: I also forgot to mention: If its not in libbareos, it does not belong into `lib/`.  Please create a different directory for it, like `websocket/`.

PR #2801, opened 2026-09-05
stored: answer network requests during device init

pstorz:
The storage daemon initializes its devices in a background thread, but it was still unreachable over the network until every device was ready. With tape drives that are missing, rewinding or behind a slow autochanger this takes minutes, so "status storage" timed out and the director considered the storage daemon to be down.

Two things blocked unrelated network traffic:

device_initialization() held the configuration write lock for the whole device loop. That same lock is needed to authenticate an incoming director connection, so every connection stalled. It now snapshots the device list under a short lock and initializes without holding it.

HandleDirectorConnection() waited for init_done before dispatching any command at all. Only commands that actually use a device wait now; status, .status, setdebug=, resolve, cancel, stats and getSecureEraseCmd are answered right away.

init_done is written by the init thread and read by the connection threads, so it becomes std::atomic<bool>.

Each device now carries its own atomic DeviceInitState. It keeps status from looking at a device object that is still being constructed, and "status storage" reports whether a device is waiting to be initialized, being initialized or could not be initialized.

#### Please check

##### General

##### Source code quality

##### Tests


core/src/stored/dir_cmd.cc
```diff
@@ -302,6 +304,11 @@ void* HandleDirectorConnection(BareosSocket* dir)
           dir->signal(BNET_EOD);
           break;
         }
+        /* Only commands that actually use a device have to wait until the
+         * background device initialization is complete. */
+        if (SdDirCommandNeedsDeviceInit(cmds[i].cmd)) {
+          while (!init_done) { Bmicrosleep(1, 0); }
```
sebsura: Why do we not use a condition variable for this?

core/src/stored/stored.cc
```diff
@@ -482,13 +485,26 @@ static void CleanUpOldFiles()
...
+   * Holding on to the loaded configuration keeps the resources alive. */
+  auto loaded_configuration = my_config->GetCurrentConfiguration();
+  std::vector<DeviceResource*> device_resources;
+  {
+    ResLocker _{my_config};
+    DeviceResource* device_resource = nullptr;
+    foreach_res (device_resource, R_DEVICE) {
+      device_resources.push_back(device_resource);
+    }
+  }
```
sebsura:
```suggestion
for (auto* res = loaded_configuration->GetNextRes(R_DEVICE, nullptr); res;
     res = res->next) {
   auto* device_resource = dynamic_cast<DeviceResource*>(res);
   ASSERT(device_resource);
  ...
}
```

Instead of creating the vector we can also iterate like this directly in the loop below.

PR #2813, opened 2026-09-14
cats: size SQL escape buffers dynamically

pstorz:
#### Please check

##### General

##### Source code quality

##### Tests


core/src/cats/bvfs.h
```diff
@@ -74,8 +76,9 @@ class Bvfs {
   void SetPattern(char* p)
   {
     uint32_t len = strlen(p);
-    pattern = CheckPoolMemorySize(pattern, len * 2 + 1);
-    db->EscapeString(jcr, pattern, p, len);
+    auto escaped_pattern = db->EscapeString(jcr, std::string_view{p, len});
```
sebsura:
```suggestion
    auto escaped_pattern = db->EscapeString(jcr, p);
```
Small nitpick: it does not make sense to take the length ourselves just to call the string_view constructor, when this is exactly what the constructor would do anyways!

core/src/cats/bvfs.h
```diff
@@ -25,6 +25,8 @@
 #define BAREOS_CATS_BVFS_H_
 
 #include "lib/attr.h"
+
+#include <string_view>
```
sebsura: That include does not make sense.  If we know about the function db->EscapeString, then we already know about string_view ...

core/src/cats/sql_create.cc
```diff
@@ -36,6 +36,7 @@ static const int dbglevel = 100;
 #include "lib/edit.h"
 
 #include <string>
+#include <string_view>
```
sebsura, line 39: Same here.  If we know about EscapeString, we already know about string_view.

core/src/cats/sql_create.cc
```diff
@@ -701,17 +682,15 @@ bool BareosDb::CreateFilesetRecord(JobControlRecord* jcr, FileSetDbRecord* fsr)
       }
       // Update existing fileset record to make sure the fileset text is
       // inserted
-      PoolMem esc_filesettext(PM_MESSAGE);
-
       len = strlen(fsr->FileSetText);
-      esc_filesettext.check_size(len * 2 + 1);
-      EscapeString(jcr, esc_filesettext.c_str(), fsr->FileSetText, len);
+      auto esc_filesettext = EscapeString(
+          jcr, std::string_view{fsr->FileSetText, static_cast<size_t>(len)});
```
sebsura: if len does not get used otherwise, we can just remove it and let the std::string_view constructor take care of it.

sebsura: If len is necessary, then its still better to do:
```
std::string_view fs_text{ fsr->FileSetText };
len = fs_text.size();
```

in case the size is somehow to big for 32bit ints (which it should not !). Either way we only introduce more bugs by casting a std::size_t to int.

core/src/cats/sql_find.cc
```diff
@@ -314,29 +304,30 @@ bool BareosDb::FindLastJobid(JobControlRecord* jcr,
...
   } else if (jr->JobLevel == L_VERIFY_VOLUME_TO_CATALOG
              || jr->JobLevel == L_VERIFY_DISK_TO_CATALOG
              || jr->JobType == JT_BACKUP) {
     if (Name) {
-      EscapeString(jcr, esc_jobname, (char*)Name,
-                   MIN(strlen(Name), sizeof(esc_jobname)));
+      auto esc_jobname = EscapeString(
+          jcr,
+          std::string_view{Name, static_cast<size_t>(MIN(
+                                     strlen(Name), MAX_ESCAPE_NAME_LENGTH))});
```
sebsura: What is happening here ? Why is this necessary ?

core/src/dird/dbcheck.cc
sebsura: Why is this defined here ? I do not see a reason for it, but if there is, please document it with a comment.

core/src/dird/dbcheck.cc
```diff
@@ -255,8 +256,8 @@ static void eliminate_duplicate_paths()
     for (int i = 0; i < name_list.num_ids; i++) {
       // Get all the Ids of each name
       size_t name_len = strlen(name_list.name[i]);
```
sebsura: Just let the std::string_view constructor take care of this.

core/src/dird/dird.cc
```diff
@@ -112,12 +113,12 @@ static bool DirDbLogInsert(JobControlRecord* jcr,
...
   char dt[MAX_TIME_LENGTH];
-  PoolMem query(PM_MESSAGE), esc_msg(PM_MESSAGE);
+  PoolMem query(PM_MESSAGE);
 
   if (!jcr || !jcr->db || !jcr->db->IsConnected()) { return false; }
   length = strlen(msg);
-  esc_msg.check_size(length * 2 + 1);
-  jcr->db->EscapeString(jcr, esc_msg.c_str(), msg, length);
+  auto esc_msg = jcr->db->EscapeString(
+      jcr, std::string_view{msg, static_cast<size_t>(length)});
```
sebsura: Same here.  Ill not repeat this from now on, so please check everywhere in this changeset.  It does not make sense to compute length ourselves (and store it as an int!), just to only use it for the std::string_view constructor.

core/src/dird/ua_query.cc
```diff
@@ -191,8 +193,9 @@ static POOLMEM* substitute_prompts(UaContext* ua,
               }
             }
             len = strlen(ua->cmd);
-            p = (char*)malloc(len * 2 + 1);
-            ua->db->EscapeString(ua->jcr, p, ua->cmd, len);
+            auto escaped = ua->db->EscapeString(
+                ua->jcr, std::string_view{ua->cmd, static_cast<size_t>(len)});
+            p = strdup(escaped.c_str());
```
sebsura: From what i can tell we just copy the string to new_query below, and then overwrite p again in the for loop, so this is not strictly necessary.

It probably makes more sense to just define `subst` as an array of std::optional<std::string> and remove the copy here (and the free later on)

sebsura (changes requested):

pstorz (commented):
Looks good!

core/src/cats/sql_create.cc
pstorz: I checked this one: `<string_view>` is still directly required in `sql_create.cc` by seven explicit `std::string_view` constructions (for paths, filenames, FileSetText, and restore objects). I will keep the include unless those call sites are refactored first; relying on transitive includes would be less explicit.

core/src/dird/ua_query.cc
```diff
@@ -192,15 +191,14 @@ static POOLMEM* substitute_prompts(UaContext* ua,
...
+            if (!subst[n]) {
+              subst[n] = ua->db->EscapeString(ua->jcr, ua->cmd);
+            }
             olen = o - new_query;
-            new_query = CheckPoolMemorySize(new_query, olen + strlen(p) + 10);
+            new_query
+                = CheckPoolMemorySize(new_query, olen + subst[n]->size() + 10);
             o = new_query + olen;
-            while (*p) { *o++ = *p++; }
+            for (const char c : *subst[n]) { *o++ = c; }
```
sebsura: Just do a memcpy and increment o once afterwards please.

core/src/cats/postgresql.cc
```diff
@@ -619,6 +619,11 @@ bool BareosDbPostgresql::SqlQueryWithoutHandler(const char* query,
                                                 query_flags flags)
 {
   CheckOwnership();
+  if (ConsumeSqlEscapeError()) {
+    Mmsg(errmsg, T_("Query skipped because SQL string escaping failed.\n"));
+    return false;
+  }
```
sebsura: This is extremely brittle.  EscapeString should return `std::optional<std::string>` (or a bool) and then every caller will check if it worked or not.

core/src/cats/cats.h
```diff
@@ -973,8 +973,16 @@ class BareosDb : public BareosDbQueryEnum {
...
-  virtual std::string EscapeString(JobControlRecord* jcr, std::string_view old)
+  virtual bool EscapeString(JobControlRecord* jcr,
+                            std::string_view input,
+                            std::string& output)
       = 0;
+  std::string EscapeString(JobControlRecord* jcr, std::string_view input)
+  {
+    std::string output;
+    if (!EscapeString(jcr, input, output)) { return {}; }
+    return output;
```
sebsura: Either we only offer the interface with out parameter + bool, or we change it to std::optional<std::string>.  Im fine with either.

core/src/cats/cats.h
```diff
@@ -1004,6 +1012,13 @@ class BareosDb : public BareosDbQueryEnum {
   virtual SQL_ROW SqlFetchRow(void) = 0;
 
  protected:
+  bool ConsumeSqlEscapeError()
+  {
+    const bool failed = sql_escape_failed_;
+    sql_escape_failed_ = false;
+    return failed;
+  }
```
sebsura: We should not add more hard to track & understand state to something that is way to complicated already.

core/src/dird/ua_query.cc
```diff
@@ -192,15 +191,14 @@ static POOLMEM* substitute_prompts(UaContext* ua,
                 break;
               }
             }
-            len = strlen(ua->cmd);
-            auto escaped = ua->db->EscapeString(
-                ua->jcr, std::string_view{ua->cmd, static_cast<size_t>(len)});
-            p = strdup(escaped.c_str());
-            subst[n] = p;
+            if (!subst[n]) {
```
sebsura: We just had this if.  Move the assignment into the if above

sebsura (changes requested):

core/src/cats/bvfs.cc
```diff
@@ -486,7 +486,8 @@ void Bvfs::GetAllFileVersions(const char* path,
   size_t path_len = strlen(path);
 
   auto path_esc = db->EscapeString(jcr, std::string_view{path, path_len});
```
sebsura: As said before, it does not make sense to use strlen() to just pass the result to std::string_view.  Please just let the constructor take care of this.
```suggestion
  auto path_esc = db->EscapeString(jcr, path);
```

core/src/cats/bvfs.cc
sebsura: Same ....

core/src/cats/bvfs.cc
```diff
@@ -752,6 +755,7 @@ bool Bvfs::compute_restore_list(char* fileid,
     size_t len = strlen(tmp.c_str());
```
sebsura: ...

core/src/cats/bvfs.h
```diff
@@ -75,8 +75,12 @@ class Bvfs {
   void SetPattern(char* p)
```
sebsura: This can be a const char*

core/src/cats/postgresql.cc
```diff
@@ -348,21 +348,20 @@ void BareosDbPostgresql::CloseDatabase(JobControlRecord* jcr)
...
+std::optional<std::string> BareosDbPostgresql::EscapeString(
+    JobControlRecord* jcr,
+    std::string_view str)
 {
   DbLocker _{this};
   int error;
 
-  result.clear();
-  if (str.size() > (result.max_size() - 1) / 2) {
+  if (str.size() > (std::string{}.max_size() - 1) / 2) {
```
sebsura: This does not make it clear what the problem is.  I think its much clearer via

```
std::string result;

if (str.size() > (result.max_size() - 1) / 2) {
...
```
As then we have a direct connection to the result variable.

core/src/dird/verify.cc
```diff
@@ -799,10 +799,11 @@ void GetAttributesAndCompareToCatalog(JobControlRecord* jcr,
           if (do_Digest != CRYPTO_DIGEST_NONE) {
             auto escaped_digest
                 = jcr->db->EscapeString(jcr, Opts_Digest.c_str());
-            if (!bstrcmp(escaped_digest.c_str(), fdbr.Digest)) {
+            if (!escaped_digest) { return; }
```
sebsura: Should this not be `goto bail_out` ? Or is fname taken care of somehow?

Also we should emit a failure message similar to IsBnetError below!  Otherwise the user will not know that there was a problem.

core/src/dird/ua_restore.cc
```diff
@@ -806,8 +806,9 @@ static int UserSelectJobidsOrFiles(UaContext* ua, RestoreContext* rx)
         len = strlen(ua->cmd);
         auto fname = ua->db->EscapeString(
             ua->jcr, std::string_view{ua->cmd, static_cast<size_t>(len)});
```
sebsura:
```c++
std::string_view cmd_view{ ua->cmd };
auto fname = ua->db->EscapeString(ua->jcr, cmd_view);
len = (int) cmd_view.size();
```

core/src/cats/sql_update.cc
```diff
@@ -53,13 +53,11 @@ bool BareosDb::AddDigestToFileRecord(JobControlRecord* jcr,
...
 {
   DbLocker _{this};
-  std::string esc_digest;
-  if (!EscapeString(jcr, digest, esc_digest)) {
-    ConsumeSqlEscapeError();
-    return false;
-  }
+  auto esc_digest = EscapeString(jcr, digest);
+  if (!esc_digest) { return false; }
+  if (!esc_digest) { return false; }
```
sebsura: I think we do not have to check for random bit flips.

sebsura (changes requested):
The changeset looks very good now.   Just some small nitpicks.  Could you also take a look at what coderabbit mentianed [above](https://github.com/bareos/bareos/pull/2813#pullrequestreview-5255802673) ?

sebsura (approved):

sebsura:
@coderabbitai could you check this pr again ?  Also make sure to check your comments in #2839, #2840, #2841 (the backports) and see if they are fixed in this branch!

sebsura:
@coderabbitai full review please

